PR #36, opened 2023-03-28
Custom DB name

AlexKalopsia:
Thanks for RomM!

This is a simple change that would allow users to optionally define their own database name. 

I, for instance, have a specific naming convention on my mariadb instance.


zurdi15:
Seems a good idea, totally didn't thought about this.

Could you open the PR against develop and also modify the docker-compose.example.yml in order to let the users know this is a possibility? 

I would appreciate that a lot, thank you! I will add it to the release notes of the next version :)

PR #142, opened 2023-04-10
Optimize Dockerfile and build workflow

duhow:
- Optimize Dockerfile layers, reduce from 23 to 15
- Remove software dependencies unneeded, reduce Docker image size by 200MB (660 -> 460)
- Rework GitHub Workflow by using Actions
- Push image to GHCR as alternative to Docker Hub

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -11,20 +11,52 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+      packages: write
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Set environment variables
         run: echo "GIT_BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
-      - name: Install buildx
-        id: buildx
-        uses: crazy-max/ghaction-docker-buildx@v1
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
         with:
-          version: latest
-      - name: Login to docker hub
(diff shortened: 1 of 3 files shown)


.github/workflows/image.yml
```diff
@@ -11,20 +11,51 @@ on:
...
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            name=zurdi15/romm
+            name=ghcr.io/zurdi15/romm
+          flavor: |
+            # latest on master branch, prefix dev on other branches
+            latest=true
+            prefix=${{ github.ref != format('refs/heads/{0}', 'master') && 'dev-' || '' }},onlatest=true
```
zurdi15, line 48: This means that the ``dev-`` prefix is being added with any branch but master right?

.github/workflows/image.yml
```diff
@@ -11,20 +11,51 @@ on:
...
+          tags: |
+            type=raw,value=${{ inputs.version }}
+          labels: |
+            org.opencontainers.image.version=${{ inputs.version }}
+      - name: Build image
```
zurdi15, line 53: Is there any possibility here to just build for ``arm64`` in every branch except ``master`` as I am developing in a RPi and at least for me doesn't make sense to build for ``amd64`` in developing branches?

.github/workflows/image.yml
duhow: That's correct, all `tags` will have the prefix `dev-` if the branch is not `master`.

.github/workflows/image.yml
duhow: Yes, you can copy the same code for github.ref and apply it in here.
https://github.com/docker/build-push-action#inputs

```yaml
platforms: |
  linux/arm64
  ${{ github.ref == format('refs/heads/{0}', 'master') && 'linux/amd64' || '' }}
```

However, the build time is going to be slow either way because of `buildx` for `arm64` which you are interested in.
There is no problem in building both, then any machine (RPi or computer) will only download its platform layers only. So I would keep the current build.

zurdi15: Ohh okay, I see. I keep then the current code. Thank you for the explanation

PR #243, opened 2023-05-09
Modernize development process and update docs

GAntoine:
While trying to setup this project locally to make changes, I came across a number of small issues that made the process kinda clunky. This PR attempts to both update the development process/tools, and the developer docs (with clear instructions).

### Changes

* Switch to [Poetry](https://python-poetry.org/docs) for dependency management in python
  * Poetry has the added benefit of managing virtual environments internally
* Use [python-dotenv](https://pypi.org/project/python-dotenv/) to consume env variables
  * `envs.env` -> `.env`
  * `os.getenv` -> `os.environ.get`
* Move the `backend` folder up a level
  * `/backend/src/*` -> `/backend/*`
* Update `Dockerfile` to support these changes
* Merge all developer docs into `DEVELOPER-SETUP.md`
  * Also expands the docs with more detailed setup instructions

### Migrating an existing project

```sh
mv envs.env .env

curl -sSL https://install.python-poetry.org | python3 -
poetry shell
poetry install
````

### CODE REVIEW

**Comment by zurdi15** [comment]
This seems so interesting and I appreciate so much the help with this because there is a lot of work to do in terms of development and ci/cd for this project

File: .gitignore [hunk 1]
```diff
@@ -27,17 +27,14 @@ coverage
...
 
 # secrets
-envs.env
+.env
 
 # database
 mariadb
 
-# config test
-romm
```
zurdi15, line 43: I think ``romm`` folder should stay because is the folder I use to mock the ``library``, ``config.yml`` and other stuff to test while developing

backend/config/__init__.py
```diff
@@ -1,4 +1,7 @@
 import os
+from dotenv import find_dotenv, load_dotenv
+
+load_dotenv(find_dotenv())
```
zurdi15: How does this works? Even having the ``.env`` file in the root of the project this library is going to find it?

.gitignore
gantoine: Ah cool I'll add a comment explaining what it's for and why it should stay

backend/config/__init__.py
gantoine: Exactly! I can actually simplify this a bit by removing `find_dotenv`, as `load_dotenv` calls it internally (if no path is given). But essentially it looks for a file called `.env` in the root directly, then starts working it's way down.

GAntoine:
When you get the chance, pull this branch locally and give it a try by going through the setup steps. Then try building it with `docker build --tag romm/modernize-dev-process -f ./docker/Dockerfile .`. FWIW I got it to build and run the dev image in docker earlier today.

backend/config/__init__.py
zurdi15: Ah I see. But how it finds the .env file if it is placed in the parent directory of backend directory?

gantoine: ✨ _Magic_ ✨ 

In all seriousness, seems I read the source wrong; it actually walks **up** to the project's root looking for `.env`. So you can call `python main.py` while in the `backend` directory and it'll still find it.

https://github.com/theskumar/python-dotenv/blob/137bc3dc0b8cf3d417a1e800c4065c526e3fb96a/src/dotenv/main.py#L252

zurdi15:
I am having problems setting up poetry, but I think it is something related with my environment (I am developing in a raspbery pi 4). It gets stuck after the installation of the first dependency. My guess is that internally the installer (I don't know if it uses pip or something similar) doesn't release the lock.

This is where is gets stuck:

```
Installing dependencies from lock file

Package operations: 25 installs, 0 updates, 0 removals

  • Installing idna (3.4): Pending...
  • Installing sniffio (1.3.0): Pending...
```

After canceling if shows where it is stucked:

```
Exception ignored in: <module 'threading' from '/usr/lib/python3.9/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1415, in _shutdown
    atexit_call()
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib/python3.9/threading.py", line 1033, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.9/threading.py", line 1049, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt: 
```

zurdi15:
> When you get the chance, pull this branch locally and give it a try by going through the setup steps. Then try building it with `docker build --tag romm/modernize-dev-process -f ./docker/Dockerfile .`. FWIW I got it to build and run the dev image in docker (with maria) earlier today.

I could steup but I think one step is missing in order to allow the frontend to load covers, screenshots and download files while developing: a symlink to the ``romm`` folder that I use to mock the library, resources, etc inside the ``assets`` folder. Something like:

``ln -s romm frontend/assets/romm``. Probably this can be done in a better way but was the faster way I found to allow the frontend to find the resources/roms

zurdi15:
I made some changes to adapt everything mock related to your proposal. Should I open a PR to your branch so you can check if it works for you and you find it useful or how should we do it?

GAntoine:
Thank you for merging it!

> Solved configuring poetry with poetry config experimental.new-installer false

Sounds like the parallel installer doesn't play nice with the environment on the RPi, but I'm glad to see you found a fix.

Issue #186, opened 2023-04-20
[Feature] ROM filename/savename correction


IngwiePhoenix:
Hello there!

I found this through Reddit r/selfhosted! :)

**Is your feature request related to a problem? Please describe.**
I have been dumping and collecting ROMs for literally one and a half decades and some of the filenames - and thus associated saves - have become increasingly mixed-named. I try to adhere to a scheme manually, but it'd be helpful if this tool could help out! Since you are already fetching data from IGDB, this might be useful.

**Describe the solution you'd like**
The software should be able to offer the user to automatically fix filenames - and associated save names - based on a preconfigured scheme. For instance, the one I use is `$title [$region $lang][$version].$ext`. The user should be able to configure this in their config to fit it into their default scheme.

**Describe alternatives you've considered**
Offer a suggested filename.

**Additional context**
None I could think of.

XargonWan:
Can I suggest `$title [$region][$lang][$version].$ext`?

GAntoine:
> Since you are already fetching data from IGDB, this might be useful

From what I can tell, IGDB can't tell us the region of a specific ROM file, just a list of regions associated with the game. An integration with no-intro might prove more fruitful, since we can compare SHA hashes.

zurdi15:
I close the issue since we already have a specific way to manage ``regions``/``languages`` and for now we are not going to allow different structures for it.

PR #253, opened 2023-05-17
Action to mass rename all ROMs for a platform

### LINKED ISSUE #251: [Feature] Library Section: Add Option to mass rename properly identified files
When you identify a rom file manually,Romm has an option for renaming the file according to a rule/set convention, but I have not found  way to do this later without re-identifying the rom manually again.
Additionally, having a way to mass rename files according to the same convention would be a godsend.

Could be a trigger option in the library settings.

GAntoine:
Closes #251 

This PR introduces a new socket endpoint that allows mass-renaming all ROMs for a platform (when in the Gallery view). This should make it easier for users with large libraries who's games have been correctly categorized to clean up their files.

### What changes

* New socket endpoint `mass_rename`
  * Calls new `rename_all_roms` with a given platform slug
* Socket event names are now prepended with the endpoint
  * `scan:done`, `scan:done_ko`, `mass_rename:done`
* Get eslint working with a basic config and ignore file

At some point I'd like to add a simple prettier config (and settle on consistent code style).

### TODO

- [ ] Add confirmation popup before mass-rename action
- [ ] Use config value for filename structure
  - #186 
- [ ] Add comments to PR explaining some changes


zurdi15:
Looks really nice, didn't know about the good practice of prepend sockets events with the endpoint.

Apart from that, I think it is better to open the pull requests to ``develop`` and when it has enough features/bug fixes and have been properly tested to release a new version, a PR against ``master`` can be opened.

Also I have the bad practice of sometimes, for small fixes or refactor changes, working directly on ``develop`` when I am the only developer, so I pushed the last commits to ``develop`` I had in local and I will start to work strictly with branches.

backend/endpoints/rom.py
```diff
@@ -14,51 +14,88 @@
...
+
+    for rom in platform_roms:
+        try:
+            igdb_rom_data = igdbh.get_rom_by_id(rom.r_igdb_id)
+            fs.rename_rom(
+                platform.fs_slug, rom.file_name, igdb_rom_data[0].get("r_name", "")
+            )
+            log.info(f'Renamed {rom.file_name} to {igdb_rom_data[0].get("r_name", "")}')
+        except RomAlreadyExistsException as e:
+            log.info(str(e))
```
zurdi15: I would change this logs from the exceptions as warnings.

Also, RomM custom exceptions have a __repr__ with the message to show, but I don't know what is better, an explicit cast to string ``str(e)`` or just using the message property ``e.message``. We should agree in one of them to have a consistent code.

frontend/src/views/Gallery.vue
```diff
@@ -1,126 +1,209 @@
...
 
 // Event listeners bus
-const emitter = inject('emitter')
-emitter.on('filter', () => { filterRoms() })
+const emitter = inject("emitter");
+emitter.on("filter", () => {
+  filterRoms();
+});
 
+const socket = io({
```
zurdi15: Is it better to open a socket connection each time the view loads instead to open it just when scanning/renaming?

GAntoine:
> Apart from that, I think it is better to open the pull requests to develop and when it has enough features/bug fixes and have been properly tested to release a new version, a PR against master can be opened.

That's debatable: in most open-source projects, `master/main` is the bleeding edge, and work should be branched from it. If you'd like to keep releases in a separate branch, you could have a `release` branch (or `release-x.x.x` branches) and merge commits into it when they're tested and ready to deploy.

For simplicity's sake, I'd recommend switching the base branch back to master.

> I pushed the last commits to develop I had in local and I will start to work strictly with branches

That's wise, I do the same thing for the projects I maintain.

frontend/src/views/Gallery.vue
gantoine: Ideally we want a single connection for the entire app, which is something I'll try to do in this PR.

As for whether to keep that connection open all the time, or disconnect and reconnect it, it really depends on our use case. Opening and closing sockets is CPU intensive, and if the typical install won't be running more then a handful of connections (<100) at a time, then there's no hard in keeping them open. However, since rescan actions are probably not very common, there is a good argument for opening the socket just-in-time, and closing it after use.

zurdi15: Oh okay, I think then we should go for a single app connection and keep it open because I guess it will be useful in the future when RomM grows much more and as you mention this is not going to have hundreds of concurrent connections/users (probably).

zurdi15:
> > Apart from that, I think it is better to open the pull requests to develop and when it has enough features/bug fixes and have been properly tested to release a new version, a PR against master can be opened.
> 
> That's debatable: in most open-source projects, `master/main` is the bleeding edge, and work should be branched from it. If you'd like to keep releases in a separate branch, you could have a `release` branch (or `release-x.x.x` branches) and merge commits into it when they're tested and ready to deploy.
> 
> For simplicity's sake, I'd recommend switching the base branch back to master.
> 
> > I pushed the last commits to develop I had in local and I will start to work strictly with branches
> 
> That's wise, I do the same thing for the projects I maintain.

You convinced me. Anyway is mostly like the same method I am proposing but instead of ``master/develop`` it will be ``release/master``. At the end I want a branch to merge all the PR's and test it with all the PR's integrated but having an extra branch for the releases. Already reverted the base branch change.

zurdi15:
Since you are making a new socket endpoint and refactoring some parts of the front code, I will merge this [PR](https://github.com/zurdi15/romm/pull/256) with the API wrapper to let you adapt the code to use it if needed. Maybe we can add the socket logic into the wrapper

GAntoine:
@zurdi15 After thinking about this today, I think there's some preliminary work that needs to happen before we can tackle a feature like this (namely an integration with no-intro). I'm going to close this PR in a bit, after I've extracted some of the refactoring we can still use into a separate PR.

zurdi15:
Yeah would be nice to first integrate no-intro in order to have a proper renaming feature 👍

PR #277, opened 2023-06-07
Configurable backend port

calebsmithdev:
As a Mac developer, my localhost port is already taken by Airplay Receiver ([see this link](https://apple.stackexchange.com/questions/431154/controlcenter-app-listens-to-port-5000-tcp-on-monterey-os-is-that-normal-why-d)). This PR allows for the backend port to be configurable, or to use the port 5000 as the fallback. This is handled at the Python and Vite level.

Additionally - I added support for Vite to pull in both the frontend `.env` and parent directory `.env` for future handling of custom configurations.

### DIFF
--- a/frontend/vite.config.js
+++ b/frontend/vite.config.js
@@ -5,64 +5,70 @@ import { VitePWA } from 'vite-plugin-pwa'
 import pluginRewriteAll from 'vite-plugin-rewrite-all'
 
 // Utilities
-import { defineConfig } from 'vite'
+import { defineConfig, loadEnv } from 'vite'
 import { fileURLToPath, URL } from 'node:url'
 
 // https://vitejs.dev/config/
-export default defineConfig({
-  plugins: [
-    pluginRewriteAll(),
-    vue({ 
-      template: { transformAssetUrls }
-    }),
-    vuetify({
-      autoImport: true,
-      styles: {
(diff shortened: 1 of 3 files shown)


zurdi15:
Oh nice, flexibility is always a good thing :)

frontend/vite.config.js
```diff
@@ -5,64 +5,70 @@ import { VitePWA } from 'vite-plugin-pwa'
...
-          target: 'http://localhost:5000',
-          changeOrigin: false,
-          secure: false,
-          rewrite: (path) => path.replace(/^\/api/, ''),
-        },
-        '/ws': {
-          target: 'http://localhost:5000',
-          changeOrigin: false,
-          ws: true,
+export default defineConfig(({mode}) =>{
```
zurdi15, line 12: What is this ``mode`` parameter?

calebsmithdev: More information from the docs on this specific use case: https://vitejs.dev/config/#using-environment-variables-in-config and https://vitejs.dev/guide/env-and-mode.html

`mode` in Vite is the env config based on the mode you are running. You can trigger dev modes, prod modes, etc. Each with their own config. In this case, the mode is just the default `.env` that Vite needs to reference from programmatically.

GAntoine (approved):

PR #279, opened 2023-06-07
Update README.md - clarify installation and folder structure sections

moyer34:
A few readme changes to better guide the user toward a successful first load: 

Updated installation section to clarify the need for the IGDB key to run the application and setting up folder structure. Failing to do these things before spinning up Romm prevents scanning from working, so they are best to handle first. Additionally, the folder structure section is updated to call the user's attention to the list of platform names and that custom names may be used in conjunction with entries in the config.yml file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,19 +71,30 @@ Inspired by [Jellyfin](https://jellyfin.org/), allows you to manage all your gam
 
 ## 🐳 Docker
 
-Last version of the docker [image](https://hub.docker.com/r/zurdi15/romm/tags).
+__Prerequisites:__
+Install Docker and Docker-compose. Last version of the docker [image](https://hub.docker.com/r/zurdi15/romm/tags).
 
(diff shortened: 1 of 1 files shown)


moyer34:
Hope this comment finds you well. I've made a few suggested edits to the Readme based on a couple pain points I had spinning up Romm (which is generally awesome once I got it working)! I'm basically suggesting that you point users to the need to get the IGDB secret and check folder structure before spinning up the container. I initially got stuck with nothing happening because I thought I could comment out the IGDB dependencies. Scanning failed as a result until I got the secret and updated the docker-compose.yml.

GAntoine (approved):
I made some small tweaks to the language while keeping the core of what you wrote. Good change IMO.

zurdi15:
Hi guys! Thank you for this. I know I need to make proper docs and improve them for new users, so this is much appreciated.

moyer34:
Honestly, it's pretty intuitive once you know you need to use _all_ the environment parameters. Awesome project and work!

PR #281, opened 2023-06-09
Pagination for large libraries

### LINKED ISSUE #89: [Bug] UI freezes up with huge/massive collection
**RomM version**
V1.5

**Describe the bug**
When attempting to load a platform with a large number of entries, there is a huge lag.
700 entries: takes 1-5 seconds, loads smoothly
8500 entires: takes 60+ seconds, sometimes Chrome will prompt to terminate the tab. Scrolling down quickly seems to lazy load the image for each game (in this case, 'cover missing' for most of them). 

**To Reproduce**
1. Have two folders, folder 'ds' has 1 file and folder 'gba' 8500 files. Scan the files.
2. After the scan, click on the 'ds' section. Then click on the 'gba' section and wait for the lag.

**Expected behavior**
It would be nice if large collections could be loaded lazily, instead of the page trying to fetch every game at once. Actually, the implementation doesn't really matter - I'd just like the page to load faster, even if the cover images load in later.

Windows 10, Chrome v109.0.5414.75

### LINKED ISSUE #280: [Feature] paginated system page
For one of the systems hi have 600+ Roms. At this number the page loads very slowly (Firefox asked me to stop the tab even).

In addition to "show all", maybe the system pages could be paginated so that 50/100 show at a time. This would optimize load time with usability.

GAntoine:
There's considerable slowdown (lag) in the app when the number of roms in a library hits 500+. This PR attempts to mitigate that issue in a couple ways.

### What changed?

* Introduce cursor-based pagination with [fastapi_pagination](https://uriyyo-fastapi-pagination.netlify.app/)
  * Gallery view holds next cursor value and requests further entries with it
  * ~~Add [axios-cache-interceptor](https://axios-cache-interceptor.js.org/) to cached cursored rom list requests~~
* Set `timestamp` on cover url when cover is updated
  * This allows for cache-busting images but only when required
* Move search to backend and use rom list endpoint with `ilike` filter
  * Add [lodash](https://lodash.com/) to debounce search input
* Fix accessing `file_name` when updating rom, but "Rename rom" not checked

### What's left?

- [x] Get [vue-virtual-scroller](https://github.com/Akryum/vue-virtual-scroller) working, at least for the list view.
- [x] Use `onScroll` instead of needing a dedicated button to load more

### How do I test this?

Either already have a very large library, or generate a ton of empty ZIP files and run a scan to load them in the DB.

```sh
for i in {0001..1000}; do touch "file_${i}.zip"; done
```

Closes #280 
Closes #89

### DIFF
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -64,7 +64,7 @@ def rom(id: int):
 
 @router.get("/platforms/{p_slug}/roms", status_code=200)
 def roms(
-    p_slug: str, size: int = 50, cursor: str = "", search_term: str = ""
+    p_slug: str, size: int = 60, cursor: str = "", search_term: str = ""
 ) -> CursorPage[RomSchema]:
     """Returns all roms of the desired platform"""
     with dbh.session.begin() as session:
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -74,7 +74,7 @@ def roms(
         if search_term:
             return paginate(
                 session,
-                qq.filter(Rom.file_name.ilike(f'%{search_term}%')),
+                qq.filter(Rom.file_name.ilike(f"%{search_term}%")),
                 cursor_params,
             )
 
@@ -90,28 +90,27 @@ async def updateRom(req: Request, p_slug: str, id: int) -> dict:
     db_rom: Rom = dbh.get_rom(id)
     platform: Platform = dbh.get_platform(p_slug)
 
+    file_name = updated_rom.get("file_name", db_rom.file_name)
+
     try:
-        fs.rename_rom(platform.fs_slug, db_rom.file_name, updated_rom["file_name"])
+        if file_name != db_rom.file_name:
+            fs.rename_rom(platform.fs_slug, db_rom.file_name, file_name)
     except RomAlreadyExistsException as e:
         log.error(str(e))
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e)
         )
 
-    updated_rom["file_name_no_tags"] = get_file_name_with_no_tags(
-        updated_rom["file_name"]
-    )
+    updated_rom["file_name_no_tags"] = get_file_name_with_no_tags(file_name)
+    updated_rom.update(fs.get_cover(True, p_slug, file_name, updated_rom["url_cover"]))
     updated_rom.update(
-        fs.get_cover(True, p_slug, updated_rom["file_name"], updated_rom["url_cover"]),
-        fs.get_screenshots(
-            p_slug, updated_rom["file_name"], updated_rom["url_screenshots"]
-        ),
+        fs.get_screenshots(p_slug, file_name, updated_rom["url_screenshots"]),
     )
     dbh.update_rom(id, updated_rom)
 
     return {
         "rom": dbh.get_rom(id),
-        "msg": f"{updated_rom['file_name']} updated successfully!",
+        "msg": f"{file_name} updated successfully!",
     }
 
 
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -1,4 +1,7 @@
 from fastapi import APIRouter, Request, status, HTTPException
+from fastapi_pagination.ext.sqlalchemy import paginate
+from fastapi_pagination.cursor import CursorPage, CursorParams
+from pydantic import BaseModel
 
 from logger.logger import log
 from handler import dbh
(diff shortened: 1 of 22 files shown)


zurdi15:
I had a lot of front code refactor in my local, so I am going to push them to master to let you resolve any kind of conflict that will appear as I see you changed some things too. I am sorry for the inconveniences you are going to find with this.

GAntoine:
> so I am going to push them to master to let you resolve any kind of conflict that will appear as I see you changed some things too

yeah that's fine, my work should be targeted enough that i can just splice it in. might need a new branch/PR but we'll see.

backend/endpoints/rom.py
```diff
@@ -10,47 +13,115 @@
 router = APIRouter()
 
 
+class RomSchema(BaseModel):
```
zurdi15, line 16: Is there a way to `merge` somehow this `RomSchema` with the `Rom` model? I guess as all the `Rom` attributes are `Columns` it can't be translated directly to python types, but I was just wondering if we can just have a single place where to put the models schamas and extend from there.

backend/endpoints/rom.py
```diff
@@ -10,47 +13,115 @@
...
+    tags: list
+
+    multi: bool
+    files: list
+
+    url_screenshots: list
+    path_screenshots: list
+
+    class Config(BaseModel.Config):
+        orm_mode = True
```
gantoine, line 55: At some point I'd like to move these model definitions into their own files, once we have more of them

backend/handler/db_handler.py
```diff
@@ -75,56 +69,51 @@ def add_rom(self, rom: Rom) -> None:
         except ProgrammingError as e:
             self.raise_error(e)
 
-
     def get_roms(self, p_slug: str) -> list[Rom]:
         try:
-            with self.session.begin() as s:
-                return s.scalars(select(Rom).filter_by(p_slug=p_slug).order_by(Rom.file_name.asc())).all()
+            return select(Rom).filter_by(p_slug=p_slug).order_by(Rom.file_name.asc())
```
gantoine, line 74: Only change here is removing the scalar so we can access the query directly. This might be something we want to do for all these, for consistency.

pyproject.toml
```diff
@@ -20,6 +20,8 @@ Unidecode = "1.3.6"
 emoji = "2.2.0"
 python-dotenv = "1.0.0"
 ipdb = "0.13.13"
+fastapi-pagination = "^0.12.4"
+sqlakeyset = "^2.0.1684285512"
```
gantoine, line 24: required for offset-free pagination with sqlalchemy

backend/endpoints/rom.py
zurdi15: I was asking the same, didn't saw this comment. I agree with that, I guess for now we can leave it here and move it when it grows a little bit. For now it is manageable

backend/handler/db_handler.py
zurdi15: Yes totally agree. To be honest I was learning SQAlchemy while developing RomM, so I see now a lot of imporvements with it.

zurdi15 (approved):

zurdi15:
Let me test the branch before merging, but everything seems really OK for me.

GAntoine:
> Let me test the branch before merging

Yes please! The most important things to test are pagination, search in a library, updating and deleting ROMs, and the 3 different views (big, small, list).

zurdi15:
First of all I must say that the load on scroll works perfectly fine, but I tested everything a little bit and I found some problems:
 - Pagination doesn't work for me in the big and small view, just in the list view
 - Editing the rom broke. Now the edit dialog doesn't show the rom info and can't edit it
 - Delete rom works fine, but I notice now that after delete/scan something the you need to reload the page to show the changes, I think is a good thing to reload the gallery itself after delete or scan.
 - Search in library works fine, but I realized that when searching with the filter, the pagination doesn't work, it will only show the first 100 entries it finds.
 
 I'll keep testing it anyway.

GAntoine:
> Pagination doesn't work for me in the big and small view, just in the list view

Scrolling wasn't hooked up correctly, fixed now

> Now the edit dialog doesn't show the rom info and can't edit it

Yep forgot to replace a `.data.data`, fixed

> after delete/scan something the you need to reload the page to show the changes

I removed `axios-cache-interceptor` which was causing the issue, we can revisit smarter caching at a later date

> when searching with the filter, the pagination doesn't work

Yeah I never build that, added it and seems to work, though there is 1 bug I need to fix

zurdi15:
Wow you are fast. I will test it asap and merge :)

GAntoine:
@zurdi15 I had a though: instead of using an infinite scroll, what it we did some simple [pagination](https://vuetifyjs.com/en/components/paginations/)? It would limit the slowdown of having to render hundreds of rows/cards, and we could do it in chunks of 54 or so.

GAntoine:
Good catch, both issues should be fixed 👍🏼

zurdi15 (approved):

PR #286, opened 2023-06-19
Allow for direct rom download + server-size ZIP

### LINKED ISSUE #266: [Bug] Can't download large files
**RomM version**
1.8.4

**Describe the bug**
downloading big ROM (2GB+) doesn't work. small ROM (20MB) download fine

**To Reproduce**
1. Select Switch as a platform
2. Download a game
3. It says "Downloading xxx..."
4. Nothing happens

**Expected behavior**
ROM should start downloading, just like when you download a smaller ROM

**Additional context**
docker logs -f romm doesn't show any error

GAntoine:
Running all file downloads through a frontend file downloader/zipper can be very slow for large files, and bypasses the browser's built-in download manager.

### What changed?

* Direct-download single-file ROMs
  * Send the download link for single-file ROMs to the frontend
* Move multi-file zipping to the backend
  * Generate ZIP file and write to temp storage
  * Stream the file response to the frontend
* Render a different button based on single-vs-multi-file ROM

Fixes #266


zurdi15:
Sorry for being so slow right now.

I could test it a little bit and I found some bugs and inconsistencies, but let me test it a little bit more and i'll write some of the issues. Thank you a lot for you effort and help, really.

GAntoine:
> let me test it a little bit more and i'll write some of the issues

Yes let me know what you find! Obviously this is an unconventional change, and since we have no testing framework in place it comes down to doing it manually. Maybe once this is merge we release an `rc` so people can test it with their setup?

As for the migration to Node, I don't have any in-progress work, so it's a good time as any if we decide to switch.

GAntoine:
@zurdi15 It's been a couple weeks since the last comments, any chance you'll have some time to look into this? Otherwise I'm tempted to just merge it and release an RC for those who want to try it in the wild (before we release a minor version).

zurdi15:
Hi dude! I am so sorry, in summer I don't have time at all, so I agree with releasing a rc version to let users test it but as soon as I have time I will go hard with RomM again :).

Thank you so much for taking care of it all this time. I think you sent me a discord request right? We can speak there better.

Again, thank you very much for your effort.

GAntoine:
> In summer I don't have time at all

No apology needed, open-source is best approached in a contribute-when-you-can way, as life/school/work is _far_ more important.

> I think you sent me a discord request right? We can speak there better.

Yep that was me, I'll share some thoughts/ideas with you there, which we can refine and build upon.

zurdi15:
Hi! 

Just tested the new changes again and the download button from gallery (list and grid) is working but the download button in the details view doesn't work. It redirects me to a url that is blocked by default in my Google Chrome at least [about:blank#blocked].

Can't figure why because the `downloadUrl` is built the same way in both cases.

GAntoine:
Ay good catch! There was a bug with how `downloadUrl.value` was set in the details view (still new to this whole vue thing), fix is in the latest commit.

frontend/src/components/GameGallery/Card/Base.vue
```diff
@@ -1,17 +1,14 @@
...
 
 // Props
 const props = defineProps(['rom'])
-const downloading = storeDownloading()
 </script>
 
 <template>
     <v-hover v-slot="{isHovering, props}">
         <v-card 
-            :loading="downloading.value.includes(rom.file_name) ? 'rommAccent1': null"
```
zurdi15, line 14: Why do we remove the loader here?

zurdi15: Never mind, just realized that web browser now handles the download itself, good work there

zurdi15:
The entire `edit` and `Search IGDB` with the `rename rom` features are broken :(

GAntoine:
I'm happy you caught that now, cause it would have bit us at some point in the near future. Since I'm now adding extra fields in the schema, which aren't present in the model, attempting to update the model with those fields fails. The fix is be explicit about which fields should be sent to `/update`, which I changed in the latest commit.

zurdi15:
> Render a different button based on single-vs-multi-file ROM

By the way I just noticed this but I can't see the difference, which glyph do you want to render?

zurdi15 (approved):

zurdi15:
Seems pretty good for me, everything I tested worked so smooth. I'll let you decide when to merge just in case you want to work in something else.

PR #310, opened 2023-08-01
Backend testing refactor

GAntoine:
There's a lot happening in this PR so I'll break down the high-level changes:

* Added `pytest` to unit test backend
  * Added `pytest-env` to define custom environment for testing
  * Added `pytest-vcr` to record & replay network requests
  * Added `pytest-mock` to spy on method calls
* New `romm_test` folder with default structure used exclusively for tests
* Updates developer setup docs to reflect changes
  * Adds `setup.sql` script for testing setup

And the minor ones:

* `ConfigLoader` now accepts a custom path (for testing falls back to default)
* A default `config` object is created and can be used to access configuration
  * `DBHandler` no longer requires a config object on init
* `SUPPORTED_DB_DRIVERS` env variable no longer required
* Renamed `get_matched_rom_by_id` -> `get_matched_roms_by_id`
* New component for platform icon + fallback to default
* General code cleanup/formatting with tiny tweaks


backend/config/config_loader.py
```diff
@@ -1,109 +1,76 @@
...
-    SUPPORTED_DB_DRIVERS,
     SQLITE_DB_BASE_PATH,
     ROMM_USER_CONFIG_PATH,
 )
 from logger.logger import log
 
 
 class ConfigLoader:
-    def __init__(self):
+    def __init__(self, config_path: str = ROMM_USER_CONFIG_PATH):
```
GAntoine, line 17: This is so I can test a custom config file

backend/config/config_loader.py
```diff
@@ -1,109 +1,76 @@
...
+        if ROMM_DB_DRIVER == "sqlite":
+            if not os.path.exists(SQLITE_DB_BASE_PATH):
+                os.makedirs(SQLITE_DB_BASE_PATH)
+            return f"sqlite:////{SQLITE_DB_BASE_PATH}/romm.db"
 
-        else:
-            log.critical(f"{ROMM_DB_DRIVER} database not supported")
-            sys.exit(3)
+        log.critical(f"{ROMM_DB_DRIVER} database not supported")
+        sys.exit(3)
```
gantoine, line 51: This is just cleanup that makes `SUPPORTED_DB_DRIVERS` no longer needed.

backend/handler/db_handler.py
```diff
@@ -80,7 +80,7 @@ def get_roms(self, p_slug: str) -> list[Rom]:
     def get_rom(self, id) -> Rom:
         try:
             with self.session.begin() as session:
-                return session.query(Rom).get(id)
+                return session.get(Rom, id)
```
gantoine, line 83: `Query.get` is deprecated

backend/handler/sgdb_handler.py
```diff
@@ -6,33 +6,33 @@
...
+            f"{self.BASE_URL}/grid/game/{game_id}", headers=self.headers
+        ).json()
+
+        if len(game_response["data"]) == 0:
+            log.info(f"Could not find {game_name} image on SteamGridDB")
+            return (game_id, game_name, self.DEFAULT_IMAGE_URL)
+
+        game_image_url = game_response["data"][0]["url"]
+
+        return (game_id, game_name, game_image_url)
```
gantoine, line 38: This was refactored for clarity (even though its unused)

backend/handler/tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.fixture(scope="session", autouse=True)
+def setup_database():
+    alembic.config.main(argv=["upgrade", "head"])
+
+
+@pytest.fixture(autouse=True)
+def clear_database():
+    with session.begin() as s:
+        s.query(Platform).delete(synchronize_session="evaluate")
+        s.query(Rom).delete(synchronize_session="evaluate")
```
gantoine, line 23: This is the best way I've found (so far) to migrate a fresh test database and clear it after the tests run.

backend/logger/tests/test_logger.py
```diff
@@ -0,0 +1,17 @@
...
+#    assert "info message" in caplog.text
+
+#    log.warning("Testing warning message")
+#    assert "warning message" in caplog.text
+
+#    log.error("Testing error message")
+#    assert "error message" in caplog.text
+
+#    log.critical("Testing critical message")
+#    assert "critical message" in caplog.text
```
GAntoine, line 17: Since the file logger was introduced this tests started failing 🤔

backend/models/base.py
```diff
@@ -1,3 +1,3 @@
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import declarative_base
```
gantoine, line 1: `ext.declerative` is deprecated, importing directly from `orm` is preferred

backend/utils/__init__.py
```diff
@@ -96,7 +96,7 @@ def parse_tags(file_name: str) -> tuple:
 
 def get_file_name_with_no_tags(file_name: str) -> str:
     # Use .rsplit to remove only the file extension
-    return re.sub("[\(\[].*?[\)\]]", "", file_name.rsplit(".", 1)[0])
+    return re.sub("[\(\[].*?[\)\]]", "", file_name.rsplit(".", 1)[0]).strip()
```
gantoine: Removes leading and trailing whitespace

backend/utils/fastapi.py
```diff
@@ -19,35 +14,75 @@ def scan_platform(fs_slug: str) -> Platform:
...
+            "file_size_units": file_size_units,
+            "multi": rom_attrs["multi"],
+            "region": reg,
+            "revision": rev,
+            "tags": other_tags,
+        }
+    )
+    rom_attrs["p_igdb_id"] = platform.igdb_id
+    rom_attrs["p_slug"] = platform.slug
+    rom_attrs["p_name"] = platform.name
```
gantoine, line 90: This is mostly unchanged, just cleaned up a bit

backend/utils/fs.py
```diff
@@ -143,11 +144,11 @@ def store_default_resources() -> None:
...
-def _exclude_platforms(platforms) -> None:
-    _ = [
-        platforms.remove(excluded)
-        for excluded in cl.config["EXCLUDED_PLATFORMS"]
-        if excluded in platforms
+def _exclude_platforms(platforms: list):
+    return [
+        platform
+        for platform in platforms
+        if platform not in config["EXCLUDED_PLATFORMS"]
```
gantoine, line 151: Returning an array is usually preferred over directly editing it

backend/handler/igdb_handler.py
```diff
@@ -9,204 +9,270 @@
...
+                data=f"""search \"{search_term}\";
+                    fields id, slug, name, summary, screenshots;
+                    where platforms=[{p_igdb_id}] {category_filter};
+                """,
+                timeout=120,
+            ).json()[0]
         except IndexError:
             return {}
 
+    def _search_rom(
```
zurdi15: is this `_search_rom` duplicated method maybe a mistake from the previous merge?

gantoine: yep, will fix in an upcoming commit

backend/handler/igdb_handler.py
```diff
@@ -9,204 +9,270 @@
...
-            res: dict = requests.post(self.platform_url,
-                                      headers=self.headers,
-                                      data=f"fields id, name; where slug=\"{slug.lower()}\";",
-                                      timeout=120).json()[0]
-            igdb_id = res['id']
-            name = res['name']
+            res = requests.post(
+                self.platform_url,
+                headers=self.headers,
+                data=f'fields id, name; where slug="{slug}";',
```
zurdi15: in the merge was missed the ``.lower()`` in the slug to avoid case-sensitive search

backend/handler/igdb_handler.py
```diff
@@ -9,204 +9,270 @@
...
+            return {
+                "igdb_id": res["id"],
+                "name": res["name"],
+                "slug": slug,
+            }
         except IndexError:
             log.warning(f"{slug} not found in IGDB")
-        return {'igdb_id': igdb_id, 'name': name, 'slug': slug, 'logo_path': ''}
 
+        return {}
```
zurdi15: Returning an empty dict will cause problems when a platform doesn't match, we need to return all the default parameters if platform is not found in  IGDB:

``return {'igdb_id': "", 'name': slug, 'slug': slug}``

backend/handler/igdb_handler.py
```diff
@@ -9,204 +9,270 @@
...
+        )
+
+        r_igdb_id = res.get("id", "")
+        r_slug = res.get("slug", "")
+        r_name = res.get("name", search_term)
+        summary = res.get("summary", "")
 
         if not r_igdb_id:
             log.warning(f"{r_name} not found in IGDB")
+            return {}
```
zurdi15: Will happen the same as with the platforms, we need to return a dict with default parameters

backend/handler/db_handler.py
zurdi15: True, I think I saw it from the tests warnings

backend/config/config_loader.py
zurdi15: Yeah I did it like that at the beginning thinking to support more ddatabases, but at this point I think that makes no sense

backend/handler/igdb_handler.py
zurdi15: fixed in the new commit!

backend/handler/igdb_handler.py
zurdi15: fixed in the new commit!

backend/handler/igdb_handler.py
zurdi15: fixed in the new commit!

backend/handler/tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+
+from config.config_loader import ConfigLoader
+from models.platform import Platform
+from models.rom import Rom
+
+engine = create_engine(ConfigLoader.get_db_engine(), pool_pre_ping=True)
+session = sessionmaker(bind=engine, expire_on_commit=False)
+
+
+@pytest.fixture(scope="session")
```
zurdi15: We still need migration because the `main.py` migration is not executed when running tests, because `maun.py` is not executed.

zurdi15 (approved):

PR #322, opened 2023-08-05
Offloading sync into worker

### LINKED ISSUE #321: More resilient scanning and socket connection
Currently longer scans can be flaky, with the socket disconnecting/reconnecting mid-scan at times, leaving the UI in a broken state. We should improve both how socket failures are handled and how scans are run.

GAntoine:
This PR introduces [RQ](https://python-rq.org/), a simple job queue and background processor. Backed by Redis, it allows us to use workers to run long tasks, like full library rescans, without having to wait on the main thread.

Since Redis is an all-in-one database, cache, streaming engine, and message broker, we can leverage its power over time to do some of the following:

* Run computationally heavy tasks in parallel
* Temporarily store large amounts ephemeral data
* Cache expensive queries

### What else changed

* Autoscroll to bottom of page when new platform/rom emitted by websocket
* Simplify the socket manager to just export the server and app
* Store the twitch access token in Redis so its accessible across processes
  * Set it to expire in Redis after `expires_in` seconds
* Move all env access to config and export from there

Closes #321


examples/docker-compose.example.yml
```diff
@@ -12,6 +12,7 @@ services:
       - DB_USER=romm-user # [Optional] Only needed if ROMM_DB_DRIVER=mariadb
       - DB_NAME=romm # [Optional] Only needed if ROMM_DB_DRIVER=mariadb. Can be optionally changed, and should match the MYSQL_DATABASE value in the mariadb container.
       - DB_PASSWD=<database password> # [Optional] Only needed if ROMM_DB_DRIVER=mariadb
+      - REDIS_HOST=redis
```
zurdi15, line 15: I don't know if it is a good idea to add any redis things to the docker-compose.

Redis should be under the hood as I understand it, what do you think? Maybe instaling it bare metal in the docker image?

zurdi15 (changes requested):

examples/docker-compose.example.yml
gantoine: That's a very good question. I based the decision on the approach taken by 3 other projects I use, [paperless-ngx](https://github.com/paperless-ngx/paperless-ngx/blob/main/docker/compose/docker-compose.mariadb.yml), [nextcloud](https://hub.docker.com/_/nextcloud/) and [immich](https://github.com/immich-app/immich/releases/download/v1.72.1/docker-compose.yml), which all require running Redis as a separate service. Since I already have Redis running, all my containers share the same instance, saving 3x the memory usage.

And not that StackOverflow is a source of truth, but a number of answers to questions about adding Redis to docker images recommend [running it as a separate service](https://stackoverflow.com/a/42376017).

~What I _can_ do is make Redis a non-required services, and fallback to running the scan task in the main thread. That way anyone who doesn't want to set it up can avoid doing so, and things will keep working for them. Kinda like falling back to using SQLite if you don't want to run a MariaDB container. What do you think?~ Turns out this harder then I thought, so this is most likely a uni-directional change.

zurdi15: Hmm very interesting, I use nextcloud and didn't needed to use redis, maybe they have some kind of callback. 

In any case, I think that the better approach for the future, when we integrate more metadata sources that allow us to use the power of redis, is to keep it as it is now in this PR.

zurdi15 (approved):

PR #327, opened 2023-08-07
Add a png version of the RomM logo

GAntoine:
We'll need this before we submit this project to Unraid Community Apps.


zurdi15:
I guess you just got the svg and converted it into png right?

I have different versions with different resolutions of the logo when I was soing it. I think for the Unraid CA template is better the logo without the name. Let me update it.

GAntoine:
> Let me update it with the directly exported one.

Thank you!

PR #335, opened 2023-08-09
Fix multi rom zipping + streaming

### LINKED ISSUE #332: [Bug] Multifiles download not downloading only eating server RAM
**RomM version**
Version of RomM you are using
1.9.1
**Describe the bug**
When trying to download files that is a multi file selection i don't see any way that the download i starting or even starts creating a zip file. But in my main system i see the docker eating more and more ram. This is for any size of multi file selection. Once i let it be for about 30 min and it had almost all of my 40 GB of server ram allocated to it and still no download completed to my system. Dose not matter the size of the files. Tested with a Gamecube dual disc game so that would be about 3GB in total.

**To Reproduce**
Download more than one file form the same game

**Expected behavior**
To create a zip and then start the download

**Screenshots**

Firefox
Version 116.0.2 (64-bit)
And
Edge
Version 115.0.1901.200 (Official build) (64-bit

**Smartphone (please complete the following information):**

**Additional context**

GAntoine:
Fixes #332


zurdi15:
I need to test this with proper big files. Once tested I merge it!

zurdi15 (approved):

PR #359, opened 2023-09-02
Use pinia to store roms and users

GAntoine:
This PR leverages pinia as a better way to store the state of roms (in the gallery) and users in the frontend. This allows us to remove `refreshView`, as a full-app refresh is no longer necessary to update the client state.

### ROMs

A store for roms was added to hold the collection of ROMs visible in the gallery, and the sub-state of that collection. When a rom is edited or deleted, we update `romsStore` to reflect that change.

The list of filteredRoms, searchRoms and selectedRoms was also moved to the store. Internally, the store maintains those lists as IDs (and not full objects), and filters against `_all` in the getters. This way, when we update the `_all` list, those sub-lists will also be updated, as they're simply filtering against `_all`.

### Users

The users list now uses `/stores/users`, which holds the list of all users. When a user is created, edited, or removed, we update the `usersStore` to reflect that change.

### DIFF
--- a/frontend/src/views/Gallery/Base.vue
+++ b/frontend/src/views/Gallery/Base.vue
@@ -281,13 +281,13 @@ onBeforeRouteUpdate(async (to, _) => {
         <template v-slot:activator="{ props }">
           <v-fab-transition>
             <v-btn
-              v-show="selectedRoms.length > 0"
+              v-show="romsStore._selectedIDs.length > 0"
               color="romm-accent-1"
               v-bind="props"
               elevation="8"
               icon
               size="large"
-              >{{ selectedRoms.length }}</v-btn
+              >{{ romsStore._selectedIDs.length }}</v-btn
             >
           </v-fab-transition>
         </template>
--- a/frontend/src/views/Gallery/Base.vue
+++ b/frontend/src/views/Gallery/Base.vue
@@ -1,7 +1,6 @@
 <script setup>
 import { ref, inject, onMounted, onBeforeUnmount } from "vue";
 import { onBeforeRouteUpdate, useRoute } from "vue-router";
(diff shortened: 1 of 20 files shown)


frontend/src/views/Gallery/Base.vue
```diff
@@ -280,13 +281,13 @@ onBeforeRouteUpdate(async (to, _) => {
         <template v-slot:activator="{ props }">
           <v-fab-transition>
             <v-btn
-              v-show="romsStore.selected.length > 0"
+              v-show="romsStore._selectedIDs.length > 0"
```
gantoine: This has to use the internal list as it ties into the v-model of the data-table

zurdi15:
I didn't researched too much why, but going to the detailed view of a rom and going back again in the browser makes the rom list duplicated.

frontend/src/views/Gallery/Base.vue
```diff
@@ -161,37 +165,35 @@ function selectRom({ event, index, selected }) {
...
   } else {
-    romsStore.updateLastSelectedRom(index);
+    romsStore.updateLastSelected(index);
   }
-  emitter.emit("refreshSelected");
 }
 
 onMounted(async () => {
-  fetchRoms(route.params.platform);
+  if(filteredRoms.value.length == 0){
```
zurdi15: Duplicates issue fixed by checking if the roms store already has roms.

frontend/src/views/Gallery/Base.vue
zurdi15: Fixed the data-table  sync issue with the grid view changing the item-value of the data-table to the rom `id`

zurdi15:
Both issues fixed. Take a look at the code and test it when you can!

PR #361, opened 2023-09-03
Upload roms and add custom cover

zurdi15:
This PR includes one of the main features of RomM to be a true rom manager: `uploading` roms directly from the UI. The option to edit the `cover` for a rom is finally enabled. 

This PR also includes a more user friendly way to `multi select` roms in a `touch screen` as a side feature.

The upload button was moved from the `drawer` to the `gallery bar`.

Roms can be uploaded as a `single` file or in `bulk` to be more efficient when uploading a set of roms.

The logic for editing a rom is now clearer and cleanded a lot of unused parameters to make requests ligther. The code is also more mainatable now.

---

Edit: This PR was merged with the [pinia-stores-for-roms-users](https://github.com/zurdi15/romm/tree/pinia-stores-for-roms-users) branch and [#359](https://github.com/zurdi15/romm/pull/359) PR:

This PR leverages pinia as a better way to store the state of roms (in the gallery) and users in the frontend. This allows us to remove `refreshView`, as a full-app refresh is no longer necessary to update the client state.

### ROMs

A store for roms was added to hold the collection of ROMs visible in the gallery, and the sub-state of that collection. When a rom is edited or deleted, we update `romsStore` to reflect that change.

The list of filteredRoms, searchRoms and selectedRoms was also moved to the store. Internally, the store maintains those lists as IDs (and not full objects), and filters against `_all` in the getters. This way, when we update the `_all` list, those sub-lists will also be updated, as they're simply filtering against `_all`.

### Users

The users list now uses `/stores/users`, which holds the list of all users. When a user is created, edited, or removed, we update the `usersStore` to reflect that change.


zurdi15:
I forgot to adap the `edit` rom test. Fiing it.

zurdi15:
I also somehow broke the `search IGDB` feature, it no longer updates the rom to the selected one.

backend/endpoints/identity.py
```diff
@@ -124,7 +124,7 @@ def __init__(
         self.avatar = avatar
 
 
-@protected_route(router.put, "/users/{user_id}", ["users.write"])
+@protected_route(router.patch, "/users/{user_id}", ["users.write"])
```
gantoine, line 127: Double-check that this stilll works case put and patch expect different data methods (body, form or query params)

backend/endpoints/rom.py
```diff
@@ -73,6 +83,21 @@ def rom(request: Request, id: int) -> RomSchema:
...
+def upload_roms(request: Request, p_slug: str, roms: list[UploadFile] = File(...)):
+    platform_fs_slug = dbh.get_platform(p_slug).fs_slug
+    log.info(f"Uploading files to: {platform_fs_slug}")
+    if roms is not None:
+        roms_path = build_upload_roms_path(platform_fs_slug)
+        for rom in roms:
+            log.info(f" - Uploading {rom.filename}")
+            file_location = f"{roms_path}/{rom.filename}"
+            with open(file_location, "wb+") as file_object:
+                file_object.write(rom.file.read())
```
gantoine: What happens if there's a name collision? This would overwrite the contents of the file right?

Also how does this handle uploading large files? Is the entire content read into memory before being written?

backend/endpoints/rom.py
```diff
@@ -129,46 +154,73 @@ def roms(
...
-    updated_rom: dict = data["updatedRom"]
+    data = await request.form()
+    cleaned_data = {}
+    cleaned_data["r_igdb_id"] = data["r_igdb_id"]
+    cleaned_data["r_name"] = data["r_name"]
+    cleaned_data["r_slug"] = data["r_slug"]
+    cleaned_data["file_name"] = data["file_name"]
+    cleaned_data["url_cover"] = data["url_cover"]
+    cleaned_data["summary"] = data["summary"]
+    cleaned_data["url_screenshots"] = json.loads(data["url_screenshots"])
```
gantoine, line 184: 👍🏼

frontend/src/components/Dialog/Rom/UploadRom.vue
```diff
@@ -0,0 +1,146 @@
...
+
+socket.on("scan:done_ko", (msg) => {
+  scanning.set(false);
+  emitter.emit("snackbarShow", {
+    msg: `Scan couldn't be completed. Something went wrong: ${msg}`,
+    icon: "mdi-close-circle",
+    color: "red",
+  });
+  socket.disconnect();
+});
```
gantoine, line 39: Don't forget to `socket.off` when the component is unmounted.

frontend/src/components/GalleryAppBar/ScanBtn.vue
```diff
@@ -0,0 +1,59 @@
...
+    icon: "mdi-loading mdi-spin",
+    color: "romm-accent-1",
+  });
+
+  if (!socket.connected) socket.connect();
+  socket.emit("scan", {
+    platforms: [route.params.platform],
+    rescan: false,
+  });
+}
```
gantoine, line 47: Same goes here `socket.off`

frontend/src/components/Game/Card/Base.vue
```diff
@@ -1,39 +1,28 @@
...
 import ActionBar from "@/components/Game/Card/ActionBar.vue";
 import Cover from "@/components/Game/Card/Cover.vue";
 
 // Props
-const props = defineProps(["rom", "index"]);
+const props = defineProps(["rom", "index", "selected"]);
 const emit = defineEmits(["selectRom"]);
-const romsStore = useRomsStore();
-const selected = ref();
+const romsStore = storeRoms();
```
gantoine, line 9: You sure we don't need `storeToRefs` here to keep this reactive? https://pinia.vuejs.org/core-concepts/#Destructuring-from-a-Store

frontend/src/services/api.js
```diff
@@ -71,24 +71,42 @@ export async function downloadRomApi(rom, files) {
...
-    file_name: renameAsIGDB
-      ? rom.file_name.replace(rom.file_name_no_tags, updatedData.r_name)
-      : updatedData.file_name,
-  };
-
-  return api.patch(`/platforms/${rom.p_slug}/roms/${rom.id}`, {
-    updatedRom,
+export async function uploadRomsApi(romsToUpload, platform) {
+  let formData = new FormData();
+  romsToUpload.map((rom) => formData.append("roms", rom));
```
gantoine:
```suggestion
  romsToUpload.forEach((rom) => formData.append("roms", rom));
```

GAntoine (changes requested):

backend/endpoints/identity.py
zurdi15: yes! it still works, changed to unify with `update_rom` endpoint

backend/endpoints/rom.py
zurdi15: Good ones, need to check

frontend/src/components/Game/Card/Base.vue
zurdi15: Yeah it works for sure, we don't need it to keep the reactivity in the way we are using it.

Edit: nevermind, I think you are totally right. I am readding it

zurdi15: Re-added and working!

backend/endpoints/rom.py
zurdi15: > What happens if there's a name collision? This would overwrite the contents of the file right?

Now when uploading files the enpoint checks and raises an exception if any exists
 
> Also how does this handle uploading large files? Is the entire content read into memory before being written?

Now the files are read by chunks of 1024bytes

backend/endpoints/rom.py
```diff
@@ -89,11 +89,22 @@ def upload_roms(request: Request, p_slug: str, roms: list[UploadFile] = File(...
     log.info(f"Uploading files to: {platform_fs_slug}")
     if roms is not None:
         roms_path = build_upload_roms_path(platform_fs_slug)
+        for rom in roms: #TODO: Refactor code to avoid double loop
```
gantoine, line 92: This is fine and consistent for now, but you're right we should refactor it at some point to fail gracefully.

GAntoine (approved):

PR #362, opened 2023-09-06
feat(pipeline): every branch's docker image is now built

XargonWan:
Every branch now produces an image tagged `dev-branchname` on PR and push.

If this is accepted we have to check how the images appears on the registry, evenutally fix the tags if we don't like it and decide how to clean up the older builds or simply replace them with the new one.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,10 @@
 name: build romm image
 
 on:
+
+  push:
+  pull_request:
+
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


GAntoine:
Copying my discord comments here:

> honestly i don't love the idea of building images for each branch, even in a case like this
imo if someone wants to test a feature that so early in development they should setup the dev environment and run it locally
when we feel a feature is ready for alpha we can manually create a tag for it

PR #365, opened 2023-09-14
Models and API refactor

### LINKED ISSUE #374: [Bug] Some games with a / in their title fail to pull IGDB metadata
**RomM version**
dev-2.0.0-rc-4

**Describe the bug**
I've noticed that some games that have a / in their title will be able to be found in search, but then fail to pull the data, with a blank error and a red x icon showing in a notification at the top.

This seems to be because the filesystem can't support file names with a / in them, so it can't rename them. If the option to rename the file is unchecked, it succeeds. Maybe it would make sense to have it check for files with - instead of / when matching to IGDB names?

**To Reproduce**
1. Go to any game in the library that should have a / in its name (ex. Kingdom Hearts 358/2 Days)
2. Click on '...'
3. Attempt to search for the game on IGDB
4. With rename file checked, hit apply.

**Expected behavior**
The error should definitely have some explanatory text- but also in general it should hopefully be able to handle these kinds of games in a way that prevents it from erroring out.

 - OS: Debian
 - Browser: Chrome
 - Version: 117

GAntoine:
This is something I've been meaning to do for a while, with the hopes of improving the clarity of some of our models and the relations between them.

# ORM

A few fields were deprecated (`p_name, p_igdb_id, p_sgdb_id`) and a couple more renamed:

* `r_igdb_id` -> `igdb_id`
* `r_sgdb_id` -> `sgdb_id`
* `r_slug` -> `slug`
* `r_name` -> `name`
* `p_slug` -> `plaforms_slug`
* `n_roms` -> `rom_count`

The PRIMARY key for `Platform` was moved from `fs_slug` to `slug`, to clarify that `slug` is the true identifier for the platform.

* New relationship between `Platform` and `Rom` in the ORM eager joins platform
  * Example: `platform_name = rom.platform.name`
* Adds a new decorator `begin_session` that begins sessions and catches errors
* Translate methods to use 2.x sqlalchemy syntax

### API

* Now exports a single object by default with all methods on it
  * Means api is now called like so: `api.deleteRom({}).then(...)`
* All api methods now accept objects as params
  * No need to worry about order of params
* Move some values to url params, form inputs or data where appropriate

### Other changes

* Update fastapi + pylance to the latest version
* Improved error handling for file upload
  * Also cleaned up double-loop
* Changed most `Rom` api paths to no longer start with `/platform/{p_slug}`
* Rom deletion endpoint accepts `delete_from_fs` boolean
* `rom_exists` -> `get_rom_by_filename` and returns the `Rom` (or `None`)
* Give all endpoints explicit response types
* Fixes #374


backend/alembic/env.py
```diff
@@ -47,6 +47,7 @@ def run_migrations_offline() -> None:
         render_as_batch=True,
         literal_binds=True,
         dialect_opts={"paramstyle": "named"},
+        compare_type=True,
```
gantoine, line 50: This is needed to change field types (str -> int in our case)

backend/alembic/versions/0009_models_refactor.py
```diff
@@ -0,0 +1,150 @@
+"""empty message
+
+Revision ID: 0009_models_refactor
+Revises: 1.9.2
```
zurdi15: This should be 2.0.0 now.

backend/alembic/versions/0009_models_refactor.py
```diff
@@ -0,0 +1,150 @@
...
+Create Date: 2023-09-12 18:18:27.158732
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import mysql
+
+# revision identifiers, used by Alembic.
+revision = "0009_models_refactor"
+down_revision = "1.9.2"
```
zurdi15: Same here, now it's 2.0.0

backend/handler/igdb_handler.py
```diff
@@ -38,6 +39,20 @@
 MAME_XML_FILE: Final = os.path.join(os.path.dirname(__file__), "fixtures", "mame.xml")
 
 
+class IGDBPlatformType(TypedDict):
+    igdb_id: str
+    name: str
+
+
+class IGDBRomType(TypedDict):
+    igdb_id: str
```
zurdi15, line 53: This should be int?

backend/handler/igdb_handler.py
```diff
@@ -38,6 +39,20 @@
 MAME_XML_FILE: Final = os.path.join(os.path.dirname(__file__), "fixtures", "mame.xml")
 
 
+class IGDBPlatformType(TypedDict):
+    igdb_id: str
```
zurdi15, line 48: Same on this, if web converted id's to int in db, shouldd this be int?

backend/handler/igdb_handler.py
```diff
@@ -137,33 +154,31 @@ def _search_screenshots(self, rom_id: int) -> list:
...
             self.platform_url,
-            data=f'fields id, name; where slug="{p_slug.lower()}";',
+            data=f'fields id, name; where slug="{slug.lower()}";',
         )
 
-        platform = pydash.get(paltforms, "[0]", None)
+        platform = pydash.get(platforms, "[0]", None)
         if not platform:
             return {
                 "igdb_id": "",
```
zurdi15: Herr maybe returning a `None`?

backend/handler/igdb_handler.py
zurdi15, line 48:
```suggestion
    igdb_id: int
```

backend/handler/igdb_handler.py
zurdi15, line 53:
```suggestion
    igdb_id: int
```

backend/handler/igdb_handler.py
```diff
@@ -137,33 +154,31 @@ def _search_screenshots(self, rom_id: int) -> list:
...
-                "name": p_slug,
-                "slug": p_slug,
+                "name": slug,
             }
 
         return {
             "igdb_id": platform["id"],
             "name": platform["name"],
-            "slug": p_slug,
         }
```
zurdi15:
```
         return IGDBPlatformType(
                igdb_id=None,
                name=slug
        )

return IGDBPlatformType(
    igdb_id=platform["id"],
    name=platform["name"]
)
```

backend/handler/igdb_handler.py
```diff
@@ -220,61 +235,65 @@ async def get_rom(self, file_name: str, p_igdb_id: int):
...
-            "r_name": rom.get("name", ""),
+            "igdb_id": igdb_id,
+            "slug": rom.get("slug", ""),
+            "name": rom.get("name", ""),
             "summary": rom.get("summary", ""),
-            "url_cover": self._search_cover(r_igdb_id),
-            "url_screenshots": self._search_screenshots(r_igdb_id),
+            "url_cover": self._search_cover(igdb_id),
+            "url_screenshots": self._search_screenshots(igdb_id),
         }
```
zurdi15, line 312:
```
return IGDBRomType(
            igdb_id=igdb_id,
            slug=rom.get("slug", ""),
            name=rom.get("name", ""),
            summary=rom.get("summary", ""),
            url_cover=self._search_cover(igdb_id),
            url_screenshots=self._search_screenshots(igdb_id),
        )
```

backend/handler/igdb_handler.py
```diff
@@ -285,9 +304,9 @@ def get_matched_roms_by_name(self, search_term: str, p_igdb_id: int):
                     "t_thumb", "t_cover_big"
                 ),
                 url_screenshots=self._search_screenshots(rom["id"]),
-                r_igdb_id=rom.pop("id"),
-                r_slug=rom.pop("slug"),
-                r_name=rom.pop("name"),
+                igdb_id=rom.pop("id"),
```
zurdi15: Same here, should we return `IGDBRomType` instead of dict to match notation?

zurdi15:
Everything seems really good, just some comments about type notations, but nothing important!

backend/handler/igdb_handler.py
gantoine: The response we get from IGDB is a string (I think), and we may pass an empty string if it's not found. The conversion happens later in the code flow.

backend/handler/igdb_handler.py
gantoine: Same as above

backend/handler/igdb_handler.py
gantoine: What do you mean? I think this is just a whitespace change.

backend/handler/igdb_handler.py
gantoine: I like this, will implement tomorrow 👍🏼

backend/handler/igdb_handler.py
zurdi15: Oh maybe I misstarget the line, I meant the IGDB I'd because I was with the idea of returning it as int, so it had more sense to return a None instead of a empty string, but if we are going to stick with strings then the empty string is fine

zurdi15 (approved):

PR #373, opened 2023-09-17
Additional platform icons!

Casuallynoted:
Added:
CD-i
3DO
Neo Geo Pocket Color
Nintendo 64DD
Satellaview (More notes about this in the swapped section)
Playdia
Pippin
Mac

Renamed (Previously weren't using IGDB naming conventions so wouldn't appear when in use):
Neo Geo CD
Neo Geo AES
Neo GEO MVS

Swapped:
SNES and SNES-Alt (Using the japanese SNES controller icon for the Satellaview and setting the US SNES icon as the default for the main console, figured it makes more sense!)

Also added accreditation for the icon used for Mac to the readme.


GAntoine (dismissed):

Casuallynoted:
Whoops, just realised that the sega pico icon needed a rename for IGDB consistency. Done! I think this should be all that's needed :)

GAntoine (approved):

PR #383, opened 2023-09-21
[ROMM-378] Fixes to search and pinia store

### LINKED ISSUE #378: [Bug] Searching for a game only returns games that have been seen in the interface
**RomM version**
2.0rc5

**Describe the bug**
When searching for a game in a big library, it only seems to search the game that have currently been scrolled through. So for example, when first loading my Xbox library if I search for Tetris, it says no results found. However, if I scroll down to where Tetris is in the games list, then I am able to search for it.

**To Reproduce**
1. Open a platform's library
2. Type the name of a game that is toward the end of the list into the search box
3. Game will not be found

**Expected behavior**
Ideally, it should be able to pull games regardless of whether the UI has loaded them- this would definitely make navigating larger libraries a lot easier.

 - OS: Debian
 - Browser: Chrome
 - Version: 117

GAntoine:
So this is a little cleanup to the way we deal with the pinia store for roms. Essentially, anytime a change is made to the `_all` rom list, we reorder the list and remove any duplicates. This is just to make sure the list is in a healthy state in all cases.

The real fix here is adding any search results to the `_all` roms list. This way, when we set the list of filtered IDs in the `romStore`, we have the results to pull from.

Fixed #378

### DIFF
--- a/frontend/src/stores/roms.js
+++ b/frontend/src/stores/roms.js
@@ -1,3 +1,4 @@
+import { uniqBy } from "lodash";
 import { defineStore } from "pinia";
 
 export default defineStore("roms", {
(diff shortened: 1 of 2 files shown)


frontend/src/stores/roms.js
```diff
@@ -23,20 +24,27 @@ export default defineStore("roms", {
   },
 
   actions: {
+    _reorder() {
+      // Sort roms by name and remove duplicates
+      this._all = uniqBy(
+        this._all.sort((a, b) =>
+          a.file_name_no_tags.localeCompare(b.file_name_no_tags)
```
zurdi15, line 31: This won't deduplicate roms with the same name but differente regions for example? (with just different tags in the filename)

zurdi15 (approved):

frontend/src/stores/roms.js
gantoine: Technically yes and we should probably change it to use `file_name`, though as-is it's already better then before.

PR #397, opened 2023-10-13
Change IGDB ID/SECRET variables to working names

Jacksaur:
Current variables do not work, as mentioned here: https://github.com/zurdi15/romm/issues/394#issuecomment-1751814068

As you say, the keys will work on dev-latest. But the example file doesn't pull dev-latest, so it'd make more sense for new users if the example file used the same conventions as the release it currently pulls for the time being.

### DIFF
--- a/examples/docker-compose.example.yml
+++ b/examples/docker-compose.example.yml
@@ -14,8 +14,8 @@ services:
       - DB_NAME=romm # Should match the MYSQL_DATABASE value in the mariadb container
       - DB_PASSWD=<database password>
       # [Optional] Used to fetch metadata from IGDB
-      - IGDB_CLIENT_ID=<IGDB client id>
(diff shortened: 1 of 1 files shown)


GAntoine:
I'm happy merging this change in, as the renaming is backwards-compatible for `dev-latest` users. However the PR is open against `master`, and the main branch for the repo is `release`. Do you think it makes more sense to open this PR against release?

Jacksaur:
Honestly, I have minimal experience with Github and didn't even check the branch I was merging with. Apologies.
Is there a way I could move this PR to the correct one?

GAntoine:
You know what it's all good, I'll merge this is and merge master into the release branch, so it shows up for everyone. Thanks for the PR!

GAntoine (approved):

PR #401, opened 2023-10-22
general docker image optimization

psych0d0g:
for kubernetes deployments and security minded folks its prefered to run containers with a readonly rootfs.
this patch changes nginx configuration in a way that it is possible to run it on such readonly instances.
(mounted volumes are by design exempt from this)

other things i changed are:
- the default listen port for nginx to not be port 80 (since that port can not be bound without root permissions)
- the connection between uvicorn and nginx was switched from using TCP to unix domain socket
- alpine base image instead of ubuntu to shave a few mb from the docker container size
- added cache cleaning for poetry
- squashed docker layers down to one as last build step to reduce final image size ( 112.32MiB   from  260.36MiB in the beginning )

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,7 +6,7 @@ RUN npm install
 RUN npm run build
 
 # Setup frontend
-FROM ubuntu/nginx:1.18-22.04_edge as production-stage
+FROM nginx:1.24-alpine3.17 as production-stage
 ARG WEBSERVER_FOLDER=/var/www/html
 COPY --from=front-build-stage /front/dist ${WEBSERVER_FOLDER}
 COPY ./frontend/assets/default_avatar.png ${WEBSERVER_FOLDER}/assets/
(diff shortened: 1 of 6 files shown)


psych0d0g:
i made further changes in a new branch that im holding back for now, but plan to open a pull request once this one is merged, mainly further docker-file refactoring, and updating of the poetry.lock file :)

docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN npm install
 RUN npm run build
 
 # Setup frontend
-FROM ubuntu/nginx:1.18-22.04_edge as production-stage
+FROM nginx:1.24-alpine3.17-slim as production-stage
```
gantoine, line 9: 👍🏼

docker/Dockerfile
```diff
@@ -16,31 +16,44 @@ RUN mkdir -p ${WEBSERVER_FOLDER}/assets/romm && \
...
+        mariadb-connector-c-dev \
+        python3-dev \
+        py3-pip
+
+FROM scratch
+COPY --from=production-stage / /
+
 # Expose ports and start
-EXPOSE 80
+EXPOSE 8080
```
gantoine, line 57: this is a breaking change right?

GAntoine (approved):
i'm all for this change 👍🏼

docker/Dockerfile
psych0d0g: at least for the compose and such, i think i missed changing the target port in the example, the ports would need to be changed to 80:8080 at least so that the user facing frontend can still be reached by http://localhost for example

zurdi15 (approved):

PR #402, opened 2023-10-23
add venv to docker and refactor

psych0d0g:
introduce python virtual environments so that project dependencies dont clash with the OS shipped python libraries.
As an example poetry tried to update `packaging` which is not possible since its usually shipped with the used OS distro and can not be managed by pip or poetry if installed in the OS, with venvs that limitation is solved and said package is able to be managed by the project.
with a venv we are indenpendent from what the OS shipps, even in core libraries and can update them however we want :)

i also added some meaningful comments to the dockerfile to explain what is happening

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,14 +19,9 @@ RUN mkdir -p ${WEBSERVER_FOLDER}/assets/romm && \
 RUN apk add --upgrade \
         bash \
         curl \
-        gcc \
-        musl-dev \
         netcat-openbsd \
         mariadb-connector-c \
-        mariadb-connector-c-dev \
-        python3 \
-        python3-dev \
-        py3-pip
+        python3
 
 # Install additional build dependencies
 RUN apk add --upgrade \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,15 +19,17 @@ RUN mkdir -p ${WEBSERVER_FOLDER}/assets/romm && \
 RUN apk add --upgrade \
         bash \
         curl \
-        netcat-openbsd \
+        libffi \
         mariadb-connector-c \
+        netcat-openbsd \
         python3
 
 # Install additional build dependencies
 RUN apk add --upgrade \
         gcc \
-        musl-dev \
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -38,12 +34,18 @@ passlib = {extras = ["bcrypt"], version = "^1.7.4"}
...
 python-multipart = "^0.0.6"
 types-python-jose = "^3.3.4.8"
 types-passlib = "^1.7.7.13"
 
+[tool.poetry.group.test.dependencies]
+pytest = "^7.4.0"
+pytest-env = "^0.8.2"
+pytest-vcr = "^1.0.2"
+pytest-mock = "^3.11.1"
+pytest-asyncio = "^0.21.1"
```
gantoine, line 47: 👍🏼

docker/Dockerfile
```diff
@@ -29,31 +28,69 @@ RUN apk update && \
...
+    /installed_poetry_requirements.txt \
+    /build_requirements.txt \
+    /docker-entrypoint.sh \
+    /docker-entrypoint.d
+
+# Move everything we prepared over to our final docker image
 FROM scratch
+LABEL org.opencontainers.image.title="zurdi15/romm" \
+      org.opencontainers.image.description="RomM (stands for Rom Manager) is a game library manager focused in retro gaming. Manage and organize all of your games from a web browser" \
+      org.opencontainers.image.licenses="GPL-3.0"
```
gantoine: 👍🏼

GAntoine (approved):
yeah really solid 👍🏼

docker/Dockerfile
Djiit: FWIW, the docker metadata action in the `build` workflow should be taking care of title and description labels automatically. Maybe add the `description` label [here](https://github.com/zurdi15/romm/blob/release/.github/workflows/build.yml#L40C10-L40C10) too ?

psych0d0g: moved the labels into the git workflow file as suggested

zurdi15 (approved):

PR #413, opened 2023-10-29
[ROMM-412] Support for secured redis added

### LINKED ISSUE #412: [Feature] Redis Password Option
Currently unable to use Redis if Redis is using a password.

A REDIS_PASSWORD environmental variable.

Being able to use password@host for the REDIS_HOST environmental variable would also work.

**Additional context**
I was referred here from Discord. Thanks!

zurdi15:
Added new `REDIS_PASSWORD` environment variable to support secured redis with pasword.

Closes #412


backend/config/__init__.py
```diff
@@ -43,6 +43,7 @@
 )
 REDIS_HOST: Final = os.environ.get("REDIS_HOST", "localhost")
 REDIS_PORT: Final = os.environ.get("REDIS_PORT", "6379")
+REDIS_PASSWORD: Final = os.environ.get("REDIS_PASSWORD")
```
gantoine, line 46:
```suggestion
REDIS_PASSWORD: Final = os.environ.get("REDIS_PASSWORD", None)
```

Or maybe it should fallback to `""`? Whichever one will let the no-password option continue to work.

docker-compose.yml
```diff
@@ -15,6 +15,16 @@ services:
...
     image: redis:alpine
     container_name: redis
+    command:
+    # - /bin/sh
+    # - -c
+    # # - Double dollars, so that the variable is not expanded by Docker Compose
+    # # - Surround by quotes, so that the shell does not split the password
+    # # - The ${variable:?message} syntax causes shell to exit with a non-zero
+    # #   code and print a message, when the variable is not set or empty
+    # - redis-server --requirepass "$${REDIS_PASSWORD:?REDIS_PASSWORD variable is not set}"
```
gantoine, line 25: Does all this need to be here if it's commented out?

env.template
```diff
@@ -23,6 +23,7 @@ DB_ROOT_PASSWD=
 REDIS_HOST=127.0.0.1
 REDIS_PORT=6379
 ENABLE_EXPERIMENTAL_REDIS=true
+#REDIS_PASSWORD=admin_redis
```
gantoine, line 26:
```suggestion
REDIS_PASSWORD=
```

GAntoine (changes requested):

backend/config/__init__.py
zurdi15: By default `os.environ.get()` returns a `None` if the variable doesn't exists, and the no-password option will work only with `None`. We can add it as fallback if you want for clearer reading

docker-compose.yml
zurdi15: I leaved it there just in case we want to test something with a pasworded redis.

env.template
zurdi15: Same with this, by default I leaved the docker-compose for the dev team with non password redis, so I leaved this commentes aswel

GAntoine (approved):

PR #420, opened 2023-10-31
Fix nginx max body size

zurdi15:
This PR fixes the ``max_body_size`` parameter of the internal nginx server to make it unlimited.
Also gives execution permissions to the scheduler and watchdog init scripts


GAntoine (approved):
nice

PR #425, opened 2023-10-31
Update env variables to include new tasks

GAntoine:
* Add the new env variables to the docker-compose example
* Change mame XML update to run at 5am by default


zurdi15:
Oh wow, how we forgot this!

zurdi15 (approved):

PR #445, opened 2023-11-05
Add check for productID in file name

GAntoine:
In #350 it was clarified that sometimes the filename will contain the `productID`, and not the `titleID`. 

* Map the `productID` to the `titleID` during async task and save result to JSON file
* Match against product ID format during scan/rescan
* Less aggressive check for multiple file extensions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,4 +55,5 @@ frontend/dev-dist
 
 # outside data
 switch_titledb.json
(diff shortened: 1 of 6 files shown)


zurdi15 (approved):

KillahB33:
Amazing! Thanks for the quick work all!!

PR #446, opened 2023-11-05
[FEATURE] add helm chart to easily deploy RomM to kubernetes

psych0d0g:
Add a helm chart so users that are running kubernetes can deploy RomM quick and easily
this is the initial working version, currently it has a few limitations:
- it can only run with redis and an already pre-existing mariadb deployment or sqlite
- if one dosnt have a pre-existing mariadb instance, redis either needs to be disabled for mariadb to be spawned with the romm deployment automatically (untested right now) or use sqlite together with redis instead.
- mariadb spawning is prepared but currently the used bitnami subcharts have a bug that prevents spawning mariadb if redis is already integrated and vice versa, so at the moment only sqlite is supported when redis is enabled.

the chart will bootstrap redis in standalone mode by default but can be set to spawn a replicating redis cluster aswell.

still missing is the integration with the github pages of RomM so that a chart repository is available to easily pull the chart using the helm command, it would be desirable to integrate "chart-releaser" into the github pages builder pipeline:
https://github.com/helm/chart-releaser

until we have a helm repo setup, one could test the chart by checking out the git branch, cd into the `/helm/romm` directory, edit the values.yaml to have the desired config, and run:
```console
helm dependencies update
helm -n test install romm .
```


psych0d0g:
please see https://github.com/psych0d0g/helm-charts/tree/main for further development for now

zurdi15:
Wow dude, this is very much beyond my knowledge x)

PR #463, opened 2023-11-14
Nginx logformat touchup

psych0d0g:
bring nginx's accesslog in line with the overall romm log format


zurdi15 (approved):

zurdi15:
This one is very satisfying

PR #487, opened 2023-12-05
[ROMM-486] Titleize slug as name when scanning platform

### LINKED ISSUE #486: [Other] Add information about custom platforms to the wiki
Since not all platforms will be covered by IGDB's platform list, it might be good to have information on the wiki about how to name folders and icons for custom platforms in the event users want to create them, especially as far as things like capital letters and spaces in platform names are concerned. 

For example, if I want to add the Konami Picno, and I create a folder called "picno", it shows up in the sidebar without the first letter capitalised. If I capitalise the letter, the sidebar looks correct but the icon doesn't work. Likewise, if I create the folder pocket-challenge-v2, it uses the dashes rather than the spaces to spell out the proper name, Pocket Challenge V2. I have a bit of understanding of how to properly name these but it's not very comprehensive, so I figured it might be worth having that information documented.

gantoine:
If the platform isn't found in IGDB, set the name as a titleized form of the slug (`pocket-challenge-v2` -> `Pocket Challenge V2`)

@zurdi15 This will only work for new platforms going forward as it needs a full rescan. Also does this change make it confusing about which platforms were not found in IGDB?

Closes #486 

Screenshot 2023-12-05 at 5 52 51 PM

### DIFF
--- a/backend/handler/igdb_handler.py
+++ b/backend/handler/igdb_handler.py
@@ -171,7 +171,7 @@ class IGDBHandler:
 
         platform = pydash.get(platforms, "[0]", None)
         if not platform:
-            return IGDBPlatformType(igdb_id=None, name=slug)
+            return IGDBPlatformType(igdb_id=None, name=slug.replace("-", " ").title())
 
         return IGDBPlatformType(
             igdb_id=platform["id"],


Casuallynoted:
One quick question about this! For platforms such as the e-Reader, which have a dash and start with a lowercase letter, would there be a way to specify formatting in the slug? No worries if not but figured I should ask!

gantoine:
Hah good question! No not in this case, it'll be formatted as `E Reader` which is less then ideal, but better then what we have today.

zurdi15:
> @zurdi15 This will only work for new platforms going forward as it needs a full rescan. Also does this change make it confusing about which platforms were not found in IGDB?

I don't think this is going to be an issue, but maybe we should add some kind of icon/warning to the plarform element in the drawer or maybe in the gallery? Just thinking a better way to let the user know that it is not be found in IGDB

gantoine:
@zurdi15 Made the change, and I added a screenshot of what it'll look like in the sidebar (on hover).

PR #506, opened 2023-12-22
[ROMM-501] Fix scan with folder overrides

gantoine:
The scan backend expects `fs_slug`, so pass that in when scanning in the `Scan` view.


zurdi15 (approved):
Oh wow, this have been here for all this time?

Good fix anyway!

DragRedSim:
Thanks for actioning a fix on this so quickly!

PR #510, opened 2023-12-23
Config file error management improved

zurdi15:
Now the error message when users bind the `config.yml` file without create it before setting up RomM is clearer.

Also introduced little front tweaks.

### DIFF
--- a/backend/config/config_loader.py
+++ b/backend/config/config_loader.py
@@ -21,6 +21,9 @@ from logger.logger import log
 class ConfigLoader:
     # Tests require custom config path
     def __init__(self, config_path: str = ROMM_USER_CONFIG_PATH):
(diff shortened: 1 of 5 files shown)


backend/config/config_loader.py
```diff
@@ -21,6 +21,9 @@
 class ConfigLoader:
     # Tests require custom config path
     def __init__(self, config_path: str = ROMM_USER_CONFIG_PATH):
+        if os.path.isdir(config_path):
+            log.critical(f"Your config file {config_path} is a directory, not a file. Docker creates folders by default for binded files that doesn't exists in advance in the host system.")
+            raise FileNotFoundError()
```
gantoine, line 26: 👍🏼

gantoine (approved):

PR #513, opened 2023-12-23
[Feature] RunR Integration - Allow games to be played

thorin8k:
Hi! 

i've been working in a project to encapsulate different JS Rom executors as an separate container called RunR.

This PR adds an integration between RomM and RunR allowing games to be launched from RomM with a Play button in the game pages for configured plaforms.

image

Pressing it will open a new tab with RunR to play the game:

image

The code of runr is available here: https://github.com/thorin8k/runr

I plan to include a lot of new features and i'm open to any proposals that you have :smile: 

If this cannot be merged i am currently using it by a Bookmarklet but i think this will be a nice addition for RomM without needing to overcomplicate its codebase.


gantoine:
Ok so first of all, I love that you started work on this, it's super cool! 🤩  However I'm torn on the idea of having the ability to play games rely on another project and container.

On the one hand, I'm all for separation of concerns and having different projects specialize on different parts, like the conversations we've had about .DAT files and hash matching. While we could built .DAT support into RomM, we feel the complexity of the work is best left to others more familiar with those things.

On the other, I don't feel like playing games in browser is so complex that users need to run an entire other container to do so. We have an [experimental branch](https://github.com/zurdi15/romm/tree/emulator-js) in the works that aims to integrate EmulatorJS directly in RomM. And with the upcoming addition of save/state/bios file support, I plan to work on that integration once that's merged in.

All that being said, if you're really keen on having your games playable in-browser, I would very much welcome any work on a native (as in within this code base) emulator-js integration, whether now or after the save/state/bios file support has been added. And I want to clarify that these opinions are my own, and that @zurdi15 might have different thoughts about this.

thorin8k:
Hi! Thanks for the feedback, I appreciate it a lot.

When I started, I thought about including this directly in the RomM codebase. I wasn't aware of your efforts on this 😅.

After a while, playing around with several implementations, I went for the easiest one. 

Said that it's a first version, and I understand your concerns about relying on an external project.


Maybe this can be a temporal thing until the final internal implementation is completed. I'm happy with that 😇, and I'm also happy to help in that implementation too.


Again, thanks. I'll join the discord so we can chat.

Regarding the PR it's failing because I forgot to include the tests for the new env vars, I'll fix it ASAP.

gantoine:
> Maybe this can be a temporal thing until the final internal implementation is completed.

I think I'll leave that decision up to zurdi; he may have some thoughts about the consistency of the experience. Regardless it would be optional to use, and the amount of code you're adding here is minimal, so there would be no difficulty in eventually switching to something build-in.

thorin8k:
Yeah, no problem at all. 


> Regardless it would be optional to use, and the amount of code you're adding here is minimal, so there would be no difficulty in eventually switching to something build-in.

My intention was exactly that, I didn't want to add a lot of complexity here.


I joined Discord, so we can chat about it 😇

Also, Merry Christmas 🙂

zurdi15:
Merry Christmas @thorin8k! 

First of all, I wanted to congratulate you for your wesome work with RunR.

From my perspective, I totally agree with @gantoine in every bit, we can add this as an ``experimental`` feature, but I belive that the user experience and the consistency with the whole ecosystem and the saves/states will be better with a built-in feature, integrating ``emulatorjs`` (or similar) directly like @gantoine did in [this branch](https://github.com/zurdi15/romm/tree/emulator-js).

That said, once I tested this properly we can merge so users that want to try this ``experimental`` feature, enabling it throught an ``environment variable``, something like ``RUNR_INTEGRATION=true``, like we did with the ``redis`` dependant features.

gantoine:
> enabling it throught an environment variable, something like RUNR_INTEGRATION=true

I'm good with that 👍🏼

thorin8k:
Hi Zurdi!

Thanks, i apreciate the feedback :smile: 


> integrating emulatorjs (or similar) directly like @gantoine did in [this branch](https://github.com/zurdi15/romm/tree/emulator-js).

After seen what he did in this branch, i totally agree with you guys. This fits better in the project. With this, i've been thinking in other aproach to integrate RunR as an external tool and allow other tools to be integrated as well. 

The thing is that we can add a link system, something similar to what sonarr,radarr and the other *arrs have, to link external platforms like imbd,tvdb and so.

image

If this links could be customized by the user, we could integrate this and other platforms with ease.

Maybe i can work in a draft and lets see how it goes? 😇

zurdi15:
> The thing is that we can add a link system, something similar to what sonarr,radarr and the other *arrs have, to link external platforms like imbd,tvdb and so.

Alright, I like this one much better. Having a ``official`` built-in feature for playing, but being able to add more options as ``add-ons``. That's a yes from me :)

gantoine:
@thorin8k Just a quick question: how is RunR different then [webrcade](https://github.com/webrcade/webrcade)? I've played around with the idea of loading games into webrcade directly from romm, with some success.

To be frank, if we're going to support an external emulation provider, I'd lean towards something that's tried and tested.

thorin8k:
@gantoine The truth is that i've discovered webRcade last week and, at that point, i had RunR working already 😅

I'm not very familiar with it but the main differences i see are:
- Selfhosting: You can selfhost RunR and it can run locally. Here i'm no sure if webRcade will work with local download paths (I'll test this).
- Complexity: For webRcade you need to create a JSON with your rom config, with RunR you just need a couple of query parameters.

Said that, i will investigate it more deeply and see what it can do, it seems like a more mature alternative with lots of configurations already included 😄

PR #525, opened 2023-12-28
Backend support for webRcade feed

gantoine:
This PR adds an API endpoint to generate and make available a [webRcade](https://github.com/webrcade/webrcade)-compatible feed file, which can be loaded in the [webRcade](https://play.webrcade.com/app/editor/). This allows users to play load their library into webrcade and play their roms directly in the browser.

The main draw of webrcade is that users don't _need_ to self-host their own instance, as this will work with the cloud instance, but they have the option to do so if they want to. And as webrcade fetches assets in the browser (not via some backend server) your local server doesn't have to be exposed to the web for this to work with the cloud instance.

In webRcade, load the feed with the following URL: `https://<host>/api/platforms/feed`. Note that only supported platforms will appear in the app. https://docs.webrcade.com/apps/

### Still left to do

* Display the URL to load the feed somewhere in the UI
* Wiki page to explain how to load your feed
* Support for loading psx (and other) bios files

### Screenshots

Screenshot 2023-12-27 at 8 08 48 PM
Screenshot 2023-12-27 at 8 09 01 PM


backend/endpoints/platform.py
```diff
@@ -25,3 +80,40 @@ class Config:
 def platforms(request: Request) -> list[PlatformSchema]:
     """Returns platforms data"""
     return dbh.get_platforms()
+
+
+@protected_route(router.get, "/platforms/feed", [])
```
zurdi15: What do you think about changing the endpoint to ``platforms/feed/webrcade``, just in case we will support any other kind of feed in the future?

docker/Dockerfile
```diff
@@ -11,6 +11,7 @@ ARG WEBSERVER_FOLDER=/var/www/html
 COPY --from=front-build-stage /front/dist ${WEBSERVER_FOLDER}
 COPY ./frontend/assets/default_avatar.png ${WEBSERVER_FOLDER}/assets/
 COPY ./frontend/assets/platforms ${WEBSERVER_FOLDER}/assets/platforms
+COPY ./frontend/assets/feed ${WEBSERVER_FOLDER}/assets/feed
```
zurdi15: same with the folder structure, changing it to ``frontend/assets/feed/webrcade``?

zurdi15 (approved):

PR #542, opened 2024-01-01
Update example docker-compose for webrcade

gantoine:


examples/docker-compose.example.yml
```diff
@@ -7,6 +7,7 @@ services:
     container_name: romm
     environment:
       - ROMM_DB_DRIVER=mariadb # mariadb | sqlite (default: sqlite)
+      - ROMM_HOST=https://localhost:3000 # your host ip or domain name (including http(s)://, subdomain and port if needed)
```
zurdi15, line 10: maybe we should set it as optional and specify that currently it's only for the webRcade support?

gantoine: Not adding it (or adding it wrong) shouldt cause any issues for users who dont use webrcade, but feel free to mark it optional

zurdi15 (approved):

PR #555, opened 2024-01-02
handle files with no extension

bfenty:
Rom scan errors with files without file extension, excluding them from the scan should fix the error.


gantoine:
@bfenty This looks great! Can you re-open this PR against the master branch?

bfenty:
My bad...re-opened.

PR #556, opened 2024-01-02
Filter out files without extension

bfenty:


gantoine (approved):

bfenty:
Found that hidden files starting with a period and without a file extension also cause issues, so added some logic to filter them out as well.

gantoine:
I don't think `max_ext_length` will work as we've come across valid extensions longer then 4 characters. If you're looking to filter hidden files check that the first char isn't a period.

bfenty:
Ok, I'll rework it. That's a better solution.

gantoine (approved):
looks great, thanks for the fix!

PR #570, opened 2024-01-04
Details view revamped and optimized

zurdi15:
Details view was a mess with too much custom css to be responsive caused by a bad design.

Now it was redesigned almost from scratch so it will be more consistent and responsive in the long term

### DIFF
--- a/frontend/src/components/Drawer/Base.vue
+++ b/frontend/src/components/Drawer/Base.vue
@@ -14,7 +14,7 @@ import RailFooter from "@/components/Drawer/Footer.vue";
 const { lgAndUp } = useDisplay();
 const platforms = storePlatforms();
 const auth = storeAuth();
-const drawer = ref(lgAndUp.value ? true : false);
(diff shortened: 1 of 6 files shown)


frontend/src/components/Drawer/Base.vue
```diff
@@ -10,9 +11,10 @@ import PlatformListItem from "@/components/Platform/PlatformListItem.vue";
 import RailFooter from "@/components/Drawer/Footer.vue";
 
 // Props
+const { lgAndUp } = useDisplay();
 const platforms = storePlatforms();
 const auth = storeAuth();
-const drawer = ref(true);
+const drawer = ref(lgAndUp.value ? true : false);
```
gantoine:
```suggestion
const drawer = ref(!!lgAndUp.value);
```

gantoine (commented):
love it!

gantoine (approved):

PR #574, opened 2024-01-05
Eliminate 400s on cover/screenshot fetch

gantoine:
Just check that there's a valid IGDB ID before trying to fetch cover and screenshots


zurdi15:
Nice! really like it! this will clean the logs a lot.

zurdi15 (approved):

PR #576, opened 2024-01-07
[ROMM-575] File exclusions accept wildcards

gantoine:
https://docs.python.org/3/library/fnmatch.html

### DIFF
--- a/backend/utils/fs.py
+++ b/backend/utils/fs.py
@@ -4,6 +4,7 @@ import shutil
 from pathlib import Path
 import datetime
 import requests
+import fnmatch
 from urllib.parse import quote
 from PIL import Image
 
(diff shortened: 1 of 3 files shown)


backend/utils/fs.py
```diff
@@ -240,18 +241,20 @@ def _exclude_files(files, filetype) -> list[str]:
     filtered_files: list = []
 
     for file in files:
-        # Exclude files starting with a period.
-        if file.startswith('.'):
+        # Split the file name to get the extension.
```
zurdi15, line 244: So, if users want to exclude files starting with ``dot`` they need to configure it manually now?

gantoine: Correct, it's a lot more flexible this way if there's a case where someone wants to scan hidden files (not sure why but 🤷🏼).

zurdi15 (approved):

PR #586, opened 2024-01-09
Checks and shows when a new version is available

zurdi15:
Now when a new version is available a message will pop in the footer of the drawer and will have a link to the release notes.

This feature can be tested locally being in a branch named ``1.0.0`` for example.

### DIFF
--- a/backend/utils/__init__.py
+++ b/backend/utils/__init__.py
@@ -148,7 +148,7 @@ def get_version() -> str | None:
     
 
 def check_new_version() -> str | None:
-    response = requests.get("https://api.github.com/repos/zurdi15/romm/releases/latest")
+    response = requests.get("https://api.github.com/repos/zurdi15/romm/releases/latest", timeout=0.5)
     try:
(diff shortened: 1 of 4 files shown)


backend/utils/__init__.py
```diff
@@ -143,3 +145,17 @@ def get_version() -> str | None:
             return None
         branch = [a for a in output.split("\n") if a.find("*") >= 0][0]
         return branch[branch.find("*") + 2 :]
+    
+
+def check_new_version() -> str | None:
+    response = requests.get("https://api.github.com/repos/zurdi15/romm/releases/latest")
```
gantoine: did you test how much this request slows down first load? like how long this query takes to resolve

zurdi15: It's imperceptible, and if you reach the api request limit it's instant. In any case I'm going to add a half second ``timeout`` just in case it get's stuck for some reason

gantoine (approved):
lgtm!

PR #590, opened 2024-01-09
Add additional comments to config.example.yml

jiannazzone:
- Added some additional comments to config.example.yml to help explain the difference between single and multi part ROMs
- Added a wildcard example under `multi_part:`


gantoine (approved):
👏🏼

PR #603, opened 2024-01-15
Refactor backend to adapt RESTful architecture

zurdi15:
This PR adds several refactors and improvements:

 - API is now following ``RESTful`` architecture. ``Endpoints``, ``forms`` and ``responses`` have been separated in different modules.

 - Frotend have been refactored to match backend refactor

 - ``Utils`` have been revamped into ``handlers``. ``Handlers`` have been modularized into ``filesystem hanlders`` and ``database hanlders`` mainly.

 - Scan is now more robust and the log is ``preserved`` between changes of sections.

 - ``Platforms binding`` now can be edited throught the UI and any change in the config.yml is being loaded at runtime without the need of restarting the container.

 - Deleting and adding/scanning games/platforms is now more responsive and will show automatically any change without need to``refresh`` the browser.

 - Added a ``/stats`` endpoint to show some general stats.

 - Calculation of the games ``size`` have been changed to clean up the database

 - ``ER diagram`` of the database have been refactorized

 - Some other little changes and fixes

Backend refactor to adapt RESTful architecture:

 - [x] Heartbeat
 - [x] Config
 - [x] Identity
 - [x] Platforms
 - [x] Roms
 - [x] Saves
 - [x] States
 - [x] Screenshots
 - [x] Users

WIP:

 - [x] Unit tests refactor

_*Managing exclusions from UI is not implemented yet because it will be refactored following #457_


backend/alembic/versions/0015_platform_id_refactor.py
```diff
@@ -0,0 +1,151 @@
...
+            None, "platforms", ["platform_slug"], ["slug"], ondelete="CASCADE"
+        )
+
+    with op.batch_alter_table("states", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("platform_slug", sa.String(length=50), nullable=False)
+        )
+        batch_op.create_foreign_key(
+            None, "platforms", ["platform_slug"], ["slug"], ondelete="CASCADE"
+        )
```
gantoine, line 150: This all looks good but will likely fail for some sqlite users (dropping/changing primary keys). I think 3.0 will likely be the version we deprecate sqlite...

backend/config/config_manager.py
```diff
@@ -57,13 +41,21 @@ def __init__(self, **entries):
...
         FileNotFoundError: Raises an error if the config.yml is not found
     """
 
+    _self = None
+
+    def __new__(cls, *args, **kwargs):
+        if cls._self is None:
+            cls._self = super().__new__(cls, *args, **kwargs)
+        
+        return cls._self
```
gantoine, line 61: performance improvement?

backend/config/config_manager.py
```diff
@@ -228,5 +222,55 @@ def _validate_config(self):
...
+            self._validate_config()
+
+    def update_config(self) -> None:
+        try:
+            with open(self.config_path, "w") as config_file:
+                yaml.dump(self._raw_config, config_file)
+        except FileNotFoundError:
+            self._raw_config = {}
+        finally:
+            self._parse_config()
```
gantoine, line 254: We might need to catch an error if the file isn't writable and return that to the UI (or at least log it)

backend/endpoints/states.py
```diff
@@ -0,0 +1,94 @@
...
+
+@protected_route(router.get, "/states/{id}", ["assets.read"])
+def get_save(request: Request, id: int) -> MessageResponse:
+    pass
+
+
+@protected_route(router.put, "/states/{id}", ["assets.write"])
+def update_save(request: Request, id: int) -> MessageResponse:
+    pass
+
```
gantoine: Update these for state and comment them out, we shouldnt have them in the openapi if they don't work

gantoine (changes requested):

zurdi15 (commented):
> This all looks good but will likely fail for some sqlite users (dropping/changing primary keys). I think 3.0 will likely be the version we deprecate sqlite... 

Yeah I agree, each time we have a migration supporting sqlite is a pain in the ass

backend/alembic/versions/0015_platform_id_refactor.py
zurdi15: Yeah I agree, each time we have a migration supporting sqlite is a pain in the ass. Added a check to log it as unsupported if someone configure it to use sqlite

backend/config/config_manager.py
zurdi15: Unless there is a extremely big amount of configuration, this is more like a design decision to be sure that everything works as expected

backend/config/config_manager.py
zurdi15: I added couple of exceptions to manage and send to the front permission issues

PR #621, opened 2024-01-19
Platforms versions support

### LINKED ISSUE #569: [Feature] Support for 'platform subversions'
Is your feature request related to a problem? Please describe:
The naomi-2 slug is a 'version' of the 'arcade' platform.  as in /arcade/version/naomi-2.  RomM is currently unable to handle this ''subversion' of arcade?

Describe the solution you'd like:
RomM should support platform 'subversions' so that users can separate rom platform subtypes

image

### LINKED ISSUE #609: [Feature] ENV to define config.yml location
I use named docker volumes and is not possible to mount a file when using NFS/SMB, is possible to create a ENV VAR to change the location of config.yml to allow put in any other random folder?

zurdi15:
This PR adds support for platform versions. For example: ``NAOMI`` as version of ``Arcade``. The association must be configured in the ``config.yml`` file in the new ``system.versions`` section.

Roms in the platform version will be scanned search using in the ``main platform id``  as IGDB doesn't provide games for api calls for platform versions.

Now ``config.yml`` file must be placed inside a ``config`` folder and bind it as follows to docker: ``/path/config:/romm/config`` and if the ``config.yml`` file doesn't exists it will be created empty automatically.

config

versions

Closes #569 
Closes #609 

WIP:

 - [x] Unit tests

### DIFF
--- a/examples/docker-compose.example.yml
+++ b/examples/docker-compose.example.yml
@@ -41,7 +41,7 @@ services:
     volumes:
       - "/path/to/library:/romm/library"
       - "/path/to/resources:/romm/resources" # [Optional] Path where roms metadata (covers) are stored
-      - "/path/to/config.yml:/romm/config.yml" # [Optional] Path where config is stored
+      - "/path/to/config:/romm/config" # [Optional] Path where config is stored
       - "/path/to/database:/romm/database" # [Optional] Only needed if ROMM_DB_DRIVER=sqlite or not set
       - "/path/to/logs:/romm/logs" # [Optional] Path where logs are stored
     ports:
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ romm_mock
 # testing
 backend/romm_test/resources
 backend/romm_test/logs
+backend/romm_test/config
 .pytest_cache
 
 # service worker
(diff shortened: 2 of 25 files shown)


examples/docker-compose.example.yml
```diff
@@ -41,7 +41,7 @@ services:
     volumes:
       - "/path/to/library:/romm/library"
       - "/path/to/resources:/romm/resources" # [Optional] Path where roms metadata (covers) are stored
-      - "/path/to/config.yml:/romm/config.yml" # [Optional] Path where config is stored
+      - "/path/to/config:/romm/config" # [Optional] Path where config is stored
```
gantoine, line 44: we'll need to mark this as a breaking change in 3.0 release

gantoine, line 44: we'll need to mark this as a breaking change in 3.0 release

gantoine (approved):
great stuff! i think doing it this way via the config, as opposed to relying on IGDB's relationships, is the smarter idea long term. good call 👍🏼

examples/docker-compose.example.yml
zurdi15: Yes absolutely!

PR #640, opened 2024-02-01
Major asset refactor + ready for 3.0

gantoine:


DEVELOPER-SETUP.md
```diff
@@ -87,6 +88,7 @@ npm install
 ```sh
 mkdir assets/romm
 ln -s ../../../romm_mock/resources assets/romm/resources
+ln -s ../../../romm_mock/assets assets/romm/assets
```
gantoine, line 91: So to keep a very distinct line between stuff uploaded by users, and stuff imported from igdb, i added a new volume `assets`. It's more likely users will want to bind this access them directly, then the igdb resources.

backend/endpoints/raw.py
```diff
@@ -1,17 +1,39 @@
...
+
+    Args:
+        request (Request): Fastapi Request object
+
+    Returns:
+        FileResponse: Returns a single rom file
+    """
+
+    rom_path = f"{LIBRARY_BASE_PATH}/{path}"
+    return FileResponse(path=rom_path, filename=path.split("/")[-1])
```
gantoine, line 27: Split into different endpoints for roms and assets, since theyre loaded from different base paths

backend/endpoints/sockets/scan.py
```diff
@@ -129,133 +116,9 @@ async def scan_platforms(
...
-
-                if rom:
-                    scanned_screenshot.rom_id = rom.id
-                    db_screenshot_handler.add_screenshot(scanned_screenshot)
-
-            db_save_handler.purge_saves(rom.id, [s for _e, s in fs_saves])
-            db_state_handler.purge_states(rom.id, [s for _e, s in fs_states])
-            db_screenshot_handler.purge_screenshots(
-                rom.id, [s for _e, s in fs_screenshots]
-            )
```
gantoine, line 224: no longer need to scan for saves/states since they're loaded from the api. i might add this back when we build device sync, since it might make more sense at that point.

zurdi15: I'm not sure if I am understanding this correctly, if we don't scan them how are we going to write them in the database?

gantoine: Right now the only way to add saves/states is by uploading them in the browser (api), or saving them when using emulatorjs (also api). Given the complexity of the save path we need to use to avoid collisions (ex `/assets/users/557365723a31/states/n64/mupen64plus/Super Mario 64.srm`), I think forcing users to upload saves via the UI makes the most sense.

When we build device sync, we can add a new scanner and watcher for the assets folder, but that all depends on how we implement sync.

zurdi15 (approved):

PR #656, opened 2024-02-10
Hotfix storing screenshots when igdb fetch fails

gantoine:
No need for an RC4, this can go out when 3.0 goes live

### DIFF
--- a/backend/handler/fs_handler/fs_resources_handler.py
+++ b/backend/handler/fs_handler/fs_resources_handler.py
@@ -3,6 +3,7 @@ import shutil
 from pathlib import Path
 from urllib.parse import quote
 from urllib3.exceptions import ProtocolError
+from logger.logger import log
 import requests
 from config import RESOURCES_BASE_PATH
 from handler.fs_handler import (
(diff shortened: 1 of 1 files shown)


backend/handler/fs_handler/fs_resources_handler.py
```diff
@@ -154,7 +156,10 @@ def _store_screenshot(self, fs_slug: str, rom_name: str, url: str, idx: int):
         if res.status_code == 200:
             Path(screenshot_path).mkdir(parents=True, exist_ok=True)
             with open(f"{screenshot_path}/{screenshot_file}", "wb") as f:
-                shutil.copyfileobj(res.raw, f)
+                try:
+                    shutil.copyfileobj(res.raw, f)
+                except ProtocolError:
+                    pass
```
zurdi15: should we log a warning message?

backend/handler/fs_handler/fs_resources_handler.py
```diff
@@ -2,7 +2,7 @@
 import shutil
 from pathlib import Path
 from urllib.parse import quote
-
+from urllib3.exceptions import ProtocolError
 import requests
 from config import RESOURCES_BASE_PATH
```
zurdi15, line 8:
```suggestion
from config import RESOURCES_BASE_PATH
from logger.logger import log
```

backend/handler/fs_handler/fs_resources_handler.py
zurdi15:
```suggestion
                    log.warning("Couldn't fetch cover from IGDB")
```

zurdi15 (approved):

PR #661, opened 2024-02-13
Mobygames API support

gantoine:


zurdi15:
It's conflicting with the scan overflow fix, but should be easy to resolve

zurdi15 (approved):

PR #662, opened 2024-02-16
Even more fixes for 3.0

gantoine:


backend/main.py
```diff
@@ -40,7 +40,7 @@
     allow_headers=["*"],
 )
 
-if "pytest" not in sys.modules:
+if "pytest" not in sys.modules and not DISABLE_CSRF_PROTECTION:
```
gantoine, line 44: Users are given the option to disable CSRF protection if it causes issues

frontend/src/views/Library/Scan/Base.vue
```diff
@@ -14,6 +14,9 @@ const rescanUnidentified = ref(false);
 const platforms = storePlatforms();
 const platformsToScan = ref<Platform[]>([]);
 
+// Connect to socket on load to catch running scans
+if (!socket.connected) socket.connect();
```
gantoine, line 18: this is important to check!

backend/main.py
zurdi15: Should we add it to the docker-compose example?

gantoine: no it'll go in https://github.com/zurdi15/romm/wiki/Environment-Variables

frontend/src/services/api/utils.ts
```diff
@@ -0,0 +1 @@
+function asyncAxiosFetch
```
zurdi15, line 1: what is this? 😕

zurdi15 (approved):

backend/main.py
zurdi15: ahh yes, forgot about it! good one wiki page

PR #672, opened 2024-02-25
An ngaging addition

Casuallynoted:
Adds Ngage icons! 👯


gantoine (approved):
as always, thank you!

PR #673, opened 2024-02-25
Mobile icons!

Casuallynoted:


gantoine (approved):
👍🏼

need4swede:
iPhone 14 Pro - can't really play games on device at all with existing layout.
Is this an emulatorJS issue? Anyone else playing on iOS without any issues?

IMG_1147
IMG_1146

gantoine:
Emulation on iOS is problematic, not so much an issue with EmulatorJS as it is with iOS as a platform.

Issue #680, opened 2024-03-09
[Bug] Underscores as word-spacing gives incorrect search scoring


sondregronas:
### **RomM version**
v2.3.1

### **Describe the bug**
Using `_` instead of ` ` as a word-spacing character results in incorrect metadata results.

I get that this might be more of an issue with IGDB (plus not following naming convention to a tee), but seems like its expected to work & should have an easy workaround. It is a bit similar to issue #454, however this is more related to the using the `_` character for word-spacing.

### **Steps to reproduce / Example**
A ps2 rom named `God of War.gz` is correctly matched as "God of War", but `God_of_War.gz` gets incorrectly matched as "God of War II". Searching for `God of War` also scores `God of War II` higher. (searching on the IGDB.com frontend however scores it as expected)

### **Screenshots**
![image](https://github.com/zurdi15/romm/assets/44143748/d6a42517-57b9-4954-acb1-a345ace67c1f)
![image](https://github.com/zurdi15/romm/assets/44143748/82087290-851a-4643-b2ec-bb6e0fbaeae7)
![image](https://github.com/zurdi15/romm/assets/44143748/02ec7f19-2dad-4912-956a-0bcae7e7c3a0)

### **Additional context**
I haven't tested the 3.0.0RC to see if it's any better with this or had any time to build it for myself, but I presume adding something to [these lines here](https://github.com/zurdi15/romm/blob/release/backend/handler/igdb_handler.py#L127) to also substitute word-spacing alternatives (primarily `_`) to regular whitespace could resolve a lot.

I've just renamed my ROMs to get around this issue myself (after all there were only a handful), feel free to close if this is fixed in `3.0.0`, or an issue with the metadata service - just couldn't find any open issues on this.

Maybe some regex in the `exact_matching` could help resolve this and #454 too:
```python
def words_only(text: str) -> str:
    return ' '.join((re.findall(r'(?:(?!_)\w)+', text)))
if words_only(rom["name"]).lower() == words_only(search_term).lower()
```

This example is of course a bit of a hack and doesn't work with titles starting with `The` or `A` should one want to name stuff like `Game Title, The` (unless IGDB accounts for this), but I think it could work.

sondregronas:
Adding to the example above one could also use `unicodedata.normalize` to remove diactritics from each word as well;

```python
import re
import unicodedata

def normalize_rom_name(text: str) -> str:
[...]

PR #681, opened 2024-03-09
Normalize rom names when parsing exact matches (remove diacritics & special characters)

### LINKED ISSUE #454: [Bug] Special characters awareness
Try to match the names by ignoring the special characters if no match is found, for example in this case the game was not found because the rom name contains a "-" instead of a "/" that of course it's an illegal character for file paths.

image

image

sondregronas:
Removes diacritics, special characters, accents and articles from rom names when comparing exact matches

Figured I'd just create a PR after all (See #680), might also resolve #454, at least when the rom's name is otherwise equivalent.

(Though this does nothing to the search term itself, so I'm not 100% sure if it solves #454)


sondregronas:
Do you want me to update the PR to go into release instead?

gantoine:
> Do you need me to update the PR to go into release instead?

No master is good, we merge master into release before each new version is released (including RCs)

sondregronas:
@gantoine Done - sorry for the delay.

zurdi15 (approved):

gantoine:
@sondregronas Sorry I think I misspoke, can you make your PR look exactly like #710 and re-open it to merge into our master branch instead? I want the work to be attributed to you as a contributor, and not have to close this PR.

PR #716, opened 2024-03-15
fix: player overflowing viewport

### LINKED ISSUE #714: [UI/UX] Player view overflowing viewport + ad hoc suggestions
On a wide viewport the player view will exceed the height of the window:
image

This is due to the forced 16:9 aspect-ratio of the EJS container [here](https://github.com/zurdi15/romm/blob/b3ad9eb22b189b4ac6fd043f0af607cfbb5aee49/frontend/src/views/Play/Base.vue#L124) along with an inherited `width: 100%;` style prop. My suggestion is to limit the wrapper size to bounds of the viewport, and the aspect-ratio property will continue to work w/ added spacing on the respective axis:
image

Additionally, I think (among some other more minor UI/UX concerns) it would be prudent to have some sort of back navigation in the provided overlay UI. e.g. a 'back to roms' button or menu entry

I took the time to clone the repo and get the app up and running, and will gladly push up my fix for one or both of these issues if this would be helpful. What is the procedure for PRs (i.e. what target branch, other expected due diligence not noted in the docs)?

JMBeresford:
Fixes #714

### DIFF
--- a/frontend/src/views/Play/Base.vue
+++ b/frontend/src/views/Play/Base.vue
@@ -37,7 +37,7 @@ function onPlay() {
 </script>
 
 <template>
-  <v-row v-if="rom" no-gutters>
+  <v-row v-if="rom" class="h-screen" no-gutters>
     <v-col v-if="!gameRunning" cols="3" class="px-3">
       <v-img
         class="mx-auto mt-6 mb-5"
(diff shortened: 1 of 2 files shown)


zurdi15:
Thank you very much for the fix! As you asked in #714 , apart from PR to ``master`` there is no other procedure to contribute, i'll just review it and merge :)

gantoine (approved):

JMBeresford:
@zurdi15 Great, thanks for the prompt update

PR #721, opened 2024-03-16
Allow redis to run in a different host

malev:
This PR will allow users to run Redis in a different host.
Two new env vars will be introduced: `REDIS_HOST` AND `REDIS_PORT`


gantoine:
A few more things need to happen in this case, namely disabling Redis from running in [init](https://github.com/zurdi15/romm/blob/release/docker/init_scripts/init). You can check if the environment variables are set, and if they are don't run the internal redis.

malev:
@gantoine now things make more sense. Thank you for pointing that out
I guess the question is: is there appetite to extract Redis out of the container? At first glance, the answers should be YES! But, it's up to the maintainers

gantoine:
Integrating Redis into RomM was a decision we made for the recent 3.0 release, based on feedback we received from users. Given how we use it, it should be completely transparent to users. Is there a specific reason you want direct access to the Redis?

malev:
It's ultimately the team's decision
Running multiple processes in a single container is not considered best practice. That's all

gantoine:
> Running multiple processes in a single container is not considered best practice

Yep I agree fully, which is why it was first setup (in 2.x) as a seperate container. However in that time we've added multiple processes in the main romm container (a filesystem watcher, a task scheduler, etc), so adding redis to the list wasn't a hude step.

As we build out further, we'll likely change what and how to runs different tasks. Any thoughts on the matter would be appreciated!

gantoine (approved):

gantoine:
Thanks for this PR!

PR #730, opened 2024-03-18
Dynamic PUID and GUID

gantoine:
* Allow setting PUID/GUID/UMASK with environment variables
* Bump alpine image version to onstall python 3.11
* Set folder user/group IDs in `init`
* Fix background scanning after socketio upgrade
* Fix download bug with file names in safari

@psych0d0g you might want to have a look at this

### DIFF
--- a/backend/handler/socket_handler.py
+++ b/backend/handler/socket_handler.py
@@ -12,4 +12,4 @@ class SocketHandler:
             client_manager=socketio.AsyncRedisManager(redis_url),
         )
 
-        self.socket_app = socketio.ASGIApp(self.socket_server)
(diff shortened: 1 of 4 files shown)


zurdi15:
Looks alright from my perspective, let's wait for what @psych0d0g needs to say!

psych0d0g:
lgfm, as far as i see the PID / GID change is optional, so should keep compatibility with readonly setups as long as one dosnt change them, which is usually not required for running on k8s, i will think about solutions to add such functionality in a dynamic way via the helm chart at one point, i would say can be merged, and thanks for consulting :)

PR #748, opened 2024-03-25
Revert PUID/GUID and use `user` property in docker-compose

gantoine:
Instead of PUID/GUID, run the image as the `romm` user and use `user: "XXXX:XXXX"` in docker-compose/`docker run` to set the owner of newly created file/folders and mounted volumes.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,8 +24,7 @@ RUN apk add --upgrade \
         mariadb-connector-c \
         netcat-openbsd \
         python3 \
-        tzdata \
-        shadow
+        tzdata
 
 # Install additional build dependencies
 RUN apk add --upgrade \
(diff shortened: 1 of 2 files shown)


gantoine:
@psych0d0g @bfenty If either of you are aware of any serious issues using `chmod 777` for our data folders please let me know!

zurdi15 (approved):

psych0d0g:
was this done because of https://github.com/rommapp/romm/issues/749 ?

if yes, it would be potentially better to just set defaults to the PUID etc vars in init and skip the -z check in the if clauses:
PUID=${PUID:1000}
# sets PUID to value of PUID and if not defined to 1000

see https://www.baeldung.com/linux/bash-variable-assign-default

Issue #761, opened 2024-03-26
[UI/UX] Add platform controller icons for PC-FX, Astrocade, Nuon, Tiger Gizmondo and Tapwave Zodiac


ameturpoet:
There are some supported platforms which still do not have specific icons - it would be nice to see them get their own platform icons instead of the default platform icon.

Casuallynoted:
Hello! This sounds like a job for me :) I'll see if I can't put some icons together this week and submit them to the project!

Casuallynoted:
NGL, Libretro has fantastic icons for all of these - I'm going to submit a PR with the icons and attribution :)

gantoine:
Thanks @Casuallynoted, these will be included in the next release

PR #764, opened 2024-03-27
Add additional platform icons (astrocade, pc-fx, etc.)

Casuallynoted:
This icon adds all of the platform icons mentioned in #761, and I threw in the hyperscan for good measure cause I'm cool like that! Also added attributions.


gantoine (approved):
Awww yeah, love me some new icons!

PR #804, opened 2024-04-08
Add platform + general upload roms dialog added

### LINKED ISSUE #472: [Feature] Upload games from the homepage & select platform during upload

zurdi15:
- Select platform to add roms added as new dialog. Platform is created if not exists.
- Fixed igdb/mobygames ID management 
- Fixed mobygames cover art pull.

### Drawer
imagen

### Upload roms
imagen

### Select source
VdeosinttuloHechoconClipchamp-ezgif com-video-to-gif-converter

Closes #472


backend/endpoints/responses/search.py
```diff
@@ -6,5 +6,6 @@ class SearchRomSchema(BaseModel):
     slug: str
     name: str
     summary: str
-    url_cover: str
+    igdb_url_cover: str = ""
+    moby_url_cover: str = ""
```
gantoine, line 10: this is a good change, would be nice to have a computed field called url_cover so we don't need to do `x or y` everywhere on the client

backend/endpoints/rom.py
```diff
@@ -331,6 +331,7 @@ async def update_rom(
         )
         cleaned_data.update({"url_cover": ""})
     else:
+        cleaned_data["url_cover"] = data.get("url_cover", db_rom.url_cover)
```
gantoine, line 334: this won't work, you should update L340 with `url_cover=data.get("url_cover", db_rom.url_cover)`

gantoine (changes requested):

backend/endpoints/rom.py
zurdi15: won't work in which terms? i tested it and it works when i want to update the cover or even when i just update some information of the rom 😕 

At that point ``cleaned_data["url_cover"]``, ``data.get("url_cover", db_rom.url_cover)`` or ``cleaned_data.get("url_cover", "")`` is literally the same, all have the same value

gantoine (approved):

backend/endpoints/responses/search.py
zurdi15: i'll add this computed variable in a next PR, but i totally agree with it

gantoine (approved):

PR #810, opened 2024-04-12
Added a button to copy download link

zurdi15:
- Added a copy download link button. Also works with the selected files in multi-part games

image


frontend/src/services/api/rom.ts
```diff
@@ -123,17 +121,22 @@ async function downloadRom({
...
-    if (!socket.connected) socket.connect();
-    storeDownload().add(rom.id);
-
-    // Clear download state after 60 seconds in case error/timeout
-    setTimeout(() => {
-      clearRomFromDownloads(rom);
-    }, 60 * 1000);
+  if (shareLink) {
+    navigator.clipboard.writeText(encodeURI(a.href));
+    return encodeURI(a.href);
```
gantoine: thb i really don't think this should live here, move the login to a util file for generating a download link, and use it here if you want

frontend/src/components/Details/ActionBar.vue
```diff
@@ -15,13 +15,37 @@ const emitter = inject<Emitter<Events>>("emitter");
...
   playInfoIcon.value = emulation.value ? "mdi-information" : "mdi-play";
   emitter?.emit("showEmulation", null);
 }
+
+async function copyDownloadLink(rom: Rom) {
+  const downloadLink = await romApi.downloadRom({
+    rom,
+    files: downloadStore.filesToDownloadMultiFileRom,
+    shareLink: true,
+  });
```
gantoine, line 42: Yeah this should be a call to a util file or something, it doesn't need to pollute `downloadRom` function

gantoine (changes requested):

gantoine:
Also you should use `mdi-content-copy` for the icon, share icon has a specific meaning as UX

frontend/src/components/Details/ActionBar.vue
zurdi15: Yeah you are right, I'll move it to a generate link function in utils

gantoine (approved):

PR #830, opened 2024-05-01
Firmware/BIOS support

gantoine:
image

image


zurdi15:
Three things here:
 - The PR needs to be changed against ``master``
 - The ``0017`` migration won't have conflicts with #828 ?
 - There are conflicts with the ``metdata handlers`` aswel

zurdi15 (approved):

PR #857, opened 2024-05-19
Platform fetch performance improvements

gantoine:
This PR should make loading the home and gallery pages a lot snappier on large libraries.

* New `GUNICORN_WORKERS` env variable lets users define how many workers to spin up
* Defer loading of app-wide resources until all high-priority requests complete
* Merge all app-wide fetches under single callback


backend/endpoints/responses/platform.py
```diff
@@ -14,16 +13,7 @@ class PlatformSchema(BaseModel):
...
     class Config:
         from_attributes = True
-
-    @classmethod
-    def from_orm_with_request(cls, db_platform: Platform, request: Request) -> "PlatformSchema":
-        platform = cls.model_validate(db_platform)
-        platform.firmware_files = [
-            FirmwareSchema.model_validate(f) for f in sorted(db_platform.firmware, key=lambda x: x.file_name)
-        ]
-        return platform
```
gantoine, line 34: This is no longer needed since as it turns out performance is unaffected by loading firmware data as well

backend/models/platform.py
```diff
@@ -24,11 +25,9 @@ class Platform(BaseModel):
...
     )
 
-    @property
-    def rom_count(self) -> int:
-        from handler.database import db_platform_handler
-
-        return db_platform_handler.get_rom_count(self.id)
+    rom_count = column_property(
+        select(func.count(Rom.id)).where(Rom.platform_id == id).scalar_subquery()
+    )
```
gantoine, line 31: Significantly faster since it runs in SQL land instead of making another DB query

frontend/src/services/api/index.ts
```diff
@@ -1,18 +1,36 @@
...
+  (response) => {
+    inflightRequests.delete(response.config.url);
+
+    // If there are no more inflight requests, fetch home data
+    if (inflightRequests.size === 0) {
+      networkQuiesced();
+    }
+
+    return response;
+  },
```
gantoine, line 37: Does what it says; fires a custom event when all network requests are complete, which triggers loading things like the platforms sidebar and heartbeat data.

zurdi15 (approved):

PR #865, opened 2024-05-22
Fix filename error

vincywindy:
If name has no ascii words，it will throw an error when starting game.
image


vincywindy:
image

pyproject.toml
```diff
@@ -50,6 +50,7 @@ Pillow = "10.3.0"
 numpy = "^1.26.4"
 fakeredis = "^2.21.3"
 yarl = "^1.9.4"
+django = "^5.0.6"
```
gantoine, line 53: This won't do, you'll need to find a way to do this without adding the entirety of django.

gantoine (changes requested):
Also please re-open this against `master`, which is the branch we develop against. Anything merged into `release` will be overridden.

PR #871, opened 2024-05-23
Fix filename error

vincywindy:
https://github.com/rommapp/romm/pull/865
Now not use django

PR #872, opened 2024-05-23
Fix filename error

vincywindy:
https://github.com/rommapp/romm/pull/871
https://github.com/rommapp/romm/pull/865


zurdi15 (approved):

zurdi15:
Seems like the import is wrong. The tests raise this error:

``ImportError: cannot import name 'quote' from 'urllib'``

In the [urllib docs](https://docs.python.org/3/library/urllib.parse.html#url-quoting) seems like it's part of the ``parse`` module

vincywindy:
Sorry，now I fix this error。

zurdi15 (approved):

gantoine:
Thanks for the PR @vincywindy!

PR #876, opened 2024-05-24
Fix rom.name char error

vincywindy:
https://github.com/rommapp/romm/pull/872
It is still has error.
image
So I fix it twice.
image


zurdi15:
Good catch!

zurdi15 (approved):

PR #912, opened 2024-05-31
GPT'd Dockerfile

gantoine:
Ran our Dockerfile though GPT 4o asking it to improve everything it can. Redis stuff in untouched since I don't understand it. Tested the output and it works as expected. 🤷🏼


zurdi15:
Does the size decrease?

gantoine:
A bit, 358MB -> 327MB

zurdi15 (approved):

PR #926, opened 2024-06-09
added linuxserver/mariadb as alternative

need4swede:
"mariadb:latest" can cause problems when mounting the database. The use of "linuxserver/mariadb:latest" may fix this issue. Added it as an alternative option.


gantoine (approved):

adamantike:
@need4swede, do you have more context on how the official `mariadb` image can cause problems? We are trying to improve stability, and recommend a single Docker image for new users to not get confused about why to pick one image or the other.

need4swede:
@adamantike, I don't recall the specifics - in hindsight I should have documented the issue in more detail. I just remember not being able to mount the database to a local directory. Using "linuxserver/mariadb:latest" fixed it for me.

When I have more time, I can try spinning up another instance and see what happens.

PR #931, opened 2024-06-14
Use OctKey directly

lepture:
joserfc 0.11 has deprecated the use of `str` and `bytes` as key to improve security. It is recommended to convert the `str` key into a `OctKey` before using it.


zurdi15 (approved):

zurdi15:
Thank you! 😄

PR #916, opened 2024-06-04
Frontend refactor

zurdi15:
General front code refactor + UI rework + improved UX:

- Redesigned drawer menu sections
- Added general action buttons for database search and games upload
- Changed gallery filters to a new secondary drawer 
- Added new drawer for firmware/bios management
- Reworked gallery table view and users managing view
- Reworked dialogs
- Reworked notes section
- Reworked saves/states section
- Reworked emulatorjs player view:
  - Added back to gallery button
  - Added back to details button
- Added 404 views when not finding/matching platforms or games

PR #933, opened 2024-06-15
Feature/improve cover handling

### LINKED ISSUE #881: [Bug] Custom hi-res cover are still low-res.
**RomM version**
3.1.0 and 3.2.0-rc.3

**Describe the bug**
When uploading a hi-res custom cover, it still default to the basic 264x352 resolution, making it even worse than the native 264x352 cover.

**To Reproduce**
1. Upload a hi-res cover.
2. Watch it become all aliased
3. You can even download the new one to see it's still low-res

**Expected behavior**
Displaying a nice and crisp cover.

 - OS: Windows 11
 - Browser: Edge + Chrome

**Additional context**
As we discussed about it on Discord, I'm creating this issue here so you can backlog it ! :)

zurdi15:
Features:
- Covers are now fetch in their 1080p version. Also fixes #881 
- Screenshots are fetch in their big version to work better with #932 

TODO:
- [x] Rename resources folders to romm id (migration)
- [x] Re-fetch covers on complete rescan

Depends on:
-  #916


backend/endpoints/rom.py
```diff
@@ -369,12 +368,10 @@ async def update_rom(
         file_location_s = f"{artwork_path}/small.{file_ext}"
         with open(file_location_s, "wb+") as artwork_s:
             artwork_s.write(artwork_file)
-        fs_resource_handler.resize_cover(file_location_s, CoverSize.SMALL)
```
gantoine, line 372: should we still not be resizing the small one for faster loading?

zurdi15: Good catch, i was focusing on sources cover but forgot about custom cover. I'll add it back for custom covers

PR #934, opened 2024-06-16
Refactor/main layout

### LINKED ISSUE #881: [Bug] Custom hi-res cover are still low-res.

zurdi15:
## Features
- Main layout refactor and redesign to improve responsiveness and navigation coherence
- Refactor and UX improvements of all views, sections and dialogs
- Resources structure migrated from `resources/{platform_slug}/{rom_name}` to ``resources/{platform_id}/{rom_id}``.
- Covers are now fetch in their 1080p version. Also fixes #881 
- Now sibling roms can have different covers.
- Screenshots are fetch in their big version to work better with #932 

## TODO
- [x] Generate small version of custom cover art 

**Breaking change❗**: Due to the resources structure migration, each rom with siblings (same game, different region, language, tag, etc) will loss the cover and a new scan/manual update will be needed. Just the main (the one that romm find first in the migration) will keep the cover.

Depends on:
- #916


frontend/src/components/common/NewVersion.vue
```diff
@@ -0,0 +1,66 @@
...
+onBeforeMount(async () => {
+  const response = await fetch(
+    "https://api.github.com/repos/rommapp/romm/releases/latest"
+  );
+  const json = await response.json();
+  GITHUB_VERSION.value = json.tag_name;
+  latestVersionDismissed.value =
+    VERSION === "development" ||
+    json.tag_name === localStorage.getItem("dismissedVersion");
+});
```
gantoine, line 28: this might perform better with `onMounted` (since it won't block rendering)

frontend/src/views/Gallery.vue
```diff
@@ -216,23 +213,14 @@ onMounted(async () => {
   } else {
     romsStore.setCurrentPlatform(routePlatform);
   }
-
-  // If platform is different, reset store and fetch roms
-  if (storedPlatformId != routePlatformId && !noPlatformError.value) {
```
gantoine, line 221: don't need this check anymore?

zurdi15: Yeah I optimized the gallery reset and new platform detection handling so this is not needed anymore. Actually it will break the gallery if we readded it as how it's the code now.

gantoine (approved):

PR #947, opened 2024-06-22
misc: Simplify backend environment configuration in Docker image

adamantike:
This change moves the virtualenv creation in the `Dockerfile` to a separate stage, to simplify isolating the process and reduce the need for uninstalling build dependencies.

The approach is similar to the one explained in [1]. It relies on building a virtualenv folder, and copying it in the final stage. Changing the `PATH` environment variable makes the virtualenv usable by default, without affecting the default Python installation.

Also, added Dockerfile arguments for Alpine, nginx, and Python versions, as some of them are reused, and also simplifies testing new versions.

An extra side effect is that the image size for the final stage is reduced from 315MB to 262MB.

[1] https://scribe.rip/@albertazzir/blazing-fast-python-docker-builds-with-poetry-a78a66f5aed0

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,11 +1,33 @@
+ARG ALPINE_VERSION=3.19
+ARG NGINX_VERSION=1.27.0
+ARG PYTHON_VERSION=3.11
+
 # Build frontend
-FROM node:lts-alpine as front-build-stage
+FROM node:lts-alpine${ALPINE_VERSION} as front-build-stage
 WORKDIR /front
 COPY ./frontend ./
 RUN npm install && npm run build
 
+# Build backend environment
+FROM python:${PYTHON_VERSION}-alpine${ALPINE_VERSION} AS backend-build
+
+RUN apk add --no-cache \
+        gcc \
+        mariadb-connector-c-dev \
+        musl-dev
+
+RUN pip install poetry
+
+ENV POETRY_NO_INTERACTION=1 \
+    POETRY_VIRTUALENVS_IN_PROJECT=1
+
+WORKDIR /src
+
+COPY ./pyproject.toml ./poetry.lock /src/
+RUN poetry install --no-ansi --no-cache --only main
+
 # Setup frontend and backend
-FROM nginx:1.27.0-alpine3.19-slim as production-stage
+FROM nginx:${NGINX_VERSION}-alpine${ALPINE_VERSION}-slim as production-stage
 ARG WEBSERVER_FOLDER=/var/www/html
 
 COPY --from=front-build-stage /front/dist ${WEBSERVER_FOLDER}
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,11 +1,33 @@
+ARG ALPINE_VERSION=3.19
+ARG NGINX_VERSION=1.27.0
+ARG PYTHON_VERSION=3.11
```
gantoine, line 3: 👍🏼

docker/Dockerfile
```diff
@@ -1,11 +1,33 @@
...
 
+# Build backend environment
+FROM python:${PYTHON_VERSION}-alpine${ALPINE_VERSION} AS backend-build
+
+RUN apk add --no-cache \
+        gcc \
+        mariadb-connector-c-dev \
+        musl-dev
+
+RUN pip install poetry
```
gantoine, line 19: We use poetry in non-package mode, can i be installed like this now that 1.8.0 is live, or do we still need to use `pip install --no-cache git+https://github.com/radoering/poetry.git@non-package-mode`?

https://python-poetry.org/blog/announcing-poetry-1.8.0/#better-support-for-non-package-projects

docker/Dockerfile
```diff
@@ -83,13 +58,18 @@ RUN addgroup -g 1000 -S romm && adduser -u 1000 -D -S -G romm romm && \
 
 # Move everything to final stage
 FROM scratch as final-stage
+
 COPY --from=production-stage / /
 
+COPY --from=backend-build /src/.venv /src/.venv
+# Fix virtualenv link to python binary
+RUN ln -sf "$(which python)" /src/.venv/bin/python
+ENV PATH="/src/.venv/bin:${PATH}"
```
gantoine, line 67: 👍🏼

docker/init_scripts/init
```diff
@@ -5,10 +5,6 @@ set -o nounset           # treat unset variables as an error
 set -o pipefail          # treat errors in pipes as fatal
 shopt -s inherit_errexit # inherit errexit
 
-# use virtualenvs
-# shellcheck disable=SC1091
-source /backend/bin/activate
```
gantoine, line 10: will be automatically sourced with this change?

gantoine (changes requested):
this is dope! will build it on my server once i'm back home to test.

docker/Dockerfile
adamantike: Exactly, `poetry` 1.8.0 now supports `package-mode = false`, so we can move to upstream and stop relying on that fork.

docker/init_scripts/init
adamantike: With the change to the `PATH` environment variable, the virtualenv basically becomes the "official" Python environment inside the Docker container, so there's no need to activate/deactivate. We get the isolation of the venv approach, without the need to remember to call the appropiate `python`/`pip` binary!

gantoine (approved):

PR #950, opened 2024-06-23
misc: Apply changes generated by pyupgrade

adamantike:
Changes automatically made by
[`pyupgrade`](https://github.com/asottile/pyupgrade), using the `--py311-plus` flag, to be compatible with Python 3.11.

### DIFF
--- a/backend/endpoints/responses/oauth.py
+++ b/backend/endpoints/responses/oauth.py
@@ -1,5 +1,6 @@
-from typing_extensions import TypedDict
 from typing import NotRequired
+
+from typing_extensions import TypedDict
 
 
 class TokenResponse(TypedDict):
--- a/backend/endpoints/responses/oauth.py
+++ b/backend/endpoints/responses/oauth.py
@@ -1,4 +1,5 @@
-from typing_extensions import NotRequired, TypedDict
+from typing_extensions import TypedDict
+from typing import NotRequired
 
 
 class TokenResponse(TypedDict):
--- a/backend/endpoints/tests/test_identity.py
+++ b/backend/endpoints/tests/test_identity.py
@@ -20,9 +20,7 @@ def test_login_logout(admin_user):
 
     assert response.status_code == 405
 
-    basic_auth = base64.b64encode(
-        b"test_admin:test_admin_password"
-    ).decode("ascii")
+    basic_auth = base64.b64encode(b"test_admin:test_admin_password").decode("ascii")
     response = client.post("/login", headers={"Authorization": f"Basic {basic_auth}"})
 
     assert response.status_code == 200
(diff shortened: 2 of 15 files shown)


backend/endpoints/tests/test_identity.py
```diff
@@ -21,7 +21,7 @@ def test_login_logout(admin_user):
     assert response.status_code == 405
 
     basic_auth = base64.b64encode(
-        "test_admin:test_admin_password".encode("ascii")
+        b"test_admin:test_admin_password"
```
gantoine: are these equivalent?

adamantike: Default argument for `.encode()` is `"utf-8"`. However, as this testing string doesn't include any non-UTF8 characters, it is indeed equivalent.

```python
>>> b64encode("test_admin:test_admin_password".encode("ascii")).decode()
'dGVzdF9hZG1pbjp0ZXN0X2FkbWluX3Bhc3N3b3Jk'
>>> b64encode("test_admin:test_admin_password".encode()).decode()
'dGVzdF9hZG1pbjp0ZXN0X2FkbWluX3Bhc3N3b3Jk'
>>> b64encode(b"test_admin:test_admin_password").decode()
'dGVzdF9hZG1pbjp0ZXN0X2FkbWluX3Bhc3N3b3Jk'
```

gantoine (approved):

gantoine:
Thanks for all these PRs! For future ones, it would be great if you could setup commit signing, so gh can validate you're authoring these commits. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

backend/endpoints/responses/oauth.py
```diff
@@ -1,4 +1,6 @@
-from typing_extensions import NotRequired, TypedDict
+from typing import NotRequired
+
+from typing_extensions import TypedDict
```
gantoine, line 3: Getting formatter errors for these changes

```
 5:1   high  Module "typing" has no attribute "NotRequired"                                        mypy/attr-defined  
 5:1   note  Use `from typing_extensions import NotRequired` instead. See                          mypy/note          
               https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-t                     
               he-typing-module
```

gantoine (changes requested):

backend/endpoints/responses/oauth.py
adamantike: The formatter should be running with an older Python version, as `NotRequired` was added to `typing` [in Python 3.11](https://docs.python.org/3/library/typing.html#typing.NotRequired).

adamantike:
> For future ones, it would be great if you could setup commit signing, so gh can validate you're authoring these commits.

All set!

backend/endpoints/responses/oauth.py
adamantike: Last commit updates Python version used by trunk, that should fix the CI run!

gantoine (approved):

PR #954, opened 2024-06-23
misc: Migrate to SQLAlchemy declarative models

adamantike:
This change applies the guided migration process recommended by SQLAlchemy [1], up to step 4, to have declarative ORM models that better support Python typing.

The change was tested by running `alembic check`, which does not find any schema changes.

Errors reported by `mypy` go down to 170, from the original 223 in the current `master` commit.

[1] https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#migrating-an-existing-mapping

### DIFF
--- a/backend/models/assets.py
+++ b/backend/models/assets.py
@@ -1,31 +1,33 @@
+from datetime import datetime
 from functools import cached_property
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from models.base import BaseModel
-from sqlalchemy import BigInteger, Column, DateTime, ForeignKey, Integer, String, func
-from sqlalchemy.orm import relationship
+from sqlalchemy import BigInteger, DateTime, ForeignKey, String, func
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+if TYPE_CHECKING:
+    from models.rom import Rom
+    from models.user import User
 
 
 class BaseAsset(BaseModel):
     __abstract__ = True
 
-    id = Column(Integer(), primary_key=True, autoincrement=True)
-    created_at = Column(
-        DateTime(timezone=True), server_default=func.now(), nullable=False
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True)
+    created_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), server_default=func.now()
     )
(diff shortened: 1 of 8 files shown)


backend/models/assets.py
```diff
@@ -1,31 +1,33 @@
...
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from models.base import BaseModel
-from sqlalchemy import BigInteger, Column, DateTime, ForeignKey, Integer, String, func
-from sqlalchemy.orm import relationship
+from sqlalchemy import BigInteger, DateTime, ForeignKey, String, func
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+if TYPE_CHECKING:
```
gantoine, line 9: AAAAHHH I forgot you can do this 😫

backend/models/assets.py
```diff
@@ -39,22 +41,18 @@ def download_path(self) -> str:
...
+    rom_id: Mapped[int] = mapped_column(ForeignKey("roms.id", ondelete="CASCADE"))
+    user_id: Mapped[int] = mapped_column(ForeignKey("users.id", ondelete="CASCADE"))
 
 
 class Save(RomAsset):
     __tablename__ = "saves"
     __table_args__ = {"extend_existing": True}
 
-    emulator = Column(String(length=50), nullable=True)
+    emulator: Mapped[str | None] = mapped_column(String(length=50))
```
gantoine, line 52: don't we have to mark this `nullable=True`?

gantoine (approved):

backend/models/assets.py
adamantike: We don't! The typehint takes care of that, as explained [in step four](https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#step-four-remove-orm-mapped-column-directives-where-no-longer-needed), so `Mapped[str]` implicitly sets `nullable=False`, while `Mapped[str | None]` sets `nullable=True`.

gantoine:
@adamantike If you happen to be in the discord give me (arcaneasada) a ping in #open-development. And if not, feel free to join. We do all of our dev planning on there, and if you have more PRs in mind I'd love to hear what you have planned. https://discord.gg/ThzzNYq9

PR #963, opened 2024-06-27
Select main sibling

### LINKED ISSUE #959: [Feature] Preferred/Default Game Language
Yes. When your library has multiple copies of the same game from multiple regions (USA, JP, GER, FRA, etc.) and you have grouped them in the UI, it is a pain to select the language on each game. It seems random which version is populated by default.

A user setting to select "preferred" or default language would be excellent. If you want your games in English, selecting English should force the game selector to an English version of the game. If there is no English version it could be random (or the setting could be a list and select by priority).

Ungrouping is the only way I think you can easily get around this now, but it's even worse with the library so bloated.

zurdi15:
Now users can choose which game will appear in the gallery when grouping sibling games.

- Added a star-like button at the right side of the version selector

Partially fixes #959


alex-taxiera:
I would say this might only partially close #959. It sounds like a "new" problem with this system is "I have to star every game in my library at least once" which is still technically better.

zurdi15:
> I would say this might only partially close #959. It sounds like a "new" problem with this system is "I have to star every game in my library at least once" which is still technically better.

Game versions can be more things than just language. It can be hacks, patches, languages, regions, custom tags, etc. Maybe at some point we can add an specific option from the settings to choose a predefined language across all RomM, but that will be an added layer to this one.

backend/alembic/versions/0021_fav_sibling.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("roms", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "fav_sibling", sa.Boolean(), nullable=True, server_default=sa.false()
+            )
+        )
```
gantoine: This needs to relate to the user somehow, we can't have it be global for every rom. You can probably rename and extend RomNote to something more generic, that includes fields like this one.

backend/endpoints/platform.py
```diff
@@ -96,7 +100,15 @@ def get_platform(request: Request, id: int) -> PlatformSchema:
...
-    return db_platform_handler.get_platforms(id)
+    platform = db_platform_handler.get_platforms(id)
+    if platform:
+        return platform
+    else:
+        msg = f"Platform with {id} not found"
+        log.critical(msg)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=msg
+        )
```
gantoine:
```suggestion
    msg = f"Platform with {id} not found"
    log.critical(msg)
    raise HTTPException(
        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=msg
    )
```

gantoine (changes requested):

backend/endpoints/responses/rom.py
```diff
@@ -26,25 +26,27 @@
 )
 
 
-class RomNoteSchema(BaseModel):
+class RomPropsSchema(BaseModel):
```
gantoine:
```suggestion
class UserRomPropsSchema(BaseModel):
```

backend/handler/auth/base_handler.py
```diff
@@ -22,8 +22,8 @@
     "assets.read": "View assets",
     "assets.write": "Modify assets",
     "firmware.read": "View firmware",
-    "notes.read": "View notes",
-    "notes.write": "Modify notes",
+    "rom_props.read": "View rom props",
+    "rom_props.write": "Modify rom props",
```
gantoine:
```suggestion
    "user_rom.read": "View user-rom props",
    "user_rom.write": "Modify user-rom props",
```

backend/handler/database/roms_handler.py
```diff
@@ -122,6 +122,17 @@ def update_rom(self, id: int, data: dict, session: Session = None) -> Rom:
...
+    @begin_session
+    def set_main_sibling(self, rom: Rom, user_id: int, session: Session = None) -> None:
+        siblings = [rom.id for rom in rom.get_sibling_roms()]
+        return session.execute(
+            update(UserRomProps)
+            .where(
+                and_(UserRomProps.rom_id.in_(siblings), UserRomProps.user_id == user_id)
+            )
+            .values(is_main_sibling=False)
+        )
```
gantoine: nothing in here sets the main sibling to true?

frontend/src/stores/roms.ts
```diff
@@ -60,10 +60,19 @@ export default defineStore("roms", {
...
-          ...(games.shift() as SimpleRom),
-          siblings: games,
-        }))
+        .map((games) => {
+          const primaryGame = games.shift();
+          // const favSiblingIndex = games.findIndex((game) => game.fav_sibling);
+          // const primaryGame =
+          //   favSiblingIndex !== -1
+          //     ? games.splice(favSiblingIndex, 1)[0]
+          //     : games.shift();
```
gantoine: can this be removed?

gantoine (changes requested):
i like `user_rom_props`. i think it makes it clear what the table does 👍🏼

backend/handler/database/roms_handler.py
zurdi15: the asignation to true is in a previous query in the endpoint function, but i need to merge both queries to make them atomic

frontend/src/stores/roms.ts
zurdi15: yeah actually i'm not filtering by the main sibling yet, still need to figure out how because here we have SimpleRoms and the user_rom_props are in the DetailedRom model. Probably need to change that property to the SimpleRom and remake how we fetch a list of roms from the backend

backend/endpoints/platform.py
```diff
@@ -100,7 +101,12 @@ def get_platform(request: Request, id: int) -> PlatformSchema:
         PlatformSchema: Platform
     """
 
-    return db_platform_handler.get_platform(id)
+    platform = db_platform_handler.get_platform(id)
+
+    if not platform:
+        raise PlatformNotFoundInDatabaseException(id)
```
gantoine, line 107: this returns a 404 status right? cause we need it to 404 for the API

backend/endpoints/sockets/scan.py
```diff
@@ -29,7 +29,7 @@
 from rq.job import Job
 from sqlalchemy.inspection import inspect
 
-EXCLUDED_FROM_DUMP: Final = {"created_at", "updated_at"}
+EXCLUDED_FROM_DUMP: Final = {"created_at", "updated_at", "rom_user"}
```
gantoine, line 32: should this be rom_user or rom_user_id? need to try running a scan to see...

backend/exceptions/endpoint_exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+from logger.logger import log
+
+
+class PlatformNotFoundInDatabaseException(Exception):
+    def __init__(self, id):
+        self.message = f"Platform with id '{id}' not found"
+        super().__init__(self.message)
+        log.critical(self.message)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=self.message
```
gantoine: these should be 404 errors

backend/handler/database/roms_handler.py
```diff
@@ -1,35 +1,55 @@
...
-from sqlalchemy import Select, and_, delete, func, or_, select, update
+from models.rom import Rom, RomUser
+from sqlalchemy import and_, delete, func, or_, select, update
 from sqlalchemy.orm import Query, Session, selectinload
 
 from .base_handler import DBBaseHandler
 
 
-def with_assets(func):
+class ImplementationError(Exception): ...
```
gantoine: unused, can be removed

frontend/src/components/Details/Info/FileInfo.vue
```diff
@@ -1,77 +1,155 @@
...
 import type { DetailedRom } from "@/stores/roms";
 import { formatBytes } from "@/utils";
+import { ref, watch } from "vue";
 
-defineProps<{ rom: DetailedRom; platform: Platform }>();
+// Props
+const props = defineProps<{ rom: DetailedRom; platform: Platform }>();
 const downloadStore = storeDownload();
+const auth = storeAuth();
+const ownProps = ref(
```
gantoine:
```suggestion
const romUser = ref(
```

frontend/src/components/Details/Saves.vue
```diff
@@ -11,14 +11,9 @@ import { useDisplay } from "vuetify";
...
 const props = defineProps<{ rom: DetailedRom }>();
-const romRef = ref<DetailedRom>(props.rom);
 const selectedSaves = ref<SaveSchema[]>([]);
 const emitter = inject<Emitter<Events>>("emitter");
-emitter?.on("romUpdated", (romUpdated) => {
-  if (romUpdated?.id === romRef.value.id) {
-    romRef.value.user_saves = romUpdated.user_saves;
-  }
-});
+// emitter?.on("romUpdated", (romUpdated) => {});
```
gantoine: remove

frontend/src/services/api/rom.ts
```diff
@@ -142,6 +142,8 @@ async function updateRom({
   const formData = new FormData();
   if (rom.igdb_id) formData.append("igdb_id", rom.igdb_id.toString());
   if (rom.moby_id) formData.append("moby_id", rom.moby_id.toString());
+  // if (rom.fav_sibling)
+  //   formData.append("fav_sibling", rom.fav_sibling.toString());
```
gantoine: remove comment

frontend/src/stores/roms.ts
```diff
@@ -60,10 +60,20 @@ export default defineStore("roms", {
             game.igdb_id || game.moby_id || nanoid(),
         ),
       )
-        .map((games) => ({
-          ...(games.shift() as SimpleRom),
-          siblings: games,
-        }))
+        .map((games) => {
+          const mainSiblingIndex = games.findIndex(
```
gantoine, line 66: comment here explaining what this does would be good for future us

gantoine (changes requested):

gantoine (approved):

PR #968, opened 2024-06-30
misc: Split database handler logic to get/list entities

adamantike:
The current implementation for some of the database handlers, where the same method is used to retrieve either a single entity (when an `id` is passed), a list of entities, or `None`, makes the typing and overall design more complex.

This change simplifies database handlers, by having two separate methods where appropiate:

* A method that receives an `id`, and returns either an entity, or `None`.
* A method that optionally receives filters, and returns (depending on the current handler implementation) a list of entities, or a `Select` object that allows chaining more SQLAlchemy operations.

### DIFF
--- a/backend/endpoints/platform.py
+++ b/backend/endpoints/platform.py
@@ -100,7 +100,7 @@ def get_platform(request: Request, id: int) -> PlatformSchema:
         PlatformSchema: Platform
     """
 
-    return db_platform_handler.get_platforms(id)
+    return db_platform_handler.get_platform(id)
 
 
 @protected_route(router.put, "/platforms/{id}", ["platforms.write"])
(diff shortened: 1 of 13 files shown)


backend/endpoints/platform.py
```diff
@@ -134,9 +134,9 @@ async def delete_platforms(request: Request, id: int) -> MessageResponse:
         MessageResponse: Standard message response
     """
 
-    platform = db_platform_handler.get_platforms(id)
+    platform = db_platform_handler.get_platform(id)
     if not platform:
-        error = f"Platform {platform.name} - [{platform.fs_slug}] not found"
+        error = f"Platform id {id} not found"
```
gantoine, line 139: 🤦🏼

gantoine (approved):
THANK YOU for finally making this change, I've been meaning to for ages but kept putting it off.

PR #969, opened 2024-06-30
fix: Add implementation for total filesize statistic

adamantike:
Add query that calculates total filesize statistic, currently hardcoded to zero.


gantoine (approved):
👏🏼

PR #970, opened 2024-06-30
misc: Simplify statistics query to retrieve platforms count

adamantike:
Instead of counting `Platform` rows, and filtering based on a subquery in `Rom`, we can simply count the different `platform_id` values in `Rom`.


gantoine (approved):
love there PRs, keep em coming!

PR #975, opened 2024-07-04
Allow users to edit own profile

zurdi15:
Now users can edit their own profile

Depends on:
- #976

### DIFF
--- a/backend/endpoints/user.py
+++ b/backend/endpoints/user.py
@@ -89,7 +89,7 @@ def get_user(request: Request, id: int) -> UserSchema:
     return user
 
 
-@protected_route(router.put, "/users/{id}", ["users.write"])
+@protected_route(router.put, "/users/{id}")
 def update_user(
     request: Request, id: int, form_data: Annotated[UserForm, Depends()]
 ) -> UserSchema:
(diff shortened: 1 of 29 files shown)


backend/endpoints/user.py
```diff
@@ -113,17 +113,23 @@ def update_user(
...
-    user = db_user_handler.get_user(id)
-    if not user:
-        raise HTTPException(status_code=404, detail="User not found")
+    db_user = db_user_handler.get_user(id)
+    if not db_user:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="User not found"
+        )
+
+    if not db_user.id == request.user.id and request.user.role != Role.ADMIN:
```
gantoine: Add short comment like "Admin users can edit any user, while other users can only edit self"

backend/endpoints/user.py
```diff
@@ -94,7 +94,7 @@ def get_user(request: Request, id: int) -> UserSchema:
     return user
 
 
-@protected_route(router.put, "/users/{id}", ["users.write"])
+@protected_route(router.put, "/users/{id}")
```
gantoine: you still want to protect this endpoint so its only available to authed users

```suggestion
@protected_route(router.put, "/users/{id}", ["me.write"])
```

gantoine (changes requested):

gantoine (approved):

PR #964, opened 2024-06-28
Collections

zurdi15:
- Collections added
- Reactivity added to all covers

Closes #219 

Depends on:
 - #960 
 - #963

PR #976, opened 2024-07-04
Setup wizard added for new installations

zurdi15:
Setup wizard added for new installations:

 - Create default admin user
 - Check metadata providers are correctly configured

Depends on:
- #964


backend/endpoints/heartbeat.py
```diff
@@ -27,6 +28,7 @@ def heartbeat() -> HeartbeatResponse:
 
     return {
         "VERSION": get_version(),
+        "SETUP_WIZARD": len(db_user_handler.get_admin_users()) == 0,
```
gantoine:
```suggestion
        "SHOW_SETUP_WIZARD": len(db_user_handler.get_admin_users()) == 0,
```

backend/endpoints/user.py
```diff
@@ -16,7 +16,10 @@
 
 
 @protected_route(
-    router.post, "/users", ["users.write"], status_code=status.HTTP_201_CREATED
+    router.post,
+    "/users",
+    [] if len(db_user_handler.get_admin_users()) == 0 else ["users.write"],
```
gantoine: 👍🏼

backend/handler/database/users_handler.py
```diff
@@ -42,4 +43,10 @@ def delete_user(self, id: int, session: Session = None):
...
     @begin_session
     def get_admin_users(self, session: Session = None) -> list[User]:
-        return session.scalars(select(User).filter_by(role=Role.ADMIN)).all()
+        try:
+            return session.scalars(select(User).filter_by(role=Role.ADMIN)).all()
+        except ProgrammingError as e:
+            if "Table" in str(e.orig) and "doesn't exist" in str(e.orig):
+                return []
+            else:
+                raise ProgrammingError from e
```
gantoine: Can you add a comment explaining this error block?

frontend/src/services/api/index.ts
```diff
@@ -37,9 +38,12 @@ api.interceptors.response.use(
   },
   (error) => {
     if (error.response?.status === 403) {
+      const allCookies = Cookies.get(); // Get all cookies
+      for (const cookie in allCookies) {
+        Cookies.remove(cookie); // Remove each cookie
+      }
       router.push({
         name: "login",
-        params: { next: router.currentRoute.value.path },
```
gantoine, line 42: why remove this? it's what brings users back to the page they were just on after login

frontend/src/views/Login.vue
```diff
@@ -1,29 +1,26 @@
...
 const username = ref("");
 const password = ref("");
 const visiblePassword = ref(false);
 const logging = ref(false);
 
-function login() {
+async function login() {
   logging.value = true;
-  identityApi
+  await identityApi
```
gantoine, line 20: this form of using await and then is odd, don't worry about fixing it in this PR, i want to refactor those anyways

frontend/src/views/Setup.vue
```diff
@@ -0,0 +1,228 @@
...
+    name: "MobyGames",
+    value: "moby",
+    logo_path: "/assets/scrappers/moby.png",
+    disabled: !heartbeat.value.METADATA_SOURCES?.MOBY_API_ENABLED,
+  },
+  {
+    name: "SteamgridDB",
+    value: "sgdb",
+    logo_path: "/assets/scrappers/sgdb.png",
+    disabled: !heartbeat.value.STEAMGRIDDB_ENABLED,
```
gantoine: should also be under `METADATA_SOURCES`

```suggestion
    disabled: !heartbeat.value.METADATA_SOURCES?.STEAMGRIDDB_ENABLED,
```

gantoine (changes requested):

backend/handler/database/users_handler.py
zurdi15: hmm i can't remember why i added that there, maybe something in the middle of the development, i'll remove it

frontend/src/views/Login.vue
zurdi15: yeah sounds good, we need to revisit the whole users management system anyways!

frontend/src/services/api/index.ts
zurdi15: it was not working for me and tried to fix it but forgot about it. Fixing it now!

gantoine (approved):

PR #978, opened 2024-07-06
Fix permission protected urls

zurdi15:
Fixed accessing urls directly by users without proper permissions

Depends on:
- #977


gantoine:
closing since most of these changes are already in master branch

PR #983, opened 2024-07-07
fix: Disable zip compression on ROM download

adamantike:
Zip compression during multi-file ROM download is not needed. Commonly, ROMs are already compressed, and trying to compress them again on the fly does not reduce the generated file's size, while spending CPU power and making the download slower because of the compression process being the bottleneck.

By setting `level=0`, compression gets disabled, increasing download speed, and reducing CPU usage. CPU still gets a usage peak, because of Zip requiring a CRC32 calculation over each file's content.

* **Before** these changes:
Screenshot_20240706_222451

* **After** these changes:
Screenshot_20240706_220942

### DIFF
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -1,4 +1,5 @@
 import os
+from collections.abc import Iterator
 from datetime import datetime
 from shutil import rmtree
 from stat import S_IFREG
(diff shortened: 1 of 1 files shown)


gantoine (approved):
Given that many users will be using this from within their home network, trading bandwidth for CPU cycles makes a lot of sense. And for those hosting on a private VPS, they can always pre-compress their games, like you mentioned. LMGT 👍🏼

PR #984, opened 2024-07-07
fix: Make some DB columns non-nullable

adamantike:
This change makes some database columns non-nullable, and includes the migration to seamlessly upgrade the schema.

It includes:
* User and Platform names
* Every boolean column

### DIFF
--- a/backend/alembic/versions/0023_make_columns_non_nullable.py
+++ b/backend/alembic/versions/0023_make_columns_non_nullable.py
@@ -1,0 +1,74 @@
+"""Make some columns non-nullable.
+
+Revision ID: 0023_make_columns_non_nullable
+Revises: 0022_collections
+Create Date: 2024-07-07 13:44:25.811184
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0023_make_columns_non_nullable"
+down_revision = "0022_collections"
+branch_labels = None
+depends_on = None
+
+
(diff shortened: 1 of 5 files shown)


backend/alembic/versions/0023_make_columns_non_nullable.py
```diff
@@ -0,0 +1,74 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("platforms", schema=None) as batch_op:
+        batch_op.execute("UPDATE platforms SET name = '' WHERE name IS NULL")
+        batch_op.alter_column(
+            "name", existing_type=sa.String(length=400), nullable=False
+        )
```
gantoine, line 24: assuming this will fail loudly if you try to create a platform without a name, should/did you check the code to make sure we don't try to do just that?

gantoine (approved):
another banger 💯

backend/alembic/versions/0023_make_columns_non_nullable.py
adamantike: I found that the only place where we create platforms already considers a string being required, as both replace and title are called on the platform slug to generate the name: https://github.com/rommapp/romm/blob/master/backend%2Fhandler%2Fscan_handler.py#L99

Even if the name is replaced in the next line with data received from one of the metadata peoviders, both IGDB and Moby are typed with Platform.name being always a non-nullable string.

PR #985, opened 2024-07-07
fix: Correctly traverse SteamGridDB API pages

adamantike:
The SteamGridDB API has a bug, where the `total` field for the paginated endpoint `/v2/grids/game/:gameId` sometimes double/triple counts results (e.g. returning `total=172` when there are actually 86 results).

To avoid this issue, this change relies on each page's number of results. If the received page has less than the requested amount of grids, we know that is the last page.


zurdi15:
Nice catch and fix! Thank you very much

zurdi15 (approved):

gantoine (approved):

PR #988, opened 2024-07-09
Fix config.example.yml

zurdi15:
Some users had issues with the current ``config.example.yml`` because it's not working when keeping untouched.
Adding some empty placeholders fix it and webUI will manage it automatically. Advanced users can manage it manually as always.

Also removed README warning about upgrading from 2.x to 3.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,6 @@
   </div>
 </div>
 
-> [!WARNING]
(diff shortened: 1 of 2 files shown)


gantoine (approved):
👏🏼

PR #992, opened 2024-07-10
Avoid orphan collections table on migration fail

zurdi15:
Doing the resources migration first avoids having an orphaned collections table on fail


backend/alembic/versions/0022_collections_.py
```diff
@@ -109,6 +78,39 @@ def upgrade() -> None:
                 "id": rom.id,
             },
         )
+
+    op.create_table(
```
gantoine: is it worth checking that the table doesn't exist before trying to create it? is that a bad idea?

zurdi15: True! forgot about it. I just added the check

gantoine (approved):
💯

PR #999, opened 2024-07-13
misc: Enable GZip compression for backend responses

adamantike:
With the introduction of non-paginated responses for ROMs, JSON responses for big collection could easily be at >1MiB.

This change adds a FastAPI-provided middleware to enable GZip compression for responses greater than 1KiB.


gantoine (approved):
omfg how did i miss this?

PR #1005, opened 2024-07-15
Calculate and store hashes for rom files

gantoine:
This PR changes the `files` field on a rom to store not just the file's name, but the calculated CRC, SHA1 and MD5 hashes for each file. This will allow us to (eventually) start matching games against no-intro/tosec/hasheous databases.


backend/handler/filesystem/roms_handler.py
```diff
@@ -84,12 +86,39 @@ def _exclude_multi_roms(self, roms) -> list[str]:
...
 
-        for path, _, files in os.walk(f"{roms_path}/{rom}"):
-            for f in self._exclude_files(files, "multi_parts"):
-                rom_files.append(f"{Path(path, f)}".replace(f"{roms_path}/{rom}/", ""))
+        # Check if rom is a multi-part rom
+        if os.path.isdir(f"{roms_path}/{rom}"):
+            multi_files = os.listdir(f"{roms_path}/{rom}")
+            for file in multi_files:
+                with open(Path(roms_path, rom, file), "rb") as f:
+                    data = f.read()
```
adamantike: A good alternative to reading the entire file in memory can be found in [this answer](https://stackoverflow.com/a/1131238).

This will be important, specially for bigger files, to avoid using a lot of system memory while hashes are being calculated. Let's note that, with the approach provided in the linked snippet, we should still open and read the file a single time, updating all three hash objects on each read chunk.

backend/handler/filesystem/roms_handler.py
```diff
@@ -123,34 +275,15 @@ def get_roms(self, platform: Platform):
         ]
 
         return [
-            dict(
-                rom,
+            FSRom(
+                multi=rom["multi"],
+                file_name=rom["file_name"],
                 files=self.get_rom_files(rom["file_name"], roms_file_path),
```
adamantike, line 292: Is this call to `get_rom_files` making us read all the rom files and calculate the hashes, every time a scan is triggered?

If that's the case, we need to find a way to only calculate hashes when it's the first time the file has been scanned, or its `mtime` has been updated.

backend/handler/filesystem/roms_handler.py
```diff
@@ -84,35 +165,106 @@ def _exclude_multi_roms(self, roms) -> list[str]:
...
+                while chunk := f.read(FILE_READ_CHUNK_SIZE):
+                    update_hashes(chunk)
+
+        return {
+            "crc_hash": (crc_c & 0xFFFFFFFF).to_bytes(4, byteorder="big").hex(),
+            "md5_hash": md5_h.hexdigest(),
+            "sha1_hash": sha1_h.hexdigest(),
+        }
+
+    def get_rom_files(self, rom: str, roms_path: str) -> list[RomFile]:
```
adamantike: Comparing this implementation with the old one, I think we need the call to `_exclude_files`, to honor any exclusion the user set.

backend/handler/filesystem/roms_handler.py
gantoine: alright i cleaned up a lot of stuff, most notably it'll only calculate hashes when adding new roms or on complete re-scans.

adamantike (approved):
Great implementation! 🚀

PR #1006, opened 2024-07-18
fix: Filter Switch titledb entries with empty id

adamantike:
The existing code was failing after changes on the upstream JSON file that included entries with `null` ID.


gantoine (approved):
awesome 💯 

btw you can change your upstream to point to this repo now that you have push access!

PR #1007, opened 2024-07-18
Fixed Docker compose to allow Romm to connect to MariaDB.

jesgp:
### Description

The current example for Docker Compose fails due to not opening the MariaDB ports.

### Related Issues

ROMM Docker container can now connect to the MariaDB container.

### Checklist
Please check all that apply.

- [-] I've updated the wiki accordingly
- [-] I've have updated relevant comments
- [-] I've assigned reviewers for this PR
- [-] I've added unit tests that cover the changes
- [-] All existing tests are passing

### Additional Notes

None.
```

### DIFF
--- a/examples/docker-compose.example.yml
+++ b/examples/docker-compose.example.yml
@@ -43,3 +43,5 @@ services:
       - MYSQL_PASSWORD=
     volumes:
       - mysql_data:/var/lib/mysql
+    ports:
+      - "3306:3306"


zodac:
I don't think this should be necessary? The MariaDB port doesn't need to be available outside the docker network, and the MariaDB image already uses `EXPOSE` for the default port:  
https://hub.docker.com/layers/library/mariadb/11.2.4/images/sha256-0a83fb6bfb25f0e585d9bb951c05c952b0acfc1193ab664182b1e1da0ce4c335?context=explore

The romm container will have access to the DB without this addition. Using `ports` will just allow for external access, which might be valuable for some, but probably isn't needed in the example.

jesgp:
I have tried the compose file as is, and apparently I'm not the only one with the problem (https://www.reddit.com/r/selfhosted/comments/1awvqmz/help_with_romm/). 

I've tried many different ways without opening the port, and this is the only solution that at least in my system worked - even if both containers are indeed in the same network, MariaDB by default did not allow connections from the same network, whereas via the default port it does work.

Maybe you're more knowledgable and can propose a more secure way, but for me this is the only way that worked - I'd be tempted to suspect this might be one of Dockers' iptables magic behind the scenes...

gantoine:
@zodac is right, you don't need to export the ports as the connection should happen in the internal docker network. what might help here is switching the mariadb image to `lscr.io/linuxserver/mariadb`, as it come pre-configured out of the box.

PR #1008, opened 2024-07-19
Catch and ignore json decode errors in 0022 migration

gantoine:


gantoine:
Not sure why trunk is failing, but it can be ignored (successfully runs locally)

zurdi15 (approved):

PR #1011, opened 2024-07-22
misc: Migrate MobyGamesHandler to async

adamantike:
Convert `MobyGamesHandler` methods to be asynchronous, and use an `httpx` async client, instead of `requests` sync client.

This change also introduces a `ContextVar` and the concept of a context shared between requests. The context is automatically instrumented for FastAPI requests, by using a middleware.

For processes that do not run from within requests, the `@initialize_context()` decorator must be used. For example, for tasks triggered from `rq` workers.

### DIFF
--- a/backend/handler/metadata/moby_handler.py
+++ b/backend/handler/metadata/moby_handler.py
@@ -1,15 +1,15 @@
+import asyncio
+import http
 import re
-import time
 from typing import Final, NotRequired
 from urllib.parse import quote
 
+import httpx
 import pydash
-import requests
 import yarl
 from config import MOBYGAMES_API_KEY
 from fastapi import HTTPException, status
 from logger.logger import log
-from requests.exceptions import HTTPError, Timeout
 from typing_extensions import TypedDict
 from unidecode import unidecode as uc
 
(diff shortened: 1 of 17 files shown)


backend/handler/metadata/moby_handler.py
```diff
@@ -81,38 +81,44 @@ def __init__(self) -> None:
         self.platform_url = "https://api.mobygames.com/v1/platforms"
         self.games_url = "https://api.mobygames.com/v1/games"
 
-    def _request(self, url: str, timeout: int = 120) -> dict:
+    async def _request(
+        self, requests_client: httpx.AsyncClient, url: str, timeout: int = 120
```
gantoine: there's no cleaner way to do this, like initializing the request client in a decorator? or is there a perf loss from initializing each time a function is called?

adamantike: Good call! I have migrated the implementation to use a `ContextVar` where the `httpx` async client is persisted to be reused between requests.

This approach will be useful to share other services in the future, without requiring passing these objects around.

gantoine (approved):
super clean, very nice 👌🏼

PR #1013, opened 2024-07-22
Create config.batocera-retrobat.yml

TyroneSlothrop:
Creates an example config file for Batocera/RetroBat-based mappings, as popular distributions so should hopefully save people a few hours.

Tested with a setup using RetroBat v6.0.3 folder structures, and only keeping matches for which I'm certain.


gantoine (approved):
looks awesome, thank you for contributing! 🙏🏼

PR #1021, opened 2024-07-24
misc: Replace pytest-vcr dependency with pytest-recording

adamantike:
`pytest-vcr` [1] is no longer maintained (last commit was in 2020), and `VCR.py` recommends [2] `pytest-recording` instead.

Also, `pytest-vcr` does not correctly support `httpx` [3].

This change also removes unused cassettes.

[1] https://github.com/ktosiek/pytest-vcr
[2] https://vcrpy.readthedocs.io/en/latest/usage.html#pytest-integration
[3] https://github.com/ktosiek/pytest-vcr/issues/46


gantoine (approved):
ah so it's a drop-in replacement, and you can still mark tests with `@pytest.mark.vcr`?

adamantike:
Exactly, decorator is the same, but this diff became huge as we had many legacy cassettes that weren't being used.
the `test_fastapi.py` file still includes tests using the same decorator, without issues.

PR #1023, opened 2024-07-25
misc: Migrate IGDBBaseHandler to async

adamantike:
Convert `IGDBBaseHandler` methods to be asynchronous, and use an `httpx` async client, instead of `requests` sync client.

This change also removes the direct dependency with `requests`, as the project no longer uses it, preferring `httpx` instead.

### DIFF
--- a/backend/handler/filesystem/resources_handler.py
+++ b/backend/handler/filesystem/resources_handler.py
@@ -73,6 +73,8 @@ class FSResourcesHandler(FSHandler):
                 status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
                 detail=f"Unable to fetch cover at {url_cover}: {str(exc)}",
             ) from exc
+        except httpx.ProtocolError:
+            log.warning(f"Failure writing cover {url_cover} to file (ProtocolError)")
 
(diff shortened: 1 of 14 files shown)


backend/handler/filesystem/resources_handler.py
```diff
@@ -176,8 +175,6 @@ async def _store_screenshot(rom: Rom, url: str, idx: int):
                 status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
                 detail=f"Unable to fetch screenshot at {url}: {str(exc)}",
             ) from exc
-        except ProtocolError:
-            log.warning(f"Failure writing screenshot {url} to file (ProtocolError)")
```
gantoine, line 180: no longer necessary, of you just want the error to bubble up to the client?

gantoine (approved):

backend/handler/filesystem/resources_handler.py
adamantike: Good call. `NetworkError` doesn't include this, so I added it back, and included the same check for covers

gantoine (approved):

PR #1024, opened 2024-07-25
misc: Refactor scan process by splitting single function

adamantike:
This change mainly refactors the `scan_platforms` function, moving part of its logic to `_identify_platform`, `_identify_firmware`, and `_identify_rom`.

The logic is simpler this way, each smaller function returns `ScanStats` that can be merged by the caller, and it simplifies future performance improvements.


backend/endpoints/sockets/scan.py
```diff
@@ -135,176 +154,221 @@ async def stop_scan():
...
-                    file_name=fs_fw,
-                    firmware=firmware,
-                )
+    await socket_manager.emit(
+        "scan:scanning_platform",
+        PlatformSchema.model_validate(platform).model_dump(
+            include={"id", "name", "slug"}
+        ),
+    )
+    await socket_manager.emit("", None)
```
gantoine, line 223: what's this empty one for?

adamantike: I'm not sure what we use these messages for, but this is ported as-is (line 170 in current `master`).

gantoine (approved):

backend/endpoints/sockets/scan.py
zurdi15: This is to flush the buffer so the messages are sent on real time to the client. If we don't do this messages the last message is never sent and messages are sent one behind the current one.

Maybe I should comment that in-code

PR #1030, opened 2024-07-27
fix: Enable flake8-async rules in ruff and fix warnings

adamantike:
`ASYNC` rules [1] try to find issues regarding asynchronous code. This change enables `ruff` to start reporting these issues, and fixes existing warnings.

[1] https://docs.astral.sh/ruff/rules/#flake8-async-async

### DIFF
--- a/backend/endpoints/collections.py
+++ b/backend/endpoints/collections.py
@@ -1,6 +1,7 @@
 import json
 from shutil import rmtree
 
+from anyio import open_file
 from config import RESOURCES_BASE_PATH
 from decorators.auth import protected_route
 from endpoints.responses import MessageResponse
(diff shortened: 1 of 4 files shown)


backend/endpoints/collections.py
```diff
@@ -183,12 +184,12 @@ async def update_collection(
...
             artwork_file = artwork.file.read()
             file_location_s = f"{artwork_path}/small.{file_ext}"
-            with open(file_location_s, "wb+") as artwork_s:
+            async with await open_file(file_location_s, "wb+") as artwork_s:
                 artwork_s.write(artwork_file)
                 fs_resource_handler.resize_cover_to_small(file_location_s)
 
             file_location_l = f"{artwork_path}/big.{file_ext}"
-            with open(file_location_l, "wb+") as artwork_l:
+            async with await open_file(file_location_l, "wb+") as artwork_l:
```
gantoine, line 192: whats the main benefit of this `async with await` pattern?

gantoine (approved):

backend/endpoints/collections.py
adamantike: That's just how the library has `open_file` implemented, so `await open_file` returns the async context manager, that can be used with `async with`!

PR #1043, opened 2024-07-29
[ROMM-1022] Add button to show duplicates

### LINKED ISSUE #1022: [Feature] Duplicate Filter
Sometimes when you dump in a lot of games you can have some that are mistagged and end up with duplicate games even tho one of them is actually a slightly different title

show duplicates the same you filter missing unmatched 

I donno

**Additional context**
Seems like a neat idea.

gantoine:
Add a button to only display duplicate games (even then grouped). This could be handy for finding games that were marked as duplicate, but are actually different games (like different IGDB/Moby IDs). Eventually this should also toss hashes into the mix, once that work is complete.

Fixes #1022 

Screenshot 2024-07-28 at 8 30 51 PM

### DIFF
--- a/frontend/src/stores/roms.ts
+++ b/frontend/src/stores/roms.ts
@@ -13,9 +13,7 @@ type GalleryFilterStore = ExtractPiniaStoreType<typeof storeGalleryFilter>;
 
 const collectionStore = storeCollection();
 
-export type SimpleRom = RomSchema & {
-  siblings?: RomSchema[]; // Added by the frontend
-};
+export type SimpleRom = RomSchema;
 
 export type DetailedRom = DetailedRomSchema;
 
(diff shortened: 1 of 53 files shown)


frontend/src/stores/roms.ts
```diff
@@ -45,10 +45,23 @@ export default defineStore("roms", {
   actions: {
     _reorder() {
       // Sort roms by comparator string
-      this.allRoms = this.allRoms.sort((a, b) => {
-        return a.sort_comparator.localeCompare(b.sort_comparator);
-      });
       this.allRoms = uniqBy(this.allRoms, "id");
+      this.allRoms = this.allRoms
+        .map((rom) => {
+          const siblings = this.allRoms
```
adamantike: This mapping of `this.allRoms` to itself will generate a O(n^2) operation, which could become expensive for large collections.

I think a simple way to avoid that, is to first create two maps:
* `roms_by_igdb_id`: map[str, list[Rom]]
* `roms_by_moby_id`: map[str, list[Rom]]

That way, we iterate through `allRoms` once to build those maps, and building `siblings` becomes running one lookup to each of those maps, and excluding the rom with the same `id` it's being evaluated.

frontend/src/components/Gallery/AppBar/common/FilterDrawer/FilterFavouritesBtn.vue
```diff
@@ -31,7 +31,7 @@ function setFavourites() {
         'text-romm-white': !filterFavourites,
         'text-romm-accent-1': filterFavourites,
       }"
-      >Filter favourites</span
+      >Show favourites</span
```
zurdi15, line 34: 👍

zurdi15 (approved):

PR #1057, opened 2024-08-05
Fix CSRF token issues

zurdi15:
RomM now detects if csrf token verification failed at login, regenerating the csrf token and attempting to login again, fixing once and for all the csrf token issues (hopefully 🥲)

Also added autocompletion option to login form


frontend/src/components/common/Navigation/SettingsDrawer.vue
```diff
@@ -22,13 +23,17 @@ const { smAndDown } = useDisplay();
 
 // Functions
 async function logout() {
-  identityApi.logout().then(({ data }) => {
+  identityApi.logout().then(async ({ data }) => {
+    // Refetch CSRF token
+    await refetchCSRFToken();
```
gantoine, line 28: Reset the CSRF token and logout and fetch a new one

frontend/src/views/Login.vue
```diff
@@ -17,11 +18,15 @@ const logging = ref(false);
...
   logging.value = true;
+
   await identityApi
     .login(username.value, password.value)
-    .then(() => {
-      const next = (router.currentRoute.value.query?.next || "/").toString();
-      router.push(next);
+    .then(async () => {
+      // Refetch CSRF token
+      await refetchCSRFToken();
```
gantoine, line 26: Same here but after login (new one will be fetched as well)

frontend/src/views/Login.vue
```diff
@@ -59,15 +65,16 @@ async function login() {
               <v-form @submit.prevent>
                 <v-text-field
                   v-model="username"
+                  autocomplete="on"
                   required
                   prepend-inner-icon="mdi-account"
                   type="text"
                   label="Username"
                   variant="underlined"
-                  @keyup.enter="login()"
```
gantoine, line 67: don't need these since you're using a password field

gantoine (approved):

PR #1059, opened 2024-08-07
misc: Make backend handle URLs with trailing slash

adamantike:
According to multiple FastAPI discussions [1], FastAPI only includes a built-in mechanism to redirect requests including a trailing slash, to its variation without slash, using a `307` status code.

This can be an issue when certain clients do not send the same headers on the redirected request.

This change adds a custom FastAPI `APIRouter`, that registers both route path variations (with and without trailing slash), while only marking the path without slash for being included in the OpenAPI schema.

[1] https://github.com/fastapi/fastapi/discussions/7298


gantoine (approved):
awesome!!

PR #1060, opened 2024-08-07
Drop ROMM_AUTH_USERNAME and ROMM_AUTH_PASSWORD env variables

gantoine:
No longer used since we switched to the setup wizard.


zurdi15 (approved):

StudioEtrange:
so we cannot autoamte admin creation anymore ?
without using a human action with the wizard ?

PR #1064, opened 2024-08-08
misc: Improve typing for feed schemas

adamantike:
* Add typing for nested objects in Tinfoil and Webrcade feed schemas.
* Do not send Rom background and thumbnail, if not available.
* Correctly build URLs using `starlette` utils.
* Deprecate the `ROMM_HOST` setting, no longer needed.
* Fix FastAPI custom router to prefer routes without trailing slash.
* Fix Webrcade background URL pointing to gallery screenshot.


gantoine (approved):
oh yeah good stuff 👌🏼

Issue #1051, opened 2024-08-04
[Bug] Tinfoil integration not connecting


Casuallynoted:
**RomM version**
3.3.0

**Describe the bug**
At the moment, the Tinfoil integration fails to list available switch games for installation.

**Additional context**
Already did some debugging with Asada, it's currently unclear how Tinfoil is actually supposed to be reaching out to Romm as web access logs aren't really showing any reach out attempts from Tinfoil on my end.

Thought I'd create an issue here to keep it documented- Asada is already aware and since none of the devs currently have access to a Switch for testing, it's likely to be here until we're better able to assess the situation. But figured it's good to have the issue on Github.

Pfuenzle:
From what I saw in my nging logs, this is because tinfoil is accessing `host.com/api/tinfoil/feed/`, but romm is expecting the path without a `/` at the end.

When accessing the path with a / at the end, it just redirects to the web-ui, which is obviously not working in tinfoil

Pfuenzle:
I think with the commit above the redirection works, but the authentication doesnt anymore

`<ip> - <user> [05/Aug/2024:20:14:26 +0100] "GET /api/tinfoil/feed/ HTTP/1.1" 307 0 "-" "-"
<ip>  - - [05/Aug/2024:20:14:26 +0100] "GET /api/tinfoil/feed HTTP/1.1" 403 22 "-" "-"`

adamantike:
Feel free to test the changes in https://github.com/rommapp/romm/pull/1059, and let us know if that makes any difference.

Pfuenzle:
The redirect works as expected now :)
When opening the "Repo" in Tinfoil directly, all games get listed.

But it seems that downloading does not work. 

When trying to access a file, the correct link gets accessed `(GET /api/roms/2/content/romname.NSP)`, but a error 403 gets returned. 
When accessing the link in this format using Insomnia (and curl etc...):
`https://user:password@domain.com/api/roms/2/content/game.NSP`
The file download is successful, so maybe Tinfoil does not provide the authentication header for downloads?

Pfuenzle:
With the fix above everything is working correctly on Tinfoil for me.
One weird thing is that it doesnt work with nginx as a reverse proxy, only when accessing it directly. With a reverse proxy, tinfoil doesnt even try to connect on startup. But on romm's side everything works now

adamantike:
Can you try setting the following environment variable for the `romm` container?

```shell
[...]

PR #1071, opened 2024-08-08
Fix tinfoil authentication & improvements

Pfuenzle:
The pull requests fixes the authentication error when using tinfoil to access roms.

Tinfoil uses the format `https://<user>:<pass>@<domain.com>/<file>` to access protected files using Basic Auth.

Previously, the path to the rom file given by Romm was hardcoded, and Tinfoil could not inject the Basic Auth parameters.
With this pull request, the path is now set relative to the Tinfoil endpoint. With this, Tinfoil can inject the login data and can access the file.

Also I added a new environment variable `TINFOIL_WELCOME_MESSAGE`, with which a custom message can be set, which will be displayed to the user when opening Tinfoil.

Related to https://github.com/rommapp/romm/issues/1051

### DIFF
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -88,3 +88,6 @@ SCHEDULED_UPDATE_SWITCH_TITLEDB_CRON: Final = os.environ.get(
 
 # TESTING
 IS_PYTEST_RUN: Final = bool(os.environ.get("PYTEST_VERSION", False))
+
+#TINFOIL
+TINFOIL_WELCOME_MESSAGE: Final = os.environ.get("TINFOIL_WELCOME_MESSAGE", "RomM Switch Library")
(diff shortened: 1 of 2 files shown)


adamantike:
The approach we have in mind to avoid using relative URLs and support other download flows in the future, is for the Tinfoil feed endpoint to pre-sign each download URL with a token, included as a URL query param and using a short expiration date.

The download URL route will include another possible auth strategy to read the token and validate the request that way. That will allow for the authentication not to be disabled globally in the RomM instance.

Pfuenzle:
That sounds good, its probably cleaner that using relative paths.

How long of an expiration duration are you thinking of?
Since Tinfoil retrieves the Rom list once at startup, it should probably be a few hours to a day?
Otherwise the download queue might have invalid links after downloading for some time.

Should I close this pull requests? Or keep it open for the custom welcome message?

gantoine:
I'm going to close this as we're taking a different approach for downloads (which will include a fix for tinfoil). As for the welcome message, it's a nice idea but not super relevant at the moment.

PR #1073, opened 2024-08-08
fix: fab overlay fixed

zurdi15:
Fab is broken in the last version in the gallery non-list views

Also removed overlays to fix browser back issues


gantoine (approved):
yeeesss awesome

PR #1076, opened 2024-08-09
Improve speed of fetching siblings for roms

### LINKED ISSUE #1075: [Bug] 3.4.0 is Significantly slower at loading large libraries than 3.3.0
**RomM version**
v3.4.0

**Describe the bug**
Loading platforms with large amounts of games takes significantly longer than it did on 3.3.0 (PC game collection is 1700+)

**To Reproduce**
Click on a platform with a large amount of games

**Expected behavior**
Platform should load just as fast as before (on 3.3.0)

- OS: Windows 11 (Server running Unraid, 7950x, m.2)
- Browser Opera
- Version of what? The OS or the browser?

**Additional context**
Nothing else has changed except the update of 3.4.0. It was a night and day difference as I was  browsing my pc library and noticed the notification on the bottom left. Updated, and continued on. That's when I noticed it was much, much slower. I restarted MariaDB and also optimized the database.

I have since rolled back RomM to 3.3.0 and the issue no longer remains.

gantoine:
This PR changes the way we generate the sibling roms list by using a virtual table (SQL VIEW), auto-populated with `rom_id` <-> `rom_id` entries based on matching `igdb_id/moby_id`s. We then create a relationship called `sibling_roms` using primary and secondary joins `rom`<->`sibling_roms`<->`rom`. Now all siblings will be fetched through the lazy join when resolving models.

Essentially this takes sibling generation from exponential to linear time!

Fixes #1075


adamantike (approved):

backend/alembic/versions/0024_sibling_roms_db_view.py
```diff
@@ -0,0 +1,57 @@
...
+            """
+            CREATE VIEW sibling_roms AS
+            SELECT
+                r1.id AS rom_id,
+                r2.id AS sibling_rom_id,
+                NOW() AS created_at,
+                NOW() AS updated_at
+            FROM
+                roms r1
+            LEFT JOIN
```
adamantike: Should this be a regular JOIN instead, to avoid rows where sibling_rom_id is null?

backend/alembic/versions/0024_sibling_roms_db_view.py
```diff
@@ -0,0 +1,68 @@
...
+    connection.execute(
+        sa.text(
+            """
+            CREATE VIEW sibling_roms AS
+            SELECT
+                r1.id AS rom_id,
+                r2.id AS sibling_rom_id,
+                r1.platform_id AS platform_id,
+                COALESCE(r1.igdb_id, r2.igdb_id) AS igdb_id,
+                COALESCE(r1.moby_id, r2.moby_id) AS moby_id,
```
adamantike: A few considerations we can discuss here:

* If we add more metadata providers, we would need to add its sibling logic here. Adding it to the `WHERE` is easy, but adding the new ID field here could require a DROP + CREATE if the field is not appended at the end of the current SELECT. Maybe we can move `created_at` and `updated_at` either at the beginning of the list of fields, or right after `sibling_rom_id`, so future migrations maintain similar fields close by.

```
rom_id
sibling_rom_id
created_at
updated_at
platform_id
igdb_id
moby_id
# New metadata ID fields...
```

* The current implementation using `COALESCE` could be confusing, if sibling roms match in one of the IDs but not the other. For example, if the `igdb_id` matches but `moby_id` does not, then the field value is not clear (it could be `r1.moby_id` if it's populated, even if `r2.moby_id` is different, or `r2.moby_id` if r1 has the field empty). I think the solution for this is to either avoid these fields, or only populate them when they match:

```sql
CASE WHEN r1.igdb_id <=> r2.igdb_id THEN r1.igdb_id END AS igdb_id
```

gantoine: Aight solid! If you see any other ways it can be improved push them directly to the branch.

adamantike (approved):

Issue #1079, opened 2024-08-10
[Bug] slow download due to max CPU usage?


gen-angry:
**RomM version**
3.4.0

This issue still happened with 3.3.0.

**Describe the bug**
The download speed caps out at 15-25MB/s regardless of whether I download a large single file ISO or multi-parts that are zipped on the fly.

**To Reproduce**

1. Go to a large ROM, either an ISO or a multi part folder
2. Click on download.
3. Look at speed. 

**Expected behavior**
ROMs download at 80-90MB/sec capped to the speed of my network card on my desktop machine.

**Desktop (please complete the following information):**

- OS: win11
- Browser firefox
- Version 129

**Additional context**

I have this installed with latest docker on ubuntu 24.04 LXC container on a Proxmox host. I use this LXC to run several containers and be able to mount to a folder to my hard drive storage array.

The LXC has been assigned 8 cores (of a E5-2680v4) and 48GB RAM. Other containers in the LXC run idle 99% of the time but have no problem requesting all of the resources when they need it for temporary short spikes. However, the problem persists even if I shut down all other containers on this LXC leaving RomM as the sole running application.

I noticed that while Im downloading, one CPU core will peg to max leaving the others relatively idle. I wonder if RomM is set to only use one core/thread? Is there a way to see/set how many cores this application can use? I went through all of the wiki documentation and couldn't find anything about it.

Thank you :)

adamantike:
Download speed and CPU usage will be fixed for single-file downloads in the next release.
For multi-file downloads, that are zipped on the fly, we are still exploring ways to reduce CPU usage.

PR #1084, opened 2024-08-12
feat: Use X-Accel-Redirect to improve file download speed

adamantike:
Instead of making FastAPI handle file download, which has serious performance issues on big files [1], this change uses nginx's `X-Accel` feature to delegate single-file downloads to nginx.

Partial fix for #1079, as it solves the CPU usage issue for single-file downloads.

[1] https://github.com/fastapi/fastapi/discussions/6050


gantoine (approved):

backend/endpoints/rom.py
```diff
@@ -214,11 +225,27 @@ async def get_rom_content(
...
 
     if len(files_to_download) == 1:
         return FileResponse(
-            path=f"{rom_path}/{files_to_download[0]}", filename=files_to_download[0]
+            path=f"{rom_path}/{files_to_download[0]}",
+            filename=files_to_download[0],
+            headers={
+                "Content-Disposition": f'attachment; filename="{quote(files_to_download[0])}"',
+                "Content-Type": "application/octet-stream",
+                "Content-Length": str(rom.file_size_bytes),
```
adamantike: This `Content-Length` value is incorrect for a multi-file rom where only a single file was selected. Ideally, we should use the `RomFile.size` value in that case.

This is making FastAPI trigger an exception, as the client keeps fetching data even after the file was completely downloaded.

backend/endpoints/rom.py
```diff
@@ -214,11 +225,27 @@ async def get_rom_content(
     files_to_download = files or [r["filename"] for r in rom.files]
 
     if not rom.multi:
-        return FileResponse(path=rom_path, filename=rom.file_name)
+        return FileResponse(
+            path=rom_path,
+            filename=rom.file_name,
```
adamantike: I found that using `FileResponse` when trying to use X-Accel, makes Gunicorn still use a lot of CPU time, and doesn't completely delegate the download to nginx. Download speeds I was able to measure were also slower when using `FileResponse`.

backend/endpoints/rom.py
gantoine: Good catch! I had the same issue trying to add the header to the zipped rom (too big or too small since we couldn't predict the output) and didn't include it. I'll fix this one as well.

gantoine (approved):

PR #1100, opened 2024-08-20
Bump emulatorjs to 4.1.1

gantoine:
Emulation is currently broken since they purged 4.0.12 from the releases (and cores in the CDN), so we need to bump this up ASAP.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15,7 +15,7 @@
         "axios": "^1.6.8",
         "core-js": "^3.37.1",
         "cronstrue": "^2.50.0",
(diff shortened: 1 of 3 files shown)


adamantike (approved):

gantoine:
On second though I just tested this on my server (after clearing IndexedDB) and it's still not working. Will keep digging into why tomorrow...

adamantike:
Have you updated the lockfile to get the new version here? https://github.com/rommapp/romm/blob/794a45498c0c55dcc2780e6fd1eeee64b88a0257/frontend/package-lock.json#L3843

gantoine:
uuggghhh i though i did but i didn't commit it 🤦🏼 it's fixed now lol

PR #1102, opened 2024-08-21
feat: Use nginx mod_zip to generate multi-file zip downloads

adamantike:
This change installs and configures the `mod_zip` nginx module [1], which allows nginx to stream ZIP files directly.

It includes a workaround needed to correctly calculate CRC-32 values for included files, by including a new `server` section listening at port 8081, only used for the file requests to be upstream subrequests that correctly trigger the CRC-32 calculation logic.

Also, to be able to provide a `m3u` file generated on the fly, we add a `/decode` endpoint fully implemented in nginx using NJS, which receives a `value` URL param, and decodes it using base64. The decoded value is returned as the response.

That way, the contents of the `m3u` file is base64-encoded, and set as part of the response, for `mod_zip` to include it in the ZIP file.

[1] https://github.com/evanmiller/mod_zip

### DIFF
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -1,7 +1,6 @@
-from collections.abc import AsyncIterator
-from datetime import datetime
+import binascii
+from base64 import b64encode
 from shutil import rmtree
-from stat import S_IFREG
 from typing import Annotated
 from urllib.parse import quote
 
@@ -13,12 +12,7 @@ from config import (
 )
 from decorators.auth import protected_route
 from endpoints.responses import MessageResponse
-from endpoints.responses.rom import (
-    CustomStreamingResponse,
-    DetailedRomSchema,
-    RomUserSchema,
-    SimpleRomSchema,
-)
+from endpoints.responses.rom import DetailedRomSchema, RomUserSchema, SimpleRomSchema
 from exceptions.endpoint_exceptions import RomNotFoundInDatabaseException
 from exceptions.fs_exceptions import RomAlreadyExistsException
 from fastapi import HTTPException, Query, Request, UploadFile, status
@@ -29,9 +23,10 @@ from handler.filesystem.base_handler import CoverSize
 from handler.metadata import meta_igdb_handler, meta_moby_handler
 from logger.logger import log
 from starlette.requests import ClientDisconnect
-from stream_zip import NO_COMPRESSION_64, ZIP_AUTO, AsyncMemberFile, async_stream_zip
 from streaming_form_data import StreamingFormDataParser
 from streaming_form_data.targets import FileTarget, NullTarget
+from utils.hashing import crc32_to_hex
+from utils.nginx import ZipContentLine, ZipResponse
 from utils.router import APIRouter
 
 router = APIRouter()
@@ -221,7 +216,7 @@ async def get_rom_content(
         FileResponse: Returns one file for single file roms
 
     Yields:
-        CustomStreamingResponse: Streams a file for multi-part roms
+        ZipResponse: Returns a response for nginx to serve a Zip file for multi-part roms
     """
 
     rom = db_rom_handler.get_rom(id)
@@ -230,7 +225,7 @@ async def get_rom_content(
         raise RomNotFoundInDatabaseException(id)
 
     rom_path = f"{LIBRARY_BASE_PATH}/{rom.full_path}"
-    files_to_download = files or [r["filename"] for r in rom.files]
+    files_to_download = sorted(files or [r["filename"] for r in rom.files])
 
     if not rom.multi:
         return Response(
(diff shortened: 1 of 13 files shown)


backend/endpoints/rom.py
```diff
@@ -230,7 +225,7 @@ async def get_rom_content(
         raise RomNotFoundInDatabaseException(id)
 
     rom_path = f"{LIBRARY_BASE_PATH}/{rom.full_path}"
-    files_to_download = files or [r["filename"] for r in rom.files]
+    files_to_download = sorted(files or [r["filename"] for r in rom.files])
```
gantoine, line 228: will something break if its not sorted? i'm just thinking of future me coming in an "cleaning up" this file in 6 months by removing the sort lol

backend/endpoints/rom.py
```diff
@@ -250,51 +245,29 @@ async def get_rom_content(
...
-        yield (
-            f"{file_name}.m3u",
-            now,
-            S_IFREG | 0o600,
-            NO_COMPRESSION_64,
-            m3u_file(),
+    content_lines = [
+        ZipContentLine(
+            # TODO: Use calculated CRC-32 if available.
+            crc32=None,
```
gantoine, line 251: if youre thinking of using the crc-32 file saved on the rom, i'm _assuming_ that won't work since that single value is the sum of all files in the multi-folder.

backend/endpoints/rom.py
```diff
@@ -250,51 +245,29 @@ async def get_rom_content(
...
+    ]
+
+    m3u_encoded_content = "\n".join([f for f in files_to_download]).encode()
+    m3u_base64_content = b64encode(m3u_encoded_content).decode()
+    m3u_line = ZipContentLine(
+        crc32=crc32_to_hex(binascii.crc32(m3u_encoded_content)),
+        size_bytes=len(m3u_encoded_content),
+        encoded_location=f"/decode?value={m3u_base64_content}",
+        filename=f"{file_name}.m3u",
+    )
```
gantoine, line 266: dope!

backend/utils/hashing.py
```diff
@@ -0,0 +1,2 @@
+def crc32_to_hex(value: int) -> str:
+    return (value & 0xFFFFFFFF).to_bytes(4, byteorder="big").hex()
```
gantoine, line 2: good cleanup 👍🏼

frontend/src/services/api/rom.ts
```diff
@@ -1,7 +1,6 @@
 import type { MessageResponse, SearchRomSchema } from "@/__generated__";
 import api from "@/services/api/index";
 import socket from "@/services/socket";
-import storeDownload from "@/stores/download";
```
gantoine, line 4: we'll clean up uses of the download store in a future PR

backend/utils/nginx.py
```diff
@@ -0,0 +1,49 @@
...
+                "Argument 'content' must not be provided, as it is generated from 'content_lines'"
+            )
+
+        kwargs["content"] = "\n".join(str(line) for line in content_lines)
+        kwargs.setdefault("headers", {}).update(
+            {
+                "Content-Disposition": f'attachment; filename="{filename}"',
+                "X-Archive-Files": "zip",
+            }
+        )
```
gantoine, line 47: any chance content-length header gets set here by `Response`?

docker/Dockerfile
```diff
@@ -33,10 +33,36 @@ WORKDIR /src
...
+
+RUN git clone https://github.com/evanmiller/mod_zip.git && \
+    cd ./mod_zip && \
+    git checkout "${NGINX_MOD_ZIP_SHA}" && \
+    cd ../ && \
+    git clone --branch "release-${NGINX_VERSION}" --depth 1 https://github.com/nginx/nginx.git && \
+    cd ./nginx && \
+    ./auto/configure --with-compat --add-dynamic-module=../mod_zip/ && \
+    make -f ./objs/Makefile modules && \
+    chmod 644 ./objs/ngx_http_zip_module.so
```
gantoine, line 64: at first i was wondering if forking mod_zip would be useful here, but we'd still need to compile nginx with `--add-dynamic-module` anyways i dont see a benefit.

however a comment explaining what this block does would be good (from my reading it compiles nginx with a specific SHA/commit of mod_zip?)

docker/nginx/default.conf
```diff
@@ -105,5 +110,35 @@ http {
...
                 alias /romm/library;
             }
+
+            # This location, and the related server at port 8081, are used to serve files when
+            # using the `mod_zip` module. This is because the `mod_zip` module does not support
+            # calculating CRC-32 values when using subrequests pointing directly to internal
+            # locations that access the filesystem.
+            # TODO: If that gets fixed, this workaround can be removed, and the `/library` location
+            # can be used directly (also removing the server at port 8081).
+            # Related issue: https://github.com/evanmiller/mod_zip/issues/90
```
gantoine, line 120: nice to see this issue link 👌🏼

backend/endpoints/rom.py
adamantike: Nothing will break, but this gives us better reproducibility for zipped files, specially for the `m3u` content to be always the same!

backend/endpoints/rom.py
adamantike: Yes, I considered exactly that, and found that blocker. I think we can maintain this `TODO`, as I assume in the future we could have hashes per file to better match external databases. If we get those hashes and explicitly send them here, `mod_zip` wouldn't need to generate them in real time, and download speed would be even faster.

backend/utils/nginx.py
adamantike: Good call! I didn't mention it, but the module automatically adds `Content-Length`, so that's another improvement compared to the `stream-zip` library.

gantoine (approved):

PR #1109, opened 2024-08-22
Replace illegal fs chars in filenames

### LINKED ISSUE #392: [Bug] Rename function struggles with special characters
I've noticed that sometimes the rename file function leaves the roms with really weird names in the file explorer. In this example, the PC game Babylon 5: Into the Fire has been renamed to BDY6C8~Q.ZIP

But what's weird is that the game still downloads in Romm and shows in the Romm editor as having the proper name.

I think this happens when I go to have it rename a game that has a character like : in the title, and I believe the reason is because my filesystem may not like : in file names. I've tried renaming these games manually via my server's SMB connection and it always errors out when a : is included in the name.

gantoine:
Some filesystems and OSs can't handle certain characters, so we remove-or-replace them when a) manually editing the filename or b) setting the filename to IGDB/Moby field value.

Fixes #392

### DIFF
--- a/backend/utils/filesystem.py
+++ b/backend/utils/filesystem.py
@@ -30,7 +30,8 @@ def iter_directories(path: str, recursive: bool = False) -> Iterator[tuple[Path,
             break
 
 
-INVALID_CHARS = r'[\\/:*?"<>|]'
+INVALID_CHARS_HYPHENS = re.compile(r"[\\/:|]")
+INVALUD_CHARS_EMPTY = re.compile(r'[*?"<>]')
 
 
 def sanitize_filename(filename):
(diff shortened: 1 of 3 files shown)


backend/utils/filesystem.py
zurdi15, line 1: 👍

zurdi15 (approved):

backend/utils/filesystem.py
```diff
@@ -27,3 +28,35 @@ def iter_directories(path: str, recursive: bool = False) -> Iterator[tuple[Path,
             yield Path(root), directory
         if not recursive:
             break
+
+
+INVALID_CHARS = r'[\\/:*?"<>|]'
```
adamantike: We aren't using this constant. Should we replace this by two constants, one for characters that get replaced by hyphen, and another one for characters that get removed?

Also, we can use `re.compile`, same as other regex constants.

PR #1116, opened 2024-08-25
feat(UX) - Allow links to open in a new tab

SaraVieira:
Adds support to open links to Games and Platforms in a new tab if you are holding the cmd/ctrl key

I may be missing some; please let me know if so


gantoine:
I think the right fix here is to use actual navigation tags (like `a`) on elements that should navigate. That should hopefully also address things like keyboard navigation and assistive tech.

gantoine (approved):

SaraVieira:
Nice!! Thank you!

PR #1119, opened 2024-08-26
[ROMM-1114] Fetch and use age ratings in UI and filters

### LINKED ISSUE #1114: [Feature] Add PEGI ratings to database from igdb calls
It would be really hepful to be able to sort and sift games by rating for users who are keeping track of games for younger players. From what I can tell, IGDB already has all this data there for the taking both as European PEGI ratings and USA style ESRB ratings. I haven't looked under the hood to see what the API calls look like, but am happy to do so if this hasn't already been undertaken by others with work underway.

gantoine:
This PR fetches age ratings from IGDB, sets them in the metadata, and uses them to populate a list of age rating images in the game details. This list can also be used to filter a list of games in a platform gallery or collection.

Closes #1114 

Screenshot 2024-08-26 at 3 47 48 PM


backend/handler/metadata/igdb_handler.py
```diff
@@ -101,6 +108,9 @@ def extract_metadata_from_igdb_rom(rom: dict) -> IGDBMetadata:
                 IGDBMetadataPlatform(igdb_id=p.get("id", ""), name=p.get("name", ""))
                 for p in rom.get("platforms", [])
             ],
+            "age_ratings": [
```
adamantike, line 115: I think we should consider the scenario where IGDB adds new age ratings, to not fail until `IGDB_AGE_RATINGS` is updated.

```python
"age_ratings": [
    IGDB_AGE_RATINGS[r["rating"]]
    for r in rom.get("age_ratings", [])
    if r["rating"] in IGDB_AGE_RATINGS
],
```

PR #1122, opened 2024-08-27
fix(icons) - match icon names to platform names

SaraVieira:
Changes some of the icons path to match the path expected by romm as specified in here:

https://github.com/rommapp/romm/wiki/Supported-Platforms

I was having some issues getting the master system and PICO logos to show up and this fixes that


gantoine:
The `pico-8` is not the sega pico, but the [modern fantasy console](https://www.lexaloffle.com/pico-8.php). We don't yet have an icon for the sega pico.

And the master system should be all three, `ms`, `sms` and `sega-master-system`.

gantoine (approved):

SaraVieira:
Sorry when I said pico I meant Pico 8 not the Sega one

Because of the revert this will still not find the pico 8 icon

PR #1125, opened 2024-08-27
Disable auth on rom content get endpoint when env variable is present

gantoine:
Also duplicate `pico-8` icon to `pico` for Mobygames slug match.

cc @SaraVieira


SaraVieira:
Niiice!! Thank you!

PR #1131, opened 2024-08-29
misc: Tag Docker image with major version only

### LINKED ISSUE #1130: [Feature] Add docker tags for major/minor/patch versions
### Description

As of the point of writing this, `3.4.0` is the latest Romm version.
As I'm manually updating my images and not using something like "latest", I have to change the image tag every new release. It would be nice if I wouldn't have to do that for patch releases (minor releases too).

Currently these tags are provided:
- `3.4.0` - Version 3.4.0
- `3.3.0` - Version 3.3.0
- `3.4`    - Version 3.4.0
- `latest` - Version 3.4.0

### Desired Behavior

I would propose a tag system like this:
- `3.4.0` - Version 3.4.0
- `3.3.0` - Version 3.3.0
- `3.4`    - Version 3.4.0
- `3`       - Version 3.4.0
- `latest` - Version 3.4.0

This would allow pinning to specific version and allows the user to choose which semver types they are okay with auto updating.
I for example would have pinned to `3` and would have automatically gotten the update to `3.4.0`.
However I will not get the update to `4.x` which would be desired for me.

adamantike:
Add Docker image tag that only includes its major version, for users to be able to decide whether to pin their stacks to a patch, minor, or major version.

Closes #1130.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,6 +47,7 @@ jobs:
           tags: |
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
           labels: |
             org.opencontainers.image.version={{version}}
             org.opencontainers.image.title="rommapp/romm"


gantoine (approved):

adamantike:
@gantoine, feel free to manually pull `:3.4.0` and push it as `:3`, so there's no need to wait for the next release.

gantoine:
We're going to deploy 3.5.0 in a day anyways, so IMO there's no need to push `:3`. Also I'm worried I'll break something manually pushing an tag lol. 😅

PR #1133, opened 2024-08-29
[ROMM-1107] Add env variable to disable in-browser emulation

gantoine:
This PR adds new env variables `DISABLE_EMULATOR_JS` and `DISABLE_RUFFLE_RS`, which when set to `true` will disable emulation in the UI (hide the play buttons). The value is send to the browser via the heartbeat response, and passed to `utils`.

* Boolean env variables now consider `"true"`, `True`, and `1` as truthy
* Increased target area of card actions in the UI

Screenshot 2024-08-29 at 10 10 21 AM


backend/config/__init__.py
```diff
@@ -53,35 +53,49 @@
...
-    os.environ.get("DISABLE_DOWNLOAD_ENDPOINT_AUTH", "false") == "true"
-)
+DISABLE_CSRF_PROTECTION = os.environ.get("DISABLE_CSRF_PROTECTION", "false") in [
+    "true",
+    True,
+    1,
+]
+DISABLE_DOWNLOAD_ENDPOINT_AUTH = os.environ.get(
+    "DISABLE_DOWNLOAD_ENDPOINT_AUTH", "false"
+) in ["true", True, 1]
```
adamantike: nit: Probably a good time to add a `str_to_bool` helper:

```python
def str_to_bool(value: str) -> bool:
    return value.lower() in ("true", "1")
```

I think `load_dotenv()` populates environment variables as strings, so maybe the existing `True` and `1` comparisons aren't working?

adamantike (approved):

PR #1136, opened 2024-08-29
Default both fast forward and rewind to enabled

SaraVieira:
These are two settings that in general people (including me) want on by default, you don't need to use them but they should be default enabled in my opinion.

According to my testing this seems to work, the issue is my local version of romm doesn't run ejs for some reason so i could only test by opening the menu and checking both of these were now enabled.


frontend/src/views/EmulatorJS/Player.vue
```diff
@@ -88,6 +89,7 @@ declare global {
 }
 
 const supportedCores = getSupportedEJSCores(romRef.value.platform_slug);
+window.EJS_DEBUG_XX = true;
```
gantoine: should this always be on, even for regular use?

truevox: I'm not OP, but reading the [documentation](https://emulatorjs.org/docs/options#ejs_debug_xx:~:text=Available%20options%20and,will%20be%20supported.), it doesn't seem it to me. It may be that it was left in for further testing?

SaraVieira: Definitely not, removed

frontend/src/views/EmulatorJS/Player.vue
```diff
@@ -101,7 +101,11 @@ window.EJS_color = "#A453FF";
 window.EJS_alignStartButton = "center";
 window.EJS_startOnLoaded = true;
 window.EJS_backgroundImage = "/assets/emulatorjs/loading_black.png";
-window.EJS_defaultOptions = { "save-state-location": "browser" };
+window.EJS_defaultOptions = {
+  "save-state-location": "browser",
+  rewindEnabled: "enabled",
+  fastForward: "enabled",
```
gantoine: i was able to pull this pr and test it on my server. `fastForward` doesn't just enable the option to fast forward, it runs the game fast forward. needs to be reverted. rewind enabled can stay though, that one works.

gantoine (changes requested):

frontend/src/views/EmulatorJS/Player.vue
SaraVieira: Oh damn, good catch and removed

gantoine:
Looks good, merged and will be in the next deploy! When you get a minute add your GPG key to github and setup git to auto-sign commits, commit signing is technically required before PRs can be approved/merged.

https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key

SaraVieira:
Cool! It should be done! The GPG stuff

PR #1137, opened 2024-08-29
[ROMM-829] Status fields on rom user

### LINKED ISSUE #829: [Feature] Completion checkbox
It would really like to put a the checkbox on what has been played. And after that sorted what has not been played and if needed, sorted by that tags. Best template how can manage by tags, by my opinion, realised in GOG Galaxy. Something like that would be very useful here.
This will greatly help you navigate the list of failed and completed games.

I think the checkbox could be looks like that:

table

gantoine:
This PR adds more fields onto the `RomUser` for setting game status (in progress, completed, 100%, etc) under the new `Personal` tab. Private and shared notes have been moved to the `Personal` tab as well.

Platforms and collections can be filtered by status (including backlogged and now playing). `Hidden` games will still be returned from the server but filtered out of the list.

### TODO

- [x] Actually hide games from the UI that are marked as "hidden"

Closes #829 

### Screenshots

<table>
  <tr>
    <td>Screenshot 2024-08-29 at 7 10 13 PM</td>
    <td rowspan="2">Screenshot 2024-09-08 at 3 55 51 PM</td>
  </tr>
  <tr>
    <td>Screenshot 2024-09-08 at 1 58 00 PM</td>
  </tr>
</table>


backend/models/rom.py
```diff
@@ -170,6 +174,14 @@ def __repr__(self) -> str:
         return self.file_name
 
 
+class RomUserStatus(enum.Enum):
```
adamantike: nit: We can use `enum.StrEnum` here.

backend/endpoints/states.py
```diff
@@ -94,6 +107,14 @@ async def update_state(request: Request, id: int) -> StateSchema:
         fs_asset_handler.write_file(file=file, path=db_state.file_path)
         db_state_handler.update_state(db_state.id, {"file_size_bytes": file.size})
 
+    # Set the last played time for the current user
+    current_user = request.user
+    rom_user = db_rom_handler.get_rom_user(db_state.rom_id, current_user.id)
+    if not rom_user:
+        rom_user = db_rom_handler.add_rom_user(db_state.rom_id, current_user.id)
+    rom_user.last_played = datetime.now()
```
adamantike: I think we should save timezone-aware timezones every time we update `last_played`. Also, just to avoid some datetime discrepancies, we can create a `now` variable, and use it to both populate `last_played` and call `update_rom_user`.

adamantike (approved):
🚀

Issue #1103, opened 2024-08-21
[Feature] Ability to unmatch an item that has been matched to the wrong game


Casuallynoted:
At the moment, if a rom is matched to the wrong game but the proper game is not listed on mobygames or igdb, there is no way to specify that it should not carry any metadata from the incorrect match. Having an option to manually unmatch a game would resolve this (and hopefully doing this would prevent it from being rematched during a search).

gantoine:
This has been addressed

PR #1138, opened 2024-08-30
Allow user to unmatch rom

SaraVieira:
Adds a new button on the edit modal that allows the user to unmatch a rom if that rom has a `rom.igdb_id || rom.moby_id || rom.sgdb_id`. When that button is clicked the rom basically reverts to it's default state

Not sure if this is the best way to go about it, would love some opinions

Modal

CleanShot 2024-08-30 at 19 15 46@2x

closes https://github.com/rommapp/romm/issues/1103


backend/endpoints/rom.py
```diff
@@ -290,6 +290,7 @@ async def update_rom(
     rename_as_source: bool = False,
     remove_cover: bool = False,
     artwork: UploadFile | None = None,
+    unmatch: bool = False
```
gantoine:
```suggestion
    unmatch_metadata: bool = False
```

backend/endpoints/rom.py
```diff
@@ -298,6 +299,7 @@ async def update_rom(
         id (Rom): Rom internal id
         rename_as_source (bool, optional): Flag to rename rom file as matched IGDB game. Defaults to False.
         artwork (UploadFile, optional): Custom artork to set as cover. Defaults to File(None).
+        unmatch: Weather to remove the match for this game. Defaults to false
```
gantoine:
```suggestion
        unmatch_metadata: Remove the metadata matches for this game. Defaults to False.
```

frontend/src/components/common/Game/Dialog/EditRom.vue
```diff
@@ -225,6 +225,14 @@ function closeDialog() {
     <template #append>
       <v-row class="justify-center mt-4 mb-2" no-gutters>
         <v-btn-group divided density="compact">
+          <v-btn
+            v-if="rom.igdb_id || rom.moby_id || rom.sgdb_id"
```
gantoine: in most cases we want buttons visible but disabled

```suggestion
           :disabled="rom.igdb_id || rom.moby_id || rom.sgdb_id"
```

frontend/src/components/common/Game/Dialog/EditRom.vue
```diff
@@ -225,6 +225,14 @@ function closeDialog() {
...
+            class="text-romm-red bg-terciary"
+            variant="flat"
+            @click="() => updateRom({ unmatch: true })"
```
gantoine: i think it would be cleaner to create a new `unmatchRom` function which just does:

```
await romApi
    .updateRom({ rom: rom.value, unmatch: true })
    .then...
```

gantoine (changes requested):
Approach looks good 👍🏼 Can you add a screenshot of what it looks like in the modal?

Make sure you have trunk installed: `npm install -D @trunkio/launcher`, it'll automatically run as post-commit and pre-push hook

frontend/src/components/common/Game/Dialog/EditRom.vue
SaraVieira: Done and added a screenshot

gantoine:
@SaraVieira I pushed a couple small changes, can you take a look at the typescript failures. You can run ts in the frontend folder with `npm run typecheck`.

SaraVieira:
It has been done and issues fixed :D

gantoine (approved):
awesome, gonna squeeze this in just before the release!

PR #1195, opened 2024-09-18
chore(README): Update dev instructions for Arch Linux users

rathmerdominik:
I wanted to start work on implementing OpenID Connect and got nasty errors when trying to install the dev setup on Arch Linux.

Here is what helped :smile:

### DIFF
--- a/DEVELOPER_SETUP.md
+++ b/DEVELOPER_SETUP.md
@@ -48,6 +48,13 @@ Then create the virtual environment
 poetry install --sync
 ```
 
+If you are on Archlinux you need to run the command as follows:
+
+```sh
+# https://bbs.archlinux.org/viewtopic.php?id=296542
+CFLAGS="-Wno-error=incompatible-pointer-types" poetry install --sync # https://bbs.archlinux.org/viewtopic.php?id=296542
+```
+
 #### - Spin up mariadb in docker
 
 ```sh
--- a/DEVELOPER_SETUP.md
+++ b/DEVELOPER_SETUP.md
@@ -48,11 +48,11 @@ Then create the virtual environment
 poetry install --sync
 ```
 
-If you are on Archlinux you need to run the command as follows:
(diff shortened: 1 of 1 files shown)


DEVELOPER_SETUP.md
```diff
@@ -48,6 +48,13 @@ Then create the virtual environment
 poetry install --sync
 ```
 
+If you are on Archlinux you need to run the command as follows:
+
+```sh
+# https://bbs.archlinux.org/viewtopic.php?id=296542
```
zurdi15, line 54: The info link is duplicated

adamantike: We could also mention the open PR that, when merged, will allow us to remove this specific instruction:
https://github.com/mariadb-corporation/mariadb-connector-python/pull/29

DEVELOPER_SETUP.md
```diff
@@ -48,6 +48,13 @@ Then create the virtual environment
 poetry install --sync
 ```
 
+If you are on Archlinux you need to run the command as follows:
```
adamantike:
```suggestion
If you are on Arch Linux or another Arch-based distro, you need to run the command as follows:
```

gantoine:
@rathmerdominik That's awesome! There's a work-in-progress branch you can use as a base, or get inspired from: https://github.com/rommapp/romm/tree/openid-connect
There's no active work on that branch right now so the feature's all yours.

DEVELOPER_SETUP.md
zurdi15: Totally agree!

rathmerdominik: oh. My bad
I forgot to remove it from behind the instruction :laughing: 
Will remove real quick

gantoine (approved):

gantoine:
Thanks for the PR! Join the discord if you haven't done so, we can chat in #open-development about this and other features.

PR #1205, opened 2024-09-22
misc: Use bundled Valkey instead of Redis server

### LINKED ISSUE #925: [Feature] Move from Redis to Valkey
Since Redis changed license it would be awesome to use [Valkey](https://github.com/valkey-io/valkey) instead. It is fully compatible because it is a fork from the last really open source version of Redis 7.2.4

adamantike:
This change replaces the bundled Redis server with Valkey. No breaking changes are introduced, as considered environment variables still maintain the `REDIS_` prefix.

Fixes #925.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,9 +11,9 @@ services:
     ports:
       - $DB_PORT:3306
 
-  redis:
-    image: redis:alpine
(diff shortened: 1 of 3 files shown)


gantoine (approved):
very nice!

PR #1222, opened 2024-10-04
Changed the WebUI port in the docker-compose.yml

Saetron:
Changed from 80 to 8080, this way the docker-compose.yml works


gantoine (approved):

adamantike:
I wonder why this change is needed. Can you tell us why the existing docker-compose configuration wasn't working for you?

With the existing `80:8080` port configuration, you would be able to access RomM without specifying the `:8080` port in the URL, when accessing from your browser.

Saetron:
> I wonder why this change is needed. Can you tell us why the existing docker-compose configuration wasn't working for you?
> 
> With the existing `80:8080` port configuration, you would be able to access RomM without specifying the `:8080` port in the URL, when accessing from your browser.

Wups ... shouldn't create pulls when half asleep. Brain read it 8080:80 ... Ignore me here ...

gantoine:
That makes 2 of us lol

PR #1235, opened 2024-10-12
fix: Improve handling of Redis URL

adamantike:
This change replaces the creation of Redis URL, from a simple string interpolation, to using `yarl.URL`. The main benefit, besides not forgetting to set all five different variables on every Redis client initialization, is that user credentials are correctly URL-encoded, if present.

Up until now, if a password had special characters, it could break the generated URL.

This change also introduces support for a `REDIS_SSL` setting, which allows the user to specify if the Redis connection should use SSL or not.


gantoine (approved):
👏🏼

Issue #1228, opened 2024-10-07
[Feature] Make upload timeout configurable


yuri-becker:
**Is your feature request related to a problem? Please describe.**
When uploading larger files to RomM, the upload is cancelled after exactly two minutes. When the instance is on a remote server, uploads of even DVD games can take far longer.

**Describe the solution you'd like**
Ideally, admins would be able to configure the timeout, or completely disabling it. Possibly via Environment Variables.

**Describe alternatives you've considered**
An alternative would be to just statically disable the timeout or set it to a high enough time. 

**Additional context**
I originally commented this request on this commit: https://github.com/rommapp/romm/commit/e20465f268ee4618577fe048fbc84182181ceb8c?diff=unified#diff-90fd36aad18db541451cbf233c961cb25c21a3f107ff47e0f8b89057c590797c
This also shows the relevant piece of code.

yuri-becker:
I would also be willing to implement this myself if you're open to it.

gantoine:
Yeah go for it! You'll want to add an env variable for it and import it in `config/__init__.py` (like we do for `SCAN_TIMEOUT`)

gantoine:
Closing this as its merged, will be in the next release.

PR #1236, opened 2024-10-12
#1228 Configurable upload timeout

yuri-becker:
Implementation https://github.com/rommapp/romm/issues/1228.

Tested by building the image and including it in docker-compose file close to my production environment. 
Uploads get correctly cancelled when the timeout passes (15000 in this case).

image

Uploads beyond 10 minutes don't fail with a timeout high enough or 0.
image

For the [wiki](https://github.com/rommapp/romm/wiki/Environment-Variables), the row would be something like this:

| Variable | Description | Required | Default
| -- | -- | -- | -- | 
| UPLOAD_TIMEOUT | Limit for how long a ROM upload is allowed to take, in Seconds. Set to 0 for no timeout. You probably need to this when your server is remote, and you regularly upload large ROMs. | | 600 |


backend/config/__init__.py
```diff
@@ -95,5 +95,8 @@ def str_to_bool(value: str) -> bool:
 DISABLE_EMULATOR_JS = str_to_bool(os.environ.get("DISABLE_EMULATOR_JS", "false"))
 DISABLE_RUFFLE_RS = str_to_bool(os.environ.get("DISABLE_RUFFLE_RS", "false"))
 
+# FRONTEND
+UPLOAD_TIMEOUT = int(os.environ.get("UPLOAD_TIMEOUT", 600000))
```
adamantike: As discussed on Discord, this variable should be in seconds, to maintain consistency with `SCAN_TIMEOUT`. I think that's good enough for most users. If at some point we need millisecond precision, we can add `UPLOAD_TIMEOUT_MS`, which is evaluated only if `UPLOAD_TIMEOUT` is unset.

yuri-becker: Done :)

gantoine (approved):

zurdi15 (approved):

Issue #1123, opened 2024-08-27
[Bug] Database timeout on large platform library


Perrylicious:
**RomM version**
RomM 3.4.0

**Describe the bug**
On platforms with large numbers of entries, the database response seemingly times out

**To Reproduce**

1. Have the DOS platform
2. Acquire 8284+ titles for the DOS platform
3. Try to open the DOS platform
4. Frontend shows Error "AxiosError: Request failed with status code 504" and a 404 page "platform not found" is displayed

**Expected behavior**
The platform should open. This is not platform specific, the Nint*ndo DS platform with 7782 titles also times out. 

**Logs**
From the log files of the container (sanitized for local hosts and IPs):

2024/08/27 11:48:05 [error] 27#27: *399 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 192.168.x.x, server: localhost, request: "GET /api/roms?platform_id=20&search_term=&order_by=name&order_dir=asc HTTP/1.1", upstream: "http://unix:/tmp/gunicorn.sock/api/roms?platform_id=20&search_term=&order_by=name&order_dir=asc", host: "xxxxxx", referrer: "https://xxxxxxxx/platform/20"

INFO:	  [nginx][2024-08-27 11:48:05]	192.168.1.88 - - "GET /api/roms?platform_id=20&search_term=&order_by=name&order_dir=asc HTTP/1.1" 504 569 "https://xxxxxxxx/platform/20" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36" rt=60.002 uct="0.000" uht="-" urt="60.003"

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser: Firefox and Chrome were tested
- Version: latest on Browsers and current on Windows updates

**Additional context**
There are requests that append a &limit=15 to their calls but this request seems to pull the entire library at once. The CPU of the virtual machine - 4 cores available and the only thing on the host pulling any significant CPU - goes up to 25% during the abovementioned DB query, then drops back to zero after the timeout.

Maybe options would be to either not pull the entirety of the platform at once OR introduce an env variable to increase timeout

gantoine:
> Acquire 8284+ titles for the DOS platform

![tenor](https://github.com/user-attachments/assets/fbbd0cf3-8b5d-46e0-9ca8-f56955bcd512)

Perrylicious:
[...]

PR #1240, opened 2024-10-12
Add offset parameter to get_roms endpoint

alisci01:
Related to #1123

I'm modifying local ROM management software (such as ES-DE) to query a RomM server for ROMs, as well as to download them for use on the local machine's emulators. Huge database sizes came to be an issue as there would be a timeout when attempting to make a query to the database, but would also create a better user experience of having some list of ROMs available immediately rather than waiting for the query for the entire set of requested ROMs to be completed.

The frontend making use of this is not a part of this pull request, but I would assume this can be used for pagination as mentioned in the referenced issue.

### DIFF
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -104,6 +104,7 @@ def get_roms(
     collection_id: int | None = None,
     search_term: str = "",
     limit: int | None = None,
+    offset: int | None = None,
     order_by: str = "name",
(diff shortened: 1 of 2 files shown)


adamantike (approved):

adamantike:
Thanks for your contribution! I tested it locally, and it works as intended.

gantoine:
sweet thanks for testing @adamantike, approved!

PR #1241, opened 2024-10-14
fix: Avoid query during import time to set add_user endpoint scopes

adamantike:
The `add_user` endpoint was querying the database at import time, to decide whether to enforce the `users.write` scope or not. This is problematic because the database might not be ready at import time.

Also, the decided `scopes` was being maintained for the entire application lifetime, which is not ideal, as users can be created without having the `users.write` scope, until the application is restarted.


gantoine (approved):
wow yeah great catch!

PR #1243, opened 2024-10-14
Update variables to MariaDB

drcursor:
- Added comment directing people. to the example docker compose file. 
- Changed variables to MARIADB as we are using the MARIADB container.
- Fixed small typo (missing "=")


gantoine (approved):

adamantike:
We will need to revert this change. As can be seen [here](https://github.com/linuxserver/docker-mariadb/blob/master/root/etc/s6-overlay/s6-rc.d/init-mariadb-initdb/run), the LinuxServer's MariaDB image only supports `MYSQL_` prefixed environment variables.

Considering that the official MariaDB image supports both `MARIADB_` and `MYSQL_` prefixes, we can go back to `MYSQL_` temporarily, until LinuxServer image gets support for this new naming convention.

PR #1254, opened 2024-10-24
feat: Enable editable exclusions from webUI

zurdi15:
Added config exclusion endpoints:

 - add exclusion
 - delete exclusion
 
 Enabled ``add``/``remove`` buttons for exclusions in the webUI

### DIFF
--- a/backend/endpoints/config.py
+++ b/backend/endpoints/config.py
@@ -115,25 +115,41 @@ async def delete_platform_version(request: Request, fs_slug: str) -> MessageResp
     return {"msg": f"{fs_slug} version removed successfully!"}
 
 
-# @protected_route(router.post, "/config/exclude", ["platforms.write"])
-# async def add_exclusion(request: Request) -> MessageResponse:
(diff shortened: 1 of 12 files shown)


backend/endpoints/config.py
```diff
@@ -115,25 +115,44 @@ async def delete_platform_version(request: Request, fs_slug: str) -> MessageResp
     return {"msg": f"{fs_slug} version removed successfully!"}
 
 
-# @protected_route(router.post, "/config/exclude", ["platforms.write"])
-# async def add_exclusion(request: Request) -> MessageResponse:
-#     """Add platform binding to the configuration"""
+@protected_route(router.post, "/config/exclude", ["platforms.write"])
```
adamantike: With #1247 now merged, we can use Scope objects for these two routes!

adamantike (approved):

backend/endpoints/config.py
zurdi15: True! let me adapt it

PR #1265, opened 2024-10-27
misc: Initial addition of IGDB typehints

adamantike:
The short-term goal is to completely typehint the IGDB API responses. This first change adds the base structures and enums RomM currently uses.

The `ExpandableField` type will allow us to model the expansion mechanism the IGDB API provides, where a field can include either an ID, or the full nested structure.


zurdi15:
Great idea! This will help to make the API interaction much robust

zurdi15 (approved):

PR #1271, opened 2024-10-29
fix: Personal tab mobile responsive

zurdi15:
Fixed personal tab in the game details view on mobile

### DIFF
--- a/frontend/src/components/Details/Personal.vue
+++ b/frontend/src/components/Details/Personal.vue
@@ -3,40 +3,27 @@ import romApi from "@/services/api/rom";
 import storeAuth from "@/stores/auth";
 import type { DetailedRom } from "@/stores/roms";
 import type { RomUserStatus } from "@/__generated__";
-import { getTextForStatus } from "@/utils";
+import { difficultyEmojis, getTextForStatus, getEmojiForStatus } from "@/utils";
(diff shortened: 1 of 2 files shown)


frontend/src/components/Details/Personal.vue
```diff
@@ -77,96 +64,148 @@ watch(
...
+            </v-col>
+          </v-row>
+          <v-row class="d-flex align-center mt-4" no-gutters>
+            <v-col cols="12" md="2">
+              <v-label>Difficulty</v-label>
+            </v-col>
+            <v-col cols="11" md="9">
+              <v-slider
+                :class="{ 'ml-4': mdAndUp }"
+                v-model="romUser.difficulty"
```
adamantike, line 143: Both Difficulty and Completion sliders are triggering API request with each value change, so when the user clicks and drags in the Completion slider, from 0 to 100, we are sending 100 API requests to the backend.

This could be fixed in a separate PR, but we could find a way to trigger a single update.

adamantike (approved):

frontend/src/components/Details/Personal.vue
zurdi15: Hmm yeah you are right, we should be able to just detect the click release or the touch release. I'll invesyit for the next PR

PR #1274, opened 2024-11-01
Update rom.py to Fix Filename error

vincywindy:
This filename error same like
https://github.com/rommapp/romm/pull/872


zurdi15:
Oh wow how we missed that! Thank you again for your contribution :)

zurdi15 (approved):

adamantike:
The latest version of the code is in the master branch, while the release branch includes the code for the current tagged version (v3.5.1 at the moment).

You will notice we already patched this issue in master. Thanks for contributing and working on RomM!

PR #1277, opened 2024-11-06
fix: Revert "Update variables to MariaDB"

adamantike:
This reverts changes introduced in #1243.

The LinuxServer's MariaDB image only supports `MYSQL_` prefixed environment variables.
Considering that the official MariaDB image supports both `MARIADB_` and `MYSQL_` prefixes, we can go back to `MYSQL_` temporarily, until LinuxServer image gets support for this new naming convention.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,10 +6,10 @@ services:
     container_name: mariadb
     restart: unless-stopped
     environment:
-      - MARIADB_ROOT_PASSWORD=$DB_ROOT_PASSWD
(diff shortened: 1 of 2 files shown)


zurdi15 (approved):

adamantike:
With #1278 removing the recommendation for `linuxserver:mariadb` as a drop-in replacement, I think we can put this PR on hold. The environment variables we currently use are the recommended ones by the official MariaDB image.

PR #1278, opened 2024-11-06
misc: Add MariaDB healthcheck recommendation to Docker Compose config

adamantike:
This commit adds a healthcheck configuration to the MariaDB service in the Docker Compose example configuration. The healthcheck script is a simple shell script that checks if the MariaDB server is ready to accept connections.

The application will wait for the MariaDB service to be healthy before starting the application service. This should solve issues where the database takes longer to start than the application, and logs some `Something went horribly wrong with our database` errors.

This change also stops recommending the `linuxserver/mariadb` image as an alternative. We have had users that change the image because of the first time run triggering those errors, but the `linuxserver/mariadb` image requires a different configuration that could be confusing for new users (e.g. volume mountpoint needs to be `/config` instead of `/var/lib/mysql`).

### DIFF
--- a/examples/docker-compose.example.yml
+++ b/examples/docker-compose.example.yml
@@ -29,10 +29,12 @@ services:
     ports:
       - 80:8080
     depends_on:
-      - romm-db
+      romm-db:
+        condition: service_healthy
+        restart: true
 
   romm-db:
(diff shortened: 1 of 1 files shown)


Moonraka:
This worked when I tested it. I was having issues when using the linuxserver and now everything works as it should using the main mariadb server.

zurdi15 (approved):

PR #1284, opened 2024-11-10
Fix #1283 - properly default groupRoms if it doesn't exist in localStorage

### LINKED ISSUE #1283: [Bug] UI Settings "Group Roms" does not initially take effect
**RomM version**
3.5.1

**Describe the bug**
On a fresh install, the "Group Roms" setting displays as True, however roms are not grouped unless the setting is re-toggled.

**Additional context**
I can see in the code that the logic to read the setting is not consistent with the other UI Settings (i.e. default to true if the returned value is null/never been set).  I can submit a quick pull request to make this consistent.

MattTW:
Makes the read from localStorage consistent with reads of the other settings elsewhere in the codebase.

Would probably be nice to have grouping happen in the "recents" view too, but that would take more substantial changes and a deeper understanding of the codebase by me.  :)   If I get more time, maybe I will submit a separate PR for that.

### DIFF
--- a/frontend/src/stores/roms.ts
+++ b/frontend/src/stores/roms.ts
@@ -3,7 +3,7 @@ import type { DetailedRomSchema, SimpleRomSchema } from "@/__generated__/";
 import { type Platform } from "@/stores/platforms";
 import { type Collection } from "@/stores/collections";
 import type { ExtractPiniaStoreType } from "@/types";
-import { groupBy, uniqBy } from "lodash";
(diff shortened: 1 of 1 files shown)


zurdi15:
Awesome! Thank you :)

zurdi15 (approved):

zurdi15:
Can you reopen it and target ``master`` branch please?

MattTW:
Oops!  I can't reopen since it is merged, it seems.  I will do a new one.

PR #1289, opened 2024-11-15
Update rom.py other filename encode error

vincywindy:
https://github.com/rommapp/romm/pull/1274
I found a few other things that need fixing


adamantike:
Thanks for your contribution, @vincywindy!

The repository configuration could be a bit confusing, but `master` is the branch with the latest progress, while `release` includes the code for the latest stable release.

These issues should be already fixed in `master`, and just waiting for the next stable release.

PR #1296, opened 2024-11-16
[ROMM-1292] Log when db entries are purged

### LINKED ISSUE #1292: [Feature] Add log information when rom metadata purged
Currently, when RomM scans a platform it will remove roms from the database that are missing from the filesystem. However, it does not make a log entry in the console when it does so.

I would like RomM to log whenever it purges rom metadata from the database.

**Additional context**
This was brought up during a discussion on whether or not RomM would retroactively remove rom metadata for roms that were newly filtered via the config.yml. The lack of logging created confusion on when RomM was or was not actually removing rom metadata.

gantoine:
Fixes #1292

### DIFF
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -168,7 +168,7 @@ async def scan_platforms(
         # Same protection for platforms
         if len(fs_platforms) > 0:
             log.info("Purging platforms not found in the filesystem:")
-            log.info("\n  - ".join(fs_platforms))
+            log.info("\n".join([f"  - {platform}" for platform in fs_platforms]))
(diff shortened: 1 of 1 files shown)


backend/endpoints/sockets/scan.py
```diff
@@ -167,6 +167,8 @@ async def stop_scan():
 
         # Same protection for platforms
         if len(fs_platforms) > 0:
+            log.info("Purging platforms not found in the filesystem:")
+            log.info("\n  - ".join(fs_platforms))
```
adamantike: I think we should use this to include the bullet in the first item:
```python
log.info("\n".join([f"  - {platform}" for platform in platforms])
```

adamantike (approved):

Issue #1208, opened 2024-09-23
[Bug] Tinfoil endpoint fails when using directories on file system


therobbiedavis:
**RomM version**
3.5.1

**Describe the bug**
When using directories to organize switch ROMs, the tinfoil API does not understand

**To Reproduce**

1. Use a directory structure like /roms/ARMS/ARMS [titleid].nsp
2. Scan this into RomM
3. Navigate to this via Tinfoil

**Expected behavior**
I figured romm would act as a proxy of sorts to obfuscate/flatten the structure for whatever structure tinfoil requires

**Screenshots**
Using directories:
![image](https://github.com/user-attachments/assets/1f035401-0543-423c-809d-44a65c1f4a06)
![image](https://github.com/user-attachments/assets/d229a0ce-ad5a-46b3-b364-a36e681ea922)
![image](https://github.com/user-attachments/assets/2909fff7-1a58-4001-9265-27fb0e360a67)

Using flat structure:
![image](https://github.com/user-attachments/assets/6d5b8980-a8e7-4dfa-bb04-1b2ed3ca0892)
![image](https://github.com/user-attachments/assets/bf2b7ab0-8315-43a2-abe6-e682395b2883)

barto28:
I encounter this issue as well.

Josh19002:
I will mention that this issue persists with the new file system introduced in 3.8.0, in fact, the preferred structure will definitely break this entirely.

Josh19002:
I wasn't really able to determine the best way to show how I structured everything in minimal screenshots with my actual files, due to how each game may or may not have the related subfolders, but I recreated it to show what the possible combinations are.

This screenshot shows the switch directory (simplified of course), showing loose games that do not have other associated files and a folder containing a game with multiple files:

![Image](https://github.com/user-attachments/assets/f0c419e3-2a37-40fc-9273-cffd716e9d86)

This screenshot shows an example of the game folder previously mentioned, showing a base game file and subfolders relating to different file types:

![Image](https://github.com/user-attachments/assets/01b8d508-3759-4d33-aa4f-38dc4f4c9347)

In the DLC folder, it specifically contains the files added to a base game as DLC:

![Image](https://github.com/user-attachments/assets/f8c563b4-1a63-41cf-89ae-31cf583b3e19)

In the mods folder it contains zip files of mod sets as applicable to the switch root (Tinfoil doesn't do these):

[...]

PR #1315, opened 2024-11-23
Fix single files in dir for tinfoil

KaitoKid:
Addresses the concerns in #1208 

### Tinfoil Parsing Issues

Root cause is the FS handler for get_roms assumes that directories with a single file inside as `multi`

```
.
├── Homebrew
│   ├── my_homebrew.nsp
```

which results in the following FSRom being created after the directory is scanned

```
{
  'multi': True, 
  'file_name': 'Homebrew', 
  'files': [{'filename': 'my_homebrew.nsp', 'size': 0, 'last_modified': 1718254179.6487029}]
}
```

This fix addresses two things:

1. Single files in directories being mistakenly addressed as `multi`
2. Changing the file path to be the actual path of the file, not the `file_name`. Instead, it renames the file in Tinfoil to the `file_name` instead.

### Known Bugs

Since this takes just the first obj of each rom file, it cannot handle cases where multiple files are stored in one directory. I'll defer to the community on the best method of handling this, but a couple ideas come to mind:

1. Creating separate directories at the root level for base/patch/update
2. Standardizing file names so that titledb/tinfoil's api/something can be matched up


gantoine:
@KaitoKid could you reopen this against the master branch?

KaitoKid:
My rebase onto master (since I branched off release) went horribly wrong, gimme a sec to fix this

PR #1316, opened 2024-11-23
fix: Fix tinfoil issue with improper rom DL urls and support for rom folders

KaitoKid:
Addresses the concerns in #1208 

### Tinfoil Parsing Issues

Root cause is the FS handler for get_roms assumes that directories with a single file inside as `multi`

```
.
├── Homebrew
│   ├── my_homebrew.nsp
```

which results in the following FSRom being created after the directory is scanned

```
{
  'multi': True, 
  'file_name': 'Homebrew', 
  'files': [{'filename': 'my_homebrew.nsp', 'size': 0, 'last_modified': 1718254179.6487029}]
}
```

This fix addresses two things:

1. Single files in directories being mistakenly addressed as `multi`
2. Changing the file path to be the actual path of the file, not the `file_name`. Instead, it renames the file in Tinfoil to the `file_name` instead.

### Known Bugs

Since this takes just the first obj of each rom file, it cannot handle cases where multiple files are stored in one directory. I'll defer to the community on the best method of handling this, but a couple ideas come to mind:

1. Creating separate directories at the root level for base/patch/update
2. Standardizing file names so that titledb/tinfoil's api/something can be matched up


KaitoKid:
Added a few more updates to address the following:

* Handling empty folders
* Addressing `rom.full_path` using the `file_name`, which is stored as the rom directory name (causes an inaccurate file redirect upon download
* Rolled back changes to the filename. I was hoping to rename it as "Rom/rom_file.nsp" within Tinfoil, but I can see that it would not pass routing rules. Leaving it without an extension causes Tinfoil to open it as a file, not as an installer

backend/handler/filesystem/roms_handler.py
```diff
@@ -348,9 +348,13 @@ def get_roms(self, platform_fs_slug: str) -> list[FSRom]:
 
         return [
             FSRom(
-                multi=rom["multi"],
```
adamantike: I don't think this PR should make any changes to the filesystem's rom handler. Let's notice that this logic makes the `multi` value we save in the database not always match what we expose here, which will add complexity and confusion to future implementations.

Deciding that `multi` should be false when a rom is within a folder but is a single file is a different discussion, and if agreed, should be fixed at the repository level, so the value saved as `multi` in the database is correct.

Still, I think the fix for Tinfoil can be implemented without touching this file.

backend/endpoints/feeds.py
```diff
@@ -161,12 +161,13 @@ async def extract_titledb(roms: list[Rom]) -> dict[str, TinfoilFeedTitleDBSchema
             TinfoilFeedFileSchema(
                 url=str(
                     request.url_for(
-                        "get_rom_content", id=rom.id, file_name=rom.file_name
+                        "get_rom_content", id=rom.id, file_name=file.get('filename')
                     )
                 ),
-                size=rom.file_size_bytes,
+                size=rom.files[0].get('size'),
```
adamantike, line 167: `rom.files[0]` here now needs to be `file`, after the latest changes.

backend/endpoints/feeds.py
```diff
@@ -161,12 +161,13 @@ async def extract_titledb(roms: list[Rom]) -> dict[str, TinfoilFeedTitleDBSchema
...
             )
-            for rom in roms
+            for rom in roms if rom.files and len(rom.files) > 0
+            for file in rom.files
```
adamantike, line 170: I don't think we want to expose all files to Tinfoil. Ideally, we could only include the ones with a valid extension that Tinfoil understands (in case users maintain other kind of resources within the rom folder).

adamantike:
@KaitoKid, please let us know if you plan to keep updating this PR.

therobbiedavis:
I see development of this has dropped off. Unfortunately the lack of rom folder support really makes the tinfoil integration unusable at scale. I am available for testing but unfortunately I lack python knowledge so I can not contribute to the code in any meaningful way.

gantoine:
@therobbiedavis I'm sure if you take a crack at it you could fix it in no time, python knowledge or not ;) As for this PR I'm going to close it as we'll be reworking the folder structure post 3.8.0 to support nested folders, DLC and patch files.

KaitoKid:
@adamantike - apologies, got caught up with some personal life, and picking this back up. 

A couple of contraints w/ Tinfoil which is seemingly no longer updated:
1. It doesn't support query params so we can't pass in file_ids
2. It uses the final path segment as the display name inside

I'm trying to create a solution that modifies the most minimal amount of code possible, but I think we have to pass in a direct url to the file for a multi-file rom, e.g. `http:/romm/api/roms/{rom_id}/content/homebrew/{romfile.file_name}.nsp` which I think requires a modification to the `roms` endpoint

Any chance you could help point me in the right direction here? I'll create a fresh PR once I have a bit more than the below:

```
    return TinfoilFeedSchema(
        files=[
            TinfoilFeedFileSchema(
                url=str(
                    request.url_for("get_rom_content", id=rom.id, file_name=rom_file.file_name) + f"?file_ids={rom_file.id}"
                ),
                size=rom_file.file_size_bytes,
            )
            for rom in roms if rom.files
            for rom_file in rom.files if rom_file.file_name.lower().endswith(('.xci', '.nsp', '.nsz', '.xcz', '.nro'))
        ],
        directories=[],
        success="RomM Switch Library",
        titledb=await extract_titledb(roms),
    )

```

gantoine:
Replaced by https://github.com/rommapp/romm/pull/1754

PR #1321, opened 2024-11-28
OIDC Support

gantoine:


backend/endpoints/auth.py
```diff
@@ -17,6 +25,58 @@
...
+        request (Request): Fastapi Request object
+
+    Returns:
+        MessageResponse: Standard message response
+    """
+
+    request.session.clear()
+
+    return {"msg": "Successfully logged out"}
+
```
gantoine, line 78: this is unchanged, just moved it up

backend/main.py
```diff
@@ -102,7 +102,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
 app.include_router(states.router, prefix="/api")
 app.include_router(tasks.router, prefix="/api")
 app.include_router(feeds.router, prefix="/api")
-app.include_router(config.router, prefix="/api")
+app.include_router(configs.router, prefix="/api")
```
gantoine, line 105: this is to stop vscode complaining when getting confused between the `backend/config` and `config/__init__`

pytest.ini
```diff
@@ -16,3 +16,4 @@ env =
    SCHEDULED_UPDATE_SWITCH_TITLEDB_CRON=0 4 * * *
    UPLOAD_TIMEOUT=20
    LOGLEVEL=DEBUG
+   OIDC_ENABLED=false
```
gantoine, line 19: we set this for all tests so they run, but override it in OIDC handler tests

zurdi15:
Just added the new ``OIDC_`` env variables to the development template

backend/config/__init__.py
```diff
@@ -76,6 +76,13 @@ def str_to_bool(value: str) -> bool:
     os.environ.get("DISABLE_DOWNLOAD_ENDPOINT_AUTH", "false")
 )
 
+# OIDC
+OIDC_ENABLED: Final = os.environ.get("OIDC_ENABLED", "false") == "true"
```
adamantike: nit: we can use `str_to_bool` for consistency here.

backend/handler/auth/base_handler.py
```diff
@@ -164,3 +156,55 @@ async def get_current_active_user_from_bearer_token(self, token: str):
...
+        with httpx.Client() as httpx_client:
+            try:
+                response = httpx_client.get(jwks_url, timeout=120)
+                key = response.json()["keys"][0]
+                self.rsa_key = RSAKey.import_key(key)
+            except httpx.HTTPStatusError as exc:
+                raise HTTPException(
+                    status_code=exc.response.status_code,
+                    detail=exc.response.text,
+                ) from exc
```
adamantike: I think it would be better to move this logic to a method, to avoid us triggering an HTTP request at import time (even if the `get_current_active_user_from_openid_token` method is never called, like in the `watcher.py` process).

```python
    async def get_rsa_key(self) -> RSAKey:
        if not getattr(self, "_rsa_key", None):
            try:
                response = await httpx_client.get(jwks_url, timeout=120)
                key = response.json()["keys"][0]
                self._rsa_key = RSAKey.import_key(key)
            except httpx.HTTPStatusError as exc:
                raise HTTPException(
                    status_code=exc.response.status_code,
                    detail=exc.response.text,
                ) from exc
        return self._rsa_key
```

adamantike (commented):
Great implementation! Just added a few small comments.

adamantike (approved):

jvanbruegge:
Nice, thanks for this!

PR #1323, opened 2024-11-28
fix: Crash on new instance from previous installation

### LINKED ISSUE #1266: [Bug] Romm does not correctly reject invalid session keys
**RomM version**
rommapp/romm:latest

**Describe the bug**
Romm does not check if a session key is invalid and reset, instead shows a corrupted interface

**To Reproduce**

1. set up romm and create a user account, then log in
2. reset romm, or set up romm on a new server that points to the same url
2a. Theoretically make a fake session key and pose as a hacker attempting to illegally access romm
4. Romm will launch to the full interface, but with all user specific sections blank. logging out will not redirect to the setup page, but opening the site in a different browser with no previous cookies or cache will launch the setup page

**Expected behavior**
romm should redirect to the login page (or a setup page if not previously set up) whenever an invalid session key is given by the user

zurdi15:
When setting up a new installation, if the browser has old cookies (romm_session cookie specifically) from an old instance or from an user that no longer exists in the database, RomM crashes without recover. This fix adds a check in the middleware to avoid the crash and clean the session cookies gracefully.

Closes #1266

### DIFF
--- a/backend/handler/auth/middleware.py
+++ b/backend/handler/auth/middleware.py
@@ -1,6 +1,8 @@
 import time
 from collections import namedtuple
 
+from fastapi import HTTPException
+from handler.database.users_handler import DBUsersHandler
 from joserfc import jwt
 from joserfc.errors import BadSignatureError
 from joserfc.jwk import OctKey
@@ -107,7 +109,15 @@ class SessionMiddleware:
                 jwt_claims = self._validate_jwt_payload(jwt_payload)
                 scope["session"] = jwt_claims
                 initial_session_was_empty = False
-            except BadSignatureError:
+
+                # Check if the user exists
+                db_user_handler = DBUsersHandler()
+                user = db_user_handler.get_user_by_username(jwt_claims.get("sub"))
+                if not user:
+                    scope["session"] = {}
+                    initial_session_was_empty = True
+
+            except (BadSignatureError, HTTPException):
                 scope["session"] = {}
         else:
             scope["session"] = {}
--- a/backend/handler/auth/middleware.py
+++ b/backend/handler/auth/middleware.py
@@ -1,7 +1,6 @@
 import time
 from collections import namedtuple
 
-from fastapi import HTTPException
 from joserfc import jwt
 from joserfc.errors import BadSignatureError
 from joserfc.jwk import OctKey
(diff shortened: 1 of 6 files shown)


backend/handler/auth/middleware.py
```diff
@@ -107,7 +109,15 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
                 scope["session"] = jwt_claims
                 initial_session_was_empty = False
-            except BadSignatureError:
+
+                # Check if the user exists
+                db_user_handler = DBUsersHandler()
+                user = db_user_handler.get_user_by_username(jwt_claims.get("sub"))
+                if not user:
+                    scope["session"] = {}
+                    initial_session_was_empty = True
```
gantoine: making a db request on every request is very expensive, and i think there's a better way to handle this (though i cant think of it right now)...

zurdi15: Yeah I know, we need to figure out this one

zurdi15: maybe we can make the db request only in the heartbeat instead on every request? that will reduce the database requests but will clear session when refreshing the page if the user no longer exists

zurdi15: i tested it and it won't work, if any requests is made before the heartbeat it will crash

zurdi15: Okay fixed! those changes should do the trick. I tested removing and/or disabling users with an existing session cookie and works like a charm without requesting the database

backend/handler/auth/middleware.py
```diff
@@ -107,7 +108,8 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                 jwt_claims = self._validate_jwt_payload(jwt_payload)
                 scope["session"] = jwt_claims
                 initial_session_was_empty = False
-            except BadSignatureError:
+
+            except (BadSignatureError, HTTPException):
```
gantoine: i don't think any calls in here will throw an `HTTPException`, do you can probably remove it.

gantoine (approved):
yeah good fix!

backend/handler/auth/middleware.py
zurdi15: You are right!

PR #1324, opened 2024-11-28
feat: Add platform info drawer + per platform settings

zurdi15:
- New drawer with platform info added
- Removed some unnecessary API calls
- Adapted covers to platform selected aspect ratio
- Added platform icon in the recently added section
- Added action bar to the search dialog
- Now the last visited platform is stored in memory, loading it faster when going back to it from another view

TODO:
- [x]  Cover aspect-ratio per platform


backend/models/platform.py
```diff
@@ -21,13 +24,23 @@ class Platform(BaseModel):
...
+    url: Mapped[str | None] = mapped_column(String(length=1000), default="")
+    url_logo: Mapped[str | None] = mapped_column(String(length=1000), default="")
     logo_path: Mapped[str | None] = mapped_column(String(length=1000), default="")
 
     roms: Mapped[list[Rom]] = relationship(back_populates="platform")
     firmware: Mapped[list[Firmware]] = relationship(
         lazy="selectin", back_populates="platform"
     )
 
+    aspect_ratio: Mapped[float] = mapped_column(
```
adamantike: I think having the `aspect_ratio` as a small char column will be simpler to maintain in the long term, to avoid issues with floating point precision. We could always handle it as `<width_units>:<height_units>` for simplicity (e.g. `16:9`), and if for some reason we need its decimal representation, parsing and calculating it would be very simple.

zurdi15: Okay sounds reasonable, let me try to handle it as ``<width_units>:<height_units>``

zurdi15: Done, changed to char and converting in the frontend properly

backend/handler/metadata/igdb_handler.py
```diff
@@ -745,6 +785,7 @@ async def get_oauth_token(self) -> str:
     "similar_games.name",
     "similar_games.cover.url",
     "age_ratings.rating",
+    "videos.video_id",
```
adamantike, line 788: 💯

adamantike (approved):

Issue #1338, opened 2024-12-02
[Feature] Changing UID/GID of romm via env variable.


Scorcerer:
**Is your feature request related to a problem? Please describe.**
The current owner of user romm within container is hardcoded to UID/GID 1000 - It'd be great to make it adjustable with environment variables like PUID/GUID.

**Describe the solution you'd like**
Well, I set the PUID/PGID to a number and automagically romm user has that ID inside

**Describe alternatives you've considered**
Changing owner of all my files to comply with the requirements of this container (and i have some containers already, so it's gonna be breaking lots of things.

**Additional context**
I know it's not a nice thing to say, but all the *arrs do have that functionality...

Scorcerer:
I saw just now how simple it was and whipped out a small patch :P

gantoine:
`PUID/PGID` is a linuxserver specification, and so isn't supported by romm. You can use the `user: XXXX:XXXX` param in `docker-compose.yaml` to run the container under a different user.

PR #1339, opened 2024-12-02
Add possibility to change romm UID/GID

Scorcerer:
Solves #1338


adamantike:
This is not as easy, because these arguments are only evaluated at build time, not during runtime. That means that this change won't make any difference to the images we publish to Docker registries.

Scorcerer:
ah, dang, I understand. Hmmmm...

PR #1345, opened 2024-12-05
fix: Update ROM deletion process to remove associated collections

zurdi15:


backend/endpoints/rom.py
```diff
@@ -497,6 +497,14 @@ async def delete_roms(
         log.info(f"Deleting {rom.file_name} from database")
         db_rom_handler.delete_rom(id)
 
+        # Update collections to remove the deleted rom
+        collections = db_collection_handler.get_collections_by_rom_id(id)
+        for collection in collections:
+            collection.roms = [rom_id for rom_id in collection.roms if rom_id != id]
+            db_collection_handler.update_collection(
+                collection.id, {"roms": collection.roms}
+            )
```
gantoine, line 506: should this instead happen in `db_rom_handler.delete_rom` as a side-effect?

zurdi15: hmm idk, I stick to the idea of just interacting with one table per handler, but idk if that's the best approach

gantoine (approved):

adamantike:
The current solution looks good. An improvement we can add in the future, to avoid any race conditions and make the update atomic, is to have `add_roms`/`remove_roms` method at the handler, and generate a single SQL statement for the operation.

adamantike (approved):

PR #1346, opened 2024-12-07
misc: Fix some typing issues

adamantike:


backend/config/__init__.py
```diff
@@ -110,6 +110,8 @@ def str_to_bool(value: str) -> bool:
 
 # LOGGING
 LOGLEVEL: Final = os.environ.get("LOGLEVEL", "INFO")
+FORCE_COLOR: Final = str_to_bool(os.environ.get("FORCE_COLOR", "false"))
+NO_COLOR: Final = str_to_bool(os.environ.get("NO_COLOR", "false"))
```
zurdi15, line 114: This is not better to set it as bool by defaut instead of converting from string?

gantoine: iirc all env variables are read in as bools, so they need to be converted anyways

gantoine (approved):

backend/config/__init__.py
adamantike: Environment variables always come as strings, so we need to convert to boolean. The `str_to_bool` util just handles the conversion to lowercase and comparison to what we consider a truthy value.

Issue #492, opened 2023-12-09
[Feature] Load IGDB id/secret from file


rthill91:
**Is your feature request related to a problem? Please describe.**
I don't like to put secrets directly in my docker-compose file and I'd rather not have to.

**Describe the solution you'd like**
I would love to have the additional (optional) environment variables `IGDB_CLIENT_ID_FILE` and `IGDB_CLIENT_SECRET_FILE` such that my docker-compose file could look something like
```yaml
secrets:
  igdb_id:
    file: ./igdb_id
  igdb_secret:
    file: ./igdb_secret

services:
  romm:
    ...
    secrets:
      - igdb_id
      - igdb_secret
    environment:
      - IGDB_CLIENT_ID_FILE=/run/secrets/igdb_id
      - IGDB_CLIENT_SECRET_FILE=/run/secrets/igdb_secret
```
_similar to [traefik config](https://doc.traefik.io/traefik/user-guides/docker-compose/acme-dns/#use-secrets)_

**Describe alternatives you've considered**
N/A

**Additional context**
I've done a cursory look through how config is loaded and I _think_ this would be a very easy feature to add with something like
```py3
# IGDB
_igdb_client_id: str = os.environ.get("IGDB_CLIENT_ID", os.environ.get("CLIENT_ID", ""))
_igdb_client_secret: str = os.environ.get(
    "IGDB_CLIENT_SECRET", os.environ.get("CLIENT_SECRET", "")
)
if not _igdb_client_id:
    if (
        _igdb_client_id_file := os.environ.get("IGDB_CLIENT_ID_FILE", "")
    ) and os.path.exists(_igdb_client_id_file):
        with open(_igdb_client_id_file) as fh:
            _igdb_client_id = fh.readline().strip()
if not _igdb_client_secret:
    if (
        _igdb_client_secret_file := os.environ.get("IGDB_CLIENT_SECRET_FILE", "")
    ) and os.path.exists(_igdb_client_secret_file):
        with open(_igdb_client_secret_file) as fh:
            _igdb_client_secret = fh.readline().strip()
IGDB_CLIENT_ID: Final = _igdb_client_id
IGDB_CLIENT_SECRET: Final = _igdb_client_secret
```
though admittedly I've done almost no real testing of it.

gantoine:
You shouldn't keeps secrets in the docker-compose file directly, instead use an `.env` file stored in the same directory:

`docker-compose.yml`
```
services:
  romm:
    ...
    env_file: .env
[...]

PR #1349, opened 2024-12-08
feat: Add support for _FILE suffix in environment variables

adamantike:
This change allows setting environment variables with a `_FILE` suffix, which will be used to load the contents of the file specified in the variable into the variable without the suffix.

For example, setting `ROMM_AUTH_SECRET_KEY_FILE=/run/secrets/romm_auth_secret_key` and creating a file with the secret key at the specified path will set `ROMM_AUTH_SECRET_KEY` to the contents of the file.

A common use case for this is to use secrets in Docker Compose [1], to avoid exposing secrets in the `docker-compose.yml` or `env` files.

[1] https://docs.docker.com/compose/how-tos/use-secrets/

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -152,5 +152,5 @@ VOLUME ["/romm/resources", "/romm/library", "/romm/assets", "/romm/config", "/re
 EXPOSE 8080 6379/tcp
 WORKDIR /romm
 
-ENTRYPOINT ["/sbin/tini", "--"]
+ENTRYPOINT ["/sbin/tini", "--", "/docker-entrypoint.sh"]
 CMD ["/init"]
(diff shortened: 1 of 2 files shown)


zurdi15 (approved):

gantoine (approved):

kuya1284:
Thank you for this. I noticed that there was no mention of either a `FILE_` prefix or `_FILE` suffix on the [Environment Variables Wiki Page](https://github.com/rommapp/romm/wiki/Environment-Variables), so decided to search through the issues. I finally discovered this by way of #492. Thank you for adding support for this because I'd much rather leverage Docker Secrets like I do with my other containers. Storing them in `.env` as suggested in #492 wouldn't be a good idea because there are other environment variables that could get stored in there and the `.env` file often gets committed to a repo.

Would it be possible to make mention of the `_FILE` suffix to the Wiki to make it easier for others? Thank you again for this major improvement.

gantoine:
@kuya1284 done

PR #1350, opened 2024-12-09
[ROMM-1348] Log when a user downloads a game

gantoine:


adamantike:
I'm getting a 500 error when the request has no user. To reproduce:
1. Set `DISABLE_DOWNLOAD_ENDPOINT_AUTH=true`.
2. Copy the download link for a game.
3. Open an incognito tab and access the link from there.

```
  File "/backend/endpoints/rom.py", line 258, in get_rom_content
    log.info(f"User {current_user.username} is downloading {rom.file_name}")
                     ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'username'
```

zurdi15 (commented):
It's not better to just have one log before all the ifs and returns instead of inside each one?

zurdi15 (approved):

PR #1355, opened 2024-12-11
fix: Setup wizard redirection

zurdi15:


danblu3:
Tested by the way, worked fine.

gantoine (approved):

PR #1369, opened 2024-12-21
Rom file and hashing refactor

gantoine:
Alright this bad boy is ready for review! The gist of changes can be found in the migration file, but basically this PR:
* Creates a model for `RomFile` and migrates `rom.files` over to it
* Renames `file_*` columns on `Rom` to `fs_*`, to clarify that these are filesystem properties
* Runs hash calculations in a background task separate from the scan
* Deeply scans game folders for content like DLC, patches, etc.

I've also cleaned up a lot of pylint errors, added extra checks all over the place for bad data/states, and added `HASH_SCAN` as a scan type in the UI.


backend/endpoints/sockets/scan.py
```diff
@@ -23,20 +24,50 @@
...
+        "psvr",
+        "psvr2",
+        "series-x",
+        "switch",
+        "wiiu",
+        "win",
+        "xbox-360",
+        "xboxone",
+    )
+)
```
gantoine, line 71: moved the location of this list and populated it with more platforms

backend/endpoints/firmware.py
```diff
@@ -70,10 +73,14 @@ def add_firmware(
         uploaded_firmware.append(scanned_firmware)
 
     db_platform = db_platform_handler.get_platform(platform_id)
```
adamantike: It seems we can avoid fetching the Platform again here, if we do the same query as `get_platform_firmware`:

```python
return {
    "uploaded": len(files),
    "firmware": [
        FirmwareSchema.model_validate(f)
        for f in db_firmware_handler.list_firmware(platform_id=platform_id)
    ]
}
```

backend/endpoints/rom.py
```diff
@@ -255,14 +257,14 @@ async def get_rom_content(
     request: Request,
     id: int,
     file_name: str,
-    files: Annotated[list[str] | None, Query()] = None,
+    file_ids: list[str] | None = None,
```
adamantike: This should match the `list[int]` type from the previous endpoint.

backend/endpoints/rom.py
```diff
@@ -424,26 +425,26 @@ async def update_rom(
         }
     )
 
-    new_file_name = data.get("file_name", rom.file_name)
+    new_fs_name = str(data.get("fs_name", rom.fs_name))
```
adamantike: Let's use `str(data.get("fs_name") or rom.fs_name)` to avoid potential `"None"` issues.

backend/endpoints/rom.py
```diff
@@ -424,26 +425,26 @@ async def update_rom(
...
 
     try:
         if rename_as_source:
-            new_file_name = rom.file_name.replace(
-                rom.file_name_no_tags or rom.file_name_no_ext,
-                data.get("name", rom.name),
+            new_fs_name = rom.fs_name.replace(
+                rom.fs_name_no_tags or rom.fs_name_no_ext,
+                str(data.get("name", rom.name)),
```
adamantike: Same here.

backend/endpoints/rom.py
```diff
@@ -497,15 +496,18 @@ async def update_rom(
                 path_cover_s, path_cover_l = await fs_resource_handler.get_cover(
                     overwrite=True,
                     entity=rom,
-                    url_cover=data.get("url_cover", ""),
+                    url_cover=str(data.get("url_cover", "")),
```
adamantike: Same here.

backend/endpoints/rom.py
```diff
@@ -535,13 +537,15 @@ async def delete_roms(
...
-        log.info(f"Deleting {rom.file_name} from database")
+        log.info(f"Deleting {rom.fs_name} from database")
         db_rom_handler.delete_rom(id)
 
         # Update collections to remove the deleted rom
         collections = db_collection_handler.get_collections_by_rom_id(id)
         for collection in collections:
-            collection.roms = [rom_id for rom_id in collection.roms if rom_id != id]
+            collection.roms = set(
+                [rom_id for rom_id in collection.roms if rom_id != id]
```
adamantike: nit: no need for the square brackets here, so Python doesn't create a temporary list before building the set. Instead:

```python
collection.roms = {rom_id for rom_id in collection.roms if rom_id != id}
```

backend/endpoints/sockets/scan.py
```diff
@@ -333,6 +366,50 @@ async def _identify_firmware(
...
+
+    Args:
+        rom_id (int): Rom id
+    """
+    rom = db_rom_handler.get_rom(rom_id)
+    if not rom:
+        return
+
+    try:
+        rom_hash, rom_file_hashes = fs_rom_handler.get_rom_hashes(rom)
```
adamantike, line 381: I would only put this line in the `try` statement, and the hash saving logic in an `else:` after the except, as the exception we are handling can only occur here.

backend/endpoints/sockets/scan.py
```diff
@@ -376,6 +447,37 @@ async def _identify_rom(
 
     _added_rom = db_rom_handler.add_rom(scanned_rom)
 
+    # Create each file entry for the rom
+    db_rom_files = db_rom_handler.get_rom_files(_added_rom.id)
+    for file in fs_rom["files"]:
+        db_rom_file = next(
+            (f for f in db_rom_files if f.file_name == file.file_name), None
+        )
```
adamantike: This could become really slow for games with multiple files. We could build a `db_rom_file_by_file_name` dictionary before the loop.

Also, same here about files with same filename for a rom.

backend/endpoints/rom.py
```diff
@@ -277,46 +279,47 @@ async def get_rom_content(
...
         ZipContentLine(
-            # TODO: Use calculated CRC-32 if available.
-            crc32=None,
-            size_bytes=(await Path(f"{rom_path}/{f}").stat()).st_size,
-            encoded_location=quote(f"/library-zip/{rom.full_path}/{f}"),
-            filename=f,
+            crc32=f.crc_hash,
+            size_bytes=(await Path(LIBRARY_BASE_PATH, f.full_path).stat()).st_size,
+            encoded_location=quote(f"/library-zip/{f.full_path}"),
+            filename=f.file_name,
```
adamantike: I think this will generate issues if you have two files with the same name for a game, but in different subfolders. For example:

```
/platform/game1/dlc/update_1.zip
/platform/game1/hack/update_1.zip
```

I think we should test `f.full_path` here, which should correctly build the same folder structure inside of the zip file, and avoid filename shadowing.

backend/endpoints/rom.py
```diff
@@ -277,46 +279,47 @@ async def get_rom_content(
...
         )
-        for f in files_to_download
+        for f in files
     ]
 
-    m3u_encoded_content = "\n".join([f for f in files_to_download]).encode()
+    m3u_encoded_content = "\n".join([f.file_name for f in files]).encode()
```
adamantike: Same here about files with same filename.

backend/handler/database/roms_handler.py
```diff
@@ -291,4 +296,35 @@ def update_rom_user(self, id: int, data: dict, session: Session = None) -> RomUs
...
+    def get_rom_file_by_id(self, id: int, session: Session = None) -> RomFile | None:
+        return session.scalar(select(RomFile).filter_by(id=id).limit(1))
+
+    @begin_session
+    def get_rom_file_by_file_name(
+        self, rom_id: int, file_name: str, session: Session = None
+    ) -> RomFile | None:
+        return session.scalar(
+            select(RomFile).filter_by(rom_id=rom_id, file_name=file_name).limit(1)
+        )
```
adamantike: Same here about files with same filename for a rom. This could return a list of multiple `RomFile`, unless we change it to be by `full_path`.

backend/handler/filesystem/roms_handler.py
```diff
@@ -229,40 +242,69 @@ def _exclude_multi_roms(self, roms: list[str]) -> list[str]:
...
+    def _build_rom_file(self, rom_path: Path, file_name: str) -> RomFile:
+        # Absolute path to roms
+        abs_file_path = Path(LIBRARY_BASE_PATH, rom_path, file_name)
+
+        path_parts_lower = list(map(str.lower, rom_path.parts))
+        matching_category = next(
+            (
+                category
+                for category in RomFileCategory
+                if category_matches(category.value, path_parts_lower)
```
adamantike, line 254: For a future improvement, I think the way a category matches should be configurable by the user. They could provide a mapping in case they prefer to use a different term for each category.

First example that comes to mind is a different language: `patch: "parche"`

backend/models/rom.py
```diff
@@ -29,10 +29,39 @@
...
+    sha1_hash: Mapped[str | None] = mapped_column(String(100))
+    category: Mapped[RomFileCategory | None] = mapped_column(
+        Enum(RomFileCategory), default=None
+    )
+
+    rom: Mapped[Rom] = relationship(lazy="joined")
+
+    @cached_property
+    def full_path(self) -> str:
+        return f"{self.file_path}/{self.file_name}"
```
adamantike, line 64: nit: This could return a `Path`, and use a notation like `Path(self.file_path) / self.file_name`

backend/endpoints/rom.py
gantoine: huh TIL

backend/handler/filesystem/roms_handler.py
gantoine: Sure something to look into, but if we're already being strict about the filesystem structure there's no reason we can't mandate specific names for DLC/folders/etc.

backend/endpoints/rom.py
gantoine: You were right, it did break! Fix is in one of the latest commits.

adamantike (approved):
Amazing work, looking forward to everything this change unblocks! :rocket:

gantoine:
Thanks for the review! I think we should release a patch version in a couple days with the fixed we have in the pipeline, then I'll merge this in. That way this won't block anything if something major is found.

PR #1378, opened 2024-12-23
OIDC Fixes Part I

### LINKED ISSUE #1377: OIDC support: hardcoded presumption that is only valid for Authentik
https://github.com/rommapp/romm/blob/9d90749122f475460be8ac7bb7911c645f8f12fa/backend/handler/auth/base_handler.py#L176

The OIDC functionality assumes that the JWKS URL is "{OIDC_SERVER_APPLICATION_URL}/jwks/"

This is incorrect behaviour - the JWKS URL can be anywhere. It should be configurable, or - better - be autodiscovered using OpenID Connect Discovery.

gantoine:
* Add `OICD_PROVIDER` env variable and display on login page with icon from CDN
* Fetch JWKS url from `/.well-known/openid-configuration` endpoint of SSO provider

cc @dataway
Fixes #1377


backend/handler/auth/base_handler.py
```diff
@@ -165,16 +165,44 @@ class OpenIDHandler:
...
+        )
+        log.info("Fetching server metadata from %s", server_metadata_url)
+
+        httpx_client = ctx_httpx_client.get()
+        try:
+            response = await httpx_client.get(server_metadata_url, timeout=120)
+            response.raise_for_status()
+
+            self._server_metadata = response.json()
+            return response.json()
```
adamantike: We can avoid a second JSON unmarshal, by just returning `self._server_metadata` here.

backend/config/__init__.py
```diff
@@ -78,6 +78,7 @@ def str_to_bool(value: str) -> bool:
 
 # OIDC
 OIDC_ENABLED: Final = str_to_bool(os.environ.get("OIDC_ENABLED", "false"))
+OICD_PROVIDER: Final = os.environ.get("OICD_PROVIDER", "")
```
dataway: I think you misspelled "OIDC" :)

zurdi15: good catch! fixed

zurdi15 (approved):

adamantike (approved):

PR #1381, opened 2024-12-23
feat: Add filter to show matched roms to gallery filter

### LINKED ISSUE #1285: [Feature] Show Matched Roms Only - Filter Option in each console
I want to filer matched roms only. So I don't have to see my un-matched roms I can't find or haven't manually matched.

When I click on each individual console and the it starts to load the collection of roms. I want a filter option to show only matched roms. (Instead of the current options of Show Favorites, Show UnMatched, Show Duplicates)

The only thing I could try is to delete all my un-matched roms using the current app features. There is no easy of moving it to a different folder using Romm or filtering them out of my storage manually to remove from Romm, if I wanted to work on them at a later date.

**Additional context**
Just add a 4th option to show matched only.

image

zurdi15:
Added filter to show only matched games

Closes #1285


gantoine (approved):
💯

PR #1383, opened 2024-12-24
feat: added custom name for platforms

zurdi15:
Now platforms can have a custom name, editing it from the platform info panel


frontend/src/components/Gallery/AppBar/Platform/PlatformInfoDrawer.vue
```diff
@@ -49,13 +49,32 @@ const aspectRatioOptions = computed(() => [
...
+function toggleEditable() {
+  isEditable.value = !isEditable.value;
+}
+
+function updatePlatformName(newName: string) {
+  if (currentPlatform.value) {
+    currentPlatform.value.custom_name = newName;
+    platformApi.updatePlatform({
+      platform: {
+        ...currentPlatform.value,
```
gantoine: you dont have to send all the fields anymore right, just the id?

frontend/src/components/common/Platform/Card.vue
```diff
@@ -17,10 +17,10 @@ defineProps<{ platform: Platform }>();
       <v-card-text>
         <v-row class="pa-1 justify-center bg-primary">
           <div
-            :title="platform.name?.toString()"
+            :title="platform.name.toString()"
             class="px-2 text-truncate text-caption"
           >
-            <span>{{ platform.name }}</span>
+            <span>{{ platform.custom_name || platform.name }}</span>
```
gantoine: this can be done better with the pydantic resolver, i'll give it a crack tomorrow

gantoine (changes requested):
can you rebase the PR to remove all the code for rom multi-select?

frontend/src/components/Gallery/AppBar/Platform/PlatformInfoDrawer.vue
zurdi15: The api method expects the platform object, so we need to send them with the updated properties

frontend/src/components/common/Platform/Card.vue
zurdi15: What do you mean? we still need to send both properties to show them in the platform info panel anyway

gantoine (approved):

PR #1353, opened 2024-12-10
feat: Multi language support

zurdi15:
### Overview
This pull request introduces multi-language support, enhancing its accessibility to a broader audience.

At launch, RomM will support ``english``, ``spanish``, ``french``, ``brazilian portuguese``, ``russian`` and ``german``

### Known Issues
Some translations might still be incomplete or require further refinement.
There may be UI adjustments needed to accommodate different languages.

### Future Improvements
Continue refining translations.
Add support for additional languages based on user feedback.
Ensure UI elements are fully responsive to different language lengths.

PR #1385, opened 2024-12-24
Added Chinese related language files (zh-CN)

NicksGhost:
I just read the update document of 3.7.0-alpha.1 #1353 , and I am very happy that RomM now has localization related features. Thank you very much for your work! I also wrote the language file related to Chinese (zh-CN), hope it can be helpful.


gantoine (approved):
These look great, thanks for the PR!

Issue #1373, opened 2024-12-22
[Bug] Outdated EmulatorJS versiom


EmPiFree:
**RomM version**
3.6.0

**Describe the bug**
I get the error "Outdated EmulatorJS version" when starting a core/rom

**To Reproduce**

1. Start any rom
2. See error

**Expected behavior**
Rom should start

**Screenshots**
![image](https://github.com/user-attachments/assets/58e73846-5bdd-4f97-8568-30cb931c3989)

**Desktop (please complete the following information):**

- OS: Win10
- Browser Brave and Edge
- Version: Most recent

**Additional context**
A new EmulatorJS update just has been released. We probably need to update the integrated EmulatorJS version.

Acciden7Prone92:
I am also experiencing this issue - across all platforms.

gantoine:
The issue is fix in the latest alpha release `3.7.0-alpha.1`, and will be generally available in the `3.7.0` release. I don't recommend upgrading to the alpha as we can't guarantee the image is stable.

r-trigo:
I can see that `3.8.0-alpha-1` includes the EmulatorJS bump in the PR list, however this issue is occurring in my docker container.

gantoine:
upgrade to 3.7.0-alpha.1, stop delete and restart the container, then clear browser cache

MambaNerd:
I am also having the sam issue in my docker contaier

r-trigo:
Sorry, I meant `3.7.0-alpha.1` (`3.8.0-alpha.1` is not released).
Clearing the cache after bumping to `3.7.0-alpha.1` did the trick, thanks 🙏

hannibalshosting88:
Guess i'm having the same issue.  So the only fix is waiting for or upgrading to the alpha?

gantoine:
Correct, though this should be the last time this happens as we're working with the emulatorjs team to make sure it doesn't happen again in the future, when they release next.

whitenet:
updated to 3.7.1-alpha is still error with outdated emulatorJs

gantoine:
@whitenet stop delete and restart the container, then clear browser cache

hannibalshosting88:
Well, whatever any of you do, don't attempt to deploy the alpha If you have this thing running as a Docker in UnRAID.

djismgaming:
> Well, whatever any of you do, don't attempt to deploy the alpha If you have this thing running as a Docker in UnRAID.

[...]

Issue #1379, opened 2024-12-23
[Feature] Docker image with included emulatorjs cores


gantoine:
Build/release a docker image which contains the emulator JS cores, and let the regular one use the CDN (so we can keep the file size low)

n00b12345:
Eagerly awaiting 😍

gantoine:
Will try to get it out by 3.7.0, but no promises lol https://github.com/rommapp/romm/pull/1406

PR #1391, opened 2024-12-26
feat: Include all EmulatorJS assets (cores, scripts) in docker image

mikhailv:
The browser caching does not work for original emulatorjs CDN, also on new emulatorjs release the emulator integration in old versions of romm become outdated and not working anymore. This forces users who using the emulator in romm [to switch to alpha/rc versions](https://github.com/rommapp/romm/issues/1373) to get it work.

This update embeds all EmulatorJS assets in romm docker image and not depend from CDN anymore.

##### Problems with current approach when emulatorjs assets loaded from CDN:
1. Caching is disabled you may check it for example for this URL: https://cdn.emulatorjs.org/stable/data/cores/pcsx_rearmed-legacy-wasm.data
there are no 'Cache-Control', 'ETag' or `Last-Modified' headers in response, caching is not explicitly enabled or disabled. My browser is always loading ~1MB core to run the game, this can be very annoying if you checking games one by one many times.
2. CDN worked unstable and slow for my internet provider (subjective, and I can't reproduce it now)
3. When EmulatorJS is updating old versions of romm become outdated and emulator does not work in such versions.
From description of EmulatorJS:
> Cloning the repository is no longer recommended for production use. You should use [releases](https://github.com/EmulatorJS/EmulatorJS/releases) or the [CDN](https://cdn.emulatorjs.org/) instead.

As I understand if you want always working version you should load all files from CDN (not loader from npm) or use full assets from release archive.

Also the author of EmulatorJS is saying the same: https://github.com/EmulatorJS/EmulatorJS/issues/930#issuecomment-2561482075

#### Pros of this update:
- faster loading of EmulatorJS assets from the same network where romm is running instead of CDN + caching in browser (nginx allows caching by default for static)
- pin version of EmulatorJS for specific romm release

#### Cons:
- Docker image increased in size +270MB

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,6 +22,15 @@ RUN npm ci
 COPY ./frontend ./
 RUN npm run build
 
+ARG EMULATORJS_VERSION=4.2.0
+
+RUN apk update && apk add 7zip wget
+RUN mkdir -p dist/assets/emulatorjs && \
+    cd dist/assets/emulatorjs && \
+    wget "https://github.com/EmulatorJS/EmulatorJS/releases/download/v${EMULATORJS_VERSION}/${EMULATORJS_VERSION}.7z" && \
+    7z x -y "${EMULATORJS_VERSION}.7z" -o. data && \
+    mv -f data/* . && \
+    rm -rf data "${EMULATORJS_VERSION}.7z"
 
 FROM python:${PYTHON_VERSION}-alpine${ALPINE_VERSION} AS backend-build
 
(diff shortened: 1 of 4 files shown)


adamantike:
A similar improvement has been documented in #1379, but we will publish different Docker images, so users can decide if getting a bigger image makes sense for their use case (as this change represents more than a 50% increase in image size).

Also, EmulatorJS maintainers are aware of the backwards compatibility issue, and the proposed fix in https://github.com/EmulatorJS/EmulatorJS/pull/924 to avoid older RomM and EmulatorJS versions to break.

mikhailv:
@adamantike I think I should close this PR then, right?

gantoine:
@mikhailv I'll close the PR but thank you for submitting it, you've saved us the legwork of having to figure out how to modify the dockerfile, which is very much appreciated!

PR #1396, opened 2024-12-27
fix: MySQL migration and driver support

adamantike:
Trying to run MySQL as database (instead of MariaDB) was failing due to incompatible migrations and lack of support for the MySQL driver.

These small changes fix the issue and recover the support for MySQL as database driver.

Support for MySQL is on a best-effort basis, as the project's main database is MariaDB.


zurdi15 (approved):

zurdi15:
Nice! It was this easy?

adamantike:
We need to keep an eye on possible issue reports, but it seems the main difference with MariaDB is JSON column handling!

PR #1399, opened 2024-12-27
feat: Add CREATE_OIDC_USER configuration to enable automatic user creation

zurdi15:
Added ``CREATE_OIDC_USER`` env variable to enable the automatic creation of a non existing RomM user from an OIDC provider user


backend/config/__init__.py
```diff
@@ -79,6 +79,7 @@ def str_to_bool(value: str) -> bool:
 # OIDC
 OIDC_ENABLED: Final = str_to_bool(os.environ.get("OIDC_ENABLED", "false"))
 OIDC_PROVIDER: Final = os.environ.get("OIDC_PROVIDER", "")
+CREATE_OIDC_USER: Final = str_to_bool(os.environ.get("CREATE_OIDC_USER", "false"))
```
adamantike: Should we use `OIDC_CREATE_USER` to maintain the same `OIDC` namespace for all these environment variables?

zurdi15: totally agree, fixed!

adamantike (approved):

PR #1402, opened 2024-12-27
refactor: Remove OIDC_CREATE_USER configuration and generate an uuid in a OIDC created user password

zurdi15:


dataway:
Should not derive the username from the email claim. It would be better to use the preferred_username claim.

Neither the email not the preferred_username are unique however. See OIDC Core, section 5.7. (https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability) Ideally the claims sub and iss should be stored in the database. This would also allow future support for multiple IdPs.

gantoine:
> Ideally the claims sub and iss should be stored in the database.

This is something we (or another contributer) can tackle in a future release, ad I'm sure there is more functionality we'll need to add in the future around openid/sso.

gantoine (approved):

PR #1404, opened 2024-12-28
OIDC fix auk

dataway:
This patch removes quite a bit of code that is unnecessary because the relevant checks are already done in authlib (specifically in https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/sync_openid.py - the parse_id_token is called when we invoke oauth.openid.authorize_access_token). Therefore it is not necessary for us to do this check again.

This also fixes an issue where the IdP jwks_url returns more than one valid key and it is necessary to use the kid to get the correct key (see line 72 of the aforementioned source).

I also verify that email_verified is set True for security reasons.


gantoine (approved):
looks great, thanks!

PR #1406, opened 2024-12-29
Docker slim images

gantoine:
Aight let's break this down.

The github `build` action now builds 2 images using different metadata tags and arguments. In the `Dockerfile`, the image built with `SLIM_IMAGE=false` will fetch the emulatorjs and ruffle assets and copy them to the frontent `/assets/`. `SLIM_IMAGE` is also passed in at runtime from the `Docekrfile` as an env variable, and is used to determine, when loading the emulatorjs or ruffle players, whether to try fetching the libs/cores locally or from CDN.

I've also refactored the `/heartbeat` response to move all fields into categories, and remove no-longer-needed libs from `package.json`.


backend/endpoints/responses/heartbeat.py
```diff
@@ -1,6 +1,12 @@
 from typing import TypedDict
 
 
+class SytemDict(TypedDict):
```
adamantike: typo: `System`

docker/Dockerfile
```diff
@@ -69,6 +70,35 @@ RUN git clone --recursive --branch "${RALIBRETRO_VERSION}" --depth 1 https://git
     make HAVE_CHD=1 -f ./Makefile.RAHasher
 
 
+FROM alpine:${ALPINE_VERSION} AS emulator-stage
```
adamantike, line 74: I wonder if it's better for this stage to **always** download the assets (not relying on a `SLIM_IMAGE` variable.

However, we would have something like this:

```dockerfile
FROM scratch AS slim-image  # Renamed from final-stage
# ... (existing steps)

FROM slim-image AS full-image
COPY --from=emulator-stage ... # All required copies to convert "slim" into "full"
```

That way, the CI/CD process decides building slim or full based on the `target` parameter instead of an argument. That should also improve Docker layer caching, making the CI reuse what it already built from `slim-image` when building `full-image`.

docker/Dockerfile
```diff
@@ -147,7 +176,10 @@ COPY --from=production-stage / /
 COPY --from=backend-build /src/.venv /src/.venv
 # Fix virtualenv link to python binary
 RUN ln -sf "$(which python)" /src/.venv/bin/python
-ENV PATH="/src/.venv/bin:${PATH}"
+
+ARG SLIM_IMAGE=false
+ENV PATH="/src/.venv/bin:${PATH}" \
+    SLIM_IMAGE="${SLIM_IMAGE}"
```
adamantike: Does it provide any extra value if we determine whether the current image is slim or full, in Python, by checking if `assets/emulatorjs` or `assets/ruffle` are present?

That way, users could still use a Slim image, mount a local copy of Ruffle (for example, if they don't need EmulatorJS), and still be able to enable Ruffle emulation.

.github/workflows/build.yml
```diff
@@ -54,11 +54,41 @@ jobs:
...
+            org.opencontainers.image.title="rommapp/romm"
+            org.opencontainers.image.description="RomM (ROM Manager) allows you to scan, enrich, and browse your game collection with a clean and responsive interface. With support for multiple platforms, various naming schemes and custom tags, RomM is a must-have for anyone who plays on emulators."
+            org.opencontainers.image.licenses="AGPLv3"
+
       - name: Set version
         run: |
           sed -i 's/<version>/${{ steps.meta.outputs.version }}/' backend/__version__.py
 
-      - name: Build base image
+      - name: Build full image
```
adamantike: I think we should build `slim` first and `full` afterwards. With that configured, we can check if the CI is correctly caching previous stages.

.github/workflows/build.yml
```diff
@@ -54,11 +54,41 @@ jobs:
...
+        id: build-full
+        uses: docker/build-push-action@v6
+        with:
+          file: docker/Dockerfile
+          context: .
+          push: true
+          platforms: linux/arm64,linux/amd64
+          tags: ${{ steps.meta-slim.outputs.tags }}
+          labels: ${{ steps.meta-slim.outputs.labels }}
```
adamantike, line 87: `steps.meta-slim` should be swapped with the `steps.meta` from the slim image.

docker/Dockerfile
```diff
@@ -69,6 +71,23 @@ RUN git clone --recursive --branch "${RALIBRETRO_VERSION}" --depth 1 https://git
...
+ARG EMULATORJS_VERSION=4.2.0
+wget "https://github.com/EmulatorJS/EmulatorJS/releases/download/v${EMULATORJS_VERSION}/${EMULATORJS_VERSION}.7z" && \
+7z x -y "${EMULATORJS_VERSION}.7z" -o/emulatorjs && \
+rm -rf "${EMULATORJS_VERSION}.7z"; \
+
+ARG RUFFLE_VERSION=nightly-2024-12-28
+ARG RUFFLE_FILE=ruffle-nightly-2024_12_28-web-selfhosted.zip
+wget "https://github.com/ruffle-rs/ruffle/releases/download/${RUFFLE_VERSION}/${RUFFLE_FILE}" && \
+unzip -o "${RUFFLE_FILE}" -d /ruffle && \
+rm -f "${RUFFLE_FILE}"; \
```
adamantike: Are these steps working? We are missing the `RUN` statement in these lines.

gantoine: ah shit nice catch, they were removed during refactor, will fix

adamantike (approved):

PR #1411, opened 2024-12-30
fix: Make migrations compatible with PostgreSQL

adamantike:
This change only verifies that Alembic can upgrade and downgrade through all the current migrations. It does not verify that the application works correctly with PostgreSQL.

It also allows setting `ROMM_DB_DRIVER` as `postgresql`.


backend/alembic/versions/0024_sibling_roms_db_view.py
```diff
@@ -22,32 +22,35 @@ def upgrade() -> None:
...
-            JOIN 
-                roms r2 
-            ON 
+            JOIN
+                roms r2
+            ON
                 r1.platform_id = r2.platform_id
                 AND r1.id != r2.id
                 AND (
-                    (r1.igdb_id = r2.igdb_id AND r1.igdb_id IS NOT NULL AND r1.igdb_id != '')
```
adamantike, line 46: `r1.igdb_id != ''` comparison is not needed, as the `igdb_id` column is Integer. This doesn't fail on MySQL/MariaDB, but PostgreSQL complain because of `''` not being a possible value for the column.

gantoine (approved):

PR #1412, opened 2024-12-31
feat: continue playing home section

zurdi15:
Added ``Continue playing`` section at home view
Added card skeleton loaders to ``recently added`` and ``continue playing`` sections 
Updated UI options icons


backend/endpoints/rom.py
```diff
@@ -115,21 +117,45 @@ def get_roms(
...
+        roms = db_rom_handler.get_roms_user(
+            user_id=request.user.id,
+            platform_id=platform_id,
+            collection_id=collection_id,
+            search_term=search_term,
+            order_by=order_by,
+            order_dir=order_dir,
+            limit=limit,
+            offset=offset,
+        )
```
gantoine, line 153: is `order_by` a required field? should it be?

gantoine (approved):

backend/endpoints/rom.py
zurdi15: It's not a required field, it is set to ``name`` by default. Same with ``order_dir`` that is set to ``asc`` by default

PR #1413, opened 2024-12-31
feat: PostgreSQL compatibility

### LINKED ISSUE #667: [Feature] Postges backend
No

I would like to host this in my kubernetes cluster that already has a cloudnative postgres instance available.  Can you please add postgres to the list of available backend databases?

N/A

**Additional context**
I don't think this should be a big ask since you're using fastapi, SQLAlchemy, and alembic which all support postgres very well.

adamantike:
This change introduces PostgreSQL compatibility, by implementing the following changes:

* Use `JSONB` instead of `JSON` for PostgreSQL databases. This is achieved by creating a custom `CustomJSON` type that uses `JSONB` on PostgreSQL and `JSON` on other databases, leveraging the variant mechanism of SQLAlchemy.
* Add `is_postgresql` function to check if the current database is PostgreSQL. Commonly used for migrations, but can be used to determine how to build queries based on the database engine.
* Add `json_array_contains_value` function to check if a JSON array includes a specific value. This function is needed as it's engine-specific.

Support for PostgreSQL is on a best-effort basis, and it relies on the community for continued support and testing. The project's main database is MariaDB.

Closes #667.


zurdi15 (approved):

backend/utils/database.py
```diff
@@ -0,0 +1,25 @@
...
+def CustomJSON(**kwargs: Any) -> sa.JSON:
+    """Custom SQLAlchemy JSON type that uses JSONB on PostgreSQL."""
+    return sa.JSON(**kwargs).with_variant(sa_pg.JSONB(**kwargs), "postgresql")
+
+
+def is_postgresql(conn: sa.Connection) -> bool:
+    return conn.engine.name == "postgresql"
+
+
+def json_array_contains_value(
```
gantoine, line 18: surprised sqlalchemy doesn't offer a utility for this

gantoine (approved):

backend/utils/database.py
adamantike: I haven't found a way for MySQL to provide a similar functionality to PostgreSQL's `?` operator, which checks if a value exists in a JSONB array, or as a key for a JSONB object.

Maybe that's why it's not simple to provide this in SQLAlchemy, unless without knowing which kind of value the JSON column will contain:

```sql
SELECT JSON_CONTAINS('{"a": 1, "b": 2, "c": {"d": 4}}', '"a"', '$');  -- FALSE
SELECT JSON_CONTAINS('["a", "b", "c"]', '"a"', '$');  -- TRUE

SELECT JSON_CONTAINS_PATH('{"a": 1, "b": 2, "c": {"d": 4}}', 'one', '$.a');  -- TRUE
SELECT JSON_CONTAINS_PATH('["a", "b", "c"]', 'one', '$.a');  -- FALSE
```

If someone finds a way to do that, I think that could be contributed to SQLAlchemy, as a `Comparator.has_key` function for the `mysql.JSON` dialect type, similar to [the PostgreSQL implementation](https://github.com/sqlalchemy/sqlalchemy/blob/c6be0c1f205cf83467d68eb40fb650b9274519f9/lib/sqlalchemy/dialects/postgresql/json.py#L264-L268).

PR #1415, opened 2025-01-01
Added a korean translation of RomM

ysteen:
Hello. I have added the korean translation of RomM with #1353. Thank you for the great work! I hope the translation is helpful for the project.

PS. since the word order of korean is different with english, I need to add the words at the end of the sentence but the code could not. for example, in the rename-file-part variable, I cannot add the words at the end of the sentence (between the rename-file-part4 and part5). I could apply more natural translation if this is fixed and I hope the dev team keeps the development regarding language support in future releases.


zurdi15:
Thank you for the translation! I get what you mean. Do you prefer to merge this as it is now and at least have a basic Korean translation or do you prefer to wait until we improve the i18n layouts so different structured languages like Korean can be fit better?

zurdi15 (approved):

zurdi15:
I just realized that the commit is not signed. We need them to be signed in order to merge them.

zurdi15:
Never mind, we decided to merge without signed commits :)

ysteen:
Thank you for the merging without signed commit. I didn't know that the sign is needed since it was my first pull request. I will sign the commit later on, and I will revise once again after the i18n layout is updated.

PR #1417, opened 2025-01-04
Default grid view for recently and continue playing to false

gantoine:


frontend/src/components/Home/ContinuePlaying.vue
```diff
@@ -14,7 +14,7 @@ const { continuePlayingRoms } = storeToRefs(romsStore);
 const gridContinuePlayingRoms = isNull(
   localStorage.getItem("settings.gridContinuePlayingRoms"),
 )
-  ? true
+  ? false
   : localStorage.getItem("settings.gridContinuePlayingRoms") === "true";
```
adamantike, line 18: With this new logic everywhere, is this now equivalent to...?

```ts
const gridContinuePlayingRoms = localStorage.getItem("settings.gridContinuePlayingRoms") === "true";
```

Same for the other changed statements.

gantoine: yeah i think so, though there's no hard to keeping it like this for code consistency

adamantike (approved):

PR #1429, opened 2025-01-07
[ROMM-1414] Env var to disable user/pass login button

gantoine:
Screenshot 2025-01-07 at 5 29 30 PM


zurdi15 (changes requested):
I would conditional render the ``v-form`` instead of the ``v-btn`` since the form is unused when standard login is disabled

zurdi15 (approved):

PR #1434, opened 2025-01-07
Fix lockfile to add integrity

jvanbruegge:
There is a npm bug that sometimes omits those fields. For reproducibility this is needed. I added a CI check in the second commit that will fail if there are missing fields.
The versions are not changed (the lockfile was not regenerated), just fixed with  https://github.com/jeslie0/npm-lockfile-fix

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -34,3 +34,8 @@ jobs:
       - name: Run typecheck
         run: npm run typecheck
         working-directory: frontend
+
+      - name: Lockfile lint
+        run: |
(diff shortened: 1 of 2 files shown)


gantoine:
Thanks for the PR! Not that I don't trust you, but we don't accept changes to `package-lock.json` (or `poetry.lock`) that we can't manually review/validate, so I've just gone ahead and regenerated the lock file, with integrity checks. I don't expect it'll break anything as all of our dependencies are pinned.

gantoine (approved):

jvanbruegge:
Sure, that works too. Thanks!

PR #1455, opened 2025-01-10
[ROMM-1433] Rework columns in table view

### LINKED ISSUE #1453: [Feature] Allow sorting games by "Release Date" when viewing platforms, collections, etc.
As per the title description,.. 

Since RomM is already pulling release date metadata for each game, (imo) allowing us to sort by release date (as we can with title, size, reg, lang, & rev currently) would be a great addition when viewing platforms, collections, etc.

image

gantoine:
Screenshot 2025-01-09 at 10 38 14 PM

Fixed #1453

### DIFF
--- a/backend/endpoints/responses/rom.py
+++ b/backend/endpoints/responses/rom.py
@@ -121,7 +121,7 @@ class RomSchema(BaseModel):
     # Metadata fields
     first_release_date: int | None
     youtube_video_id: str | None
-    average_rating: float
+    average_rating: float | None
     alternative_names: list[str]
     genres: list[str]
     franchises: list[str]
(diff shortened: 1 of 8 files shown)


backend/endpoints/responses/rom.py
```diff
@@ -121,6 +121,7 @@ class RomSchema(BaseModel):
     # Metadata fields
     first_release_date: int | None
     youtube_video_id: str | None
+    average_rating: float
```
adamantike: I would ideally mark this field as nullable. In case a game doesn't match any metadata providers, it would be better for the frontend to show an empty rating rather than zero, which could be confusing for the user.

PR #1461, opened 2025-01-10
[ROMM-1427] Add option to remove game from continue playing

gantoine:


frontend/src/components/common/Game/AdminMenu.vue
```diff
@@ -86,6 +87,31 @@ async function switchFromFavourites() {
...
+
+async function resetLastPlayed() {
+  await romApi
+    .updateUserRomProps({
+      romId: props.rom.id,
+      data: { last_played: null },
+    })
+    .then(() => {
+      emitter?.emit("snackbarShow", {
+        msg: `${props.rom.name} removed from Continue Playing`,
```
adamantike, line 99: Do we currently support i18n here?

backend/endpoints/rom.py
```diff
@@ -558,6 +558,7 @@ async def delete_roms(
 @protected_route(router.put, "/roms/{id}/props", [Scope.ROMS_USER_WRITE])
 async def update_rom_user(request: Request, id: int) -> RomUserSchema:
     data = await request.json()
+    data = data["data"] or {}
```
zurdi15:
```suggestion
    data = data.get("data", {})
```

zurdi15 (changes requested):

frontend/src/components/common/Game/AdminMenu.vue
gantoine: Yes, but we already have a ton of these error messages that aren't translated. At some point we'll do a second pass and correct these, but since they're transient IMO they're lower priority.

zurdi15 (approved):

PR #1474, opened 2025-01-12
fix: SocketIO JSON dumps with sane defaults

adamantike:
Similar to how `engineio` provides a JSON-compatible module, this change adds a custom JSON encoder with support for additional types.

Changing SocketIO's JSON module fixes the encoding issue when the scanning process tried to send a datetime, failing and the frontend not displaying the scanned game (commonly, when it had sibling games)


gantoine (approved):
ayo nice find!

PR #1476, opened 2025-01-12
misc: Upgrade Poetry to v2.0

adamantike:
Release notes:
https://python-poetry.org/blog/announcing-poetry-2.0.0/


pyproject.toml
```diff
@@ -1,59 +1,63 @@
...
 license = "GNU AGPLv3"
-repository = "https://github.com/rommapp/romm"
-authors = ["Zurdi <zurdi@romm.app>", "Arcane <arcane@romm.app>"]
+readme = "README.md"
+authors = [
+  { name = "Zurdi", email = "zurdi@romm.app" },
+  { name = "Arcane", email = "arcane@romm.app" },
+]
+requires-python = "^3.12"
+dependencies = [
```
gantoine, line 12: should not go under `[project.dependencies]`?

gantoine (changes requested):

pyproject.toml
```diff
@@ -1,59 +1,63 @@
...
-version = "0.0.1"
 description = "A beautiful, powerful, self-hosted rom manager"
 license = "GNU AGPLv3"
-repository = "https://github.com/rommapp/romm"
-authors = ["Zurdi <zurdi@romm.app>", "Arcane <arcane@romm.app>"]
+readme = "README.md"
+authors = [
+  { name = "Zurdi", email = "zurdi@romm.app" },
+  { name = "Arcane", email = "arcane@romm.app" },
+]
```
gantoine, line 10: missing your name and email here lol

pyproject.toml
adamantike: Yes, but as it's just an array of strings, that's the way to represent it in TOML.
Example: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies

pyproject.toml
adamantike: You're right! I finally added myself haha

gantoine (approved):

PR #1483, opened 2025-01-13
fix: Poetry dependency installation during pytest job

adamantike:
With Poetry 2.0, we now need to explicitly tell it to install the `test` extra.


gantoine (approved):
💯

PR #1490, opened 2025-01-14
fix: Set users email to null if empty

### LINKED ISSUE #1445: [Bug] Can't create second user without mail adress
**RomM version**
3.7.0

**Describe the bug**
If I try to add a second user without adding a mail I am getting a message "Unable to create user: Uesr with email already exists."

Side note: Check the spelling

**To Reproduce**

1. Create any user without an email
2. Create second user without an email

**Expected behavior**
It should create the user without issues

**Screenshots**
Screenshot 2025-01-08 150551

adamantike:
When a user does not set an email address, we now set it to `NULL` in the database. That bypasses the unique constraint on the `email` column, allowing multiple users to have no email address set.

Fixes #1445.

### DIFF
--- a/backend/alembic/versions/0030_user_email_null.py
+++ b/backend/alembic/versions/0030_user_email_null.py
@@ -1,0 +1,25 @@
+"""Change empty string in users.email to NULL.
+
+Revision ID: 951473b0c581
+Revises: 0029_platforms_custom_name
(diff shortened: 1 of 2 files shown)


gantoine (approved):
yaaaass very nice

PR #1492, opened 2025-01-14
Case-insensitive username and email check

### LINKED ISSUE #1486: [Bug] Unable to login after fresh install
**RomM version**
3.7.0

**Describe the bug**
I am unable to login after going through the setup. Getting 401 Incorrect username or password. Have tried three times with clean install removing containers and volumes. Tried different username/password combinations. Never filled out email. Using postgres database.

- OS: Ubuntu 22.04
- Browser: Firefox 133.0.3

adamantike:
Move case-insensitivity to the database layer, for both `username` and `email` fields.

Fixes #1486.


backend/handler/database/users_handler.py
```diff
@@ -15,11 +15,13 @@ def add_user(self, user: User, session: Session = None) -> User:
     def get_user_by_username(
         self, username: str, session: Session = None
     ) -> User | None:
-        return session.scalar(select(User).filter_by(username=username).limit(1))
+        return session.scalar(
+            select(User).filter_by(username=username.lower()).limit(1)
```
gantoine: would this need to be something like `filter(func.lower(User.username) == username.lower())`?

gantoine (approved):

adamantike:
The missing change I would add here, is replacing existing unique constraints on email and username, for case-insensitive unique constraints.

PR #1504, opened 2025-01-18
fix: Improve support for older PostgreSQL versions

adamantike:
Functions like `JSON_ARRAY` and `JSON_OBJECT` have been introduced in PostgreSQL 16 [1]. Older PostgreSQL versions were failing to run the migrations due to the usage of these functions.

This change uses `jsonb_build_array` and `jsonb_build_object` functions in PostgreSQL, which have been available since PostgreSQL 9.5 [2].

[1] https://www.postgresql.org/docs/release/16.0/
[2] https://www.postgresql.org/docs/release/9.5.0/#AEN132010


gantoine (approved):
💯

PR #1511, opened 2025-01-20
misc: Do not fail if LOGLEVEL is not uppercase

adamantike:
The `logging.setLevel` method expects the log level to be in uppercase. This change ensures that the `LOGLEVEL` config value is always in uppercase.


gantoine (approved):
mhmm nice

PR #1514, opened 2025-01-20
[ROMM-1503] Remember search query when going back to page

### LINKED ISSUE #1503: [UI/UX] Search is lost when you go back to the search results
I noticed something that I find mildly inconvenient. If you use the search bar, get the results, and then click on a game, once you move back (by clicking the browser button), the search is empty. I'd be convenient if the search results were preserved. Not a bit deal, though.

gantoine:
Store the query in the URL as a query param, and use it on search page load to prepopulate the results

Fixes #1503


frontend/src/views/Gallery/Platform.vue
```diff
@@ -56,7 +54,6 @@ async function fetchRoms() {
   try {
     const { data } = await romApi.getRoms({
       platformId: romsStore.currentPlatform?.id,
-      searchTerm: normalizeString(galleryFilterStore.filterText),
```
gantoine, line 59: these haven't been required for a while since filtering happens client-side

zurdi15:
Shouldn't it better to just store the roms fetched by the search in a store like we do with the platforms and collections so the results will be there until another search happens?
Also, the searched term can be stored in a pinia store and just show it each time the user visits the search view.

Apart from that, I should keep the url query params since that will allow to share or bookmark searches as URLs

gantoine:
> Shouldn't it better to just store the roms fetched by the search in a store like we do with the platforms and collections so the results will be there until another search happens?

Yes that would be good, but more work then what I've done here IMO

> Also, the searched term can be stored in a pinia store and just show it each time the user visits the search view.

That I'm not sure about, having two sources of truth (pinia store + url query param) would cause issues. i was just dealing with something similar at work a couple weeks ago, but with x-state and query params lol

PR #1515, opened 2025-01-20
[ROMM-403] Import metadata from Launchbox database

gantoine:
This PR adds Lanchboxe's [GamesDB](https://gamesdb.launchbox-app.com) as a metadata source. The entire `Metadata.xml` is pulled nightly, parsed and stored in Valkey. It matches games based on _exact_ file names (excluding tags and extensions).


adamantike (approved):

backend/alembic/versions/0041_launchbox_id.py
```diff
@@ -0,0 +1,55 @@
...
+
+revision = "0041_launchbox_id"
+down_revision = "0040_migrate_assets_paths"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("rom_user", schema=None) as batch_op:
+        batch_op.drop_column("ra_metadata")
```
gantoine: @zurdi15 seems this column is gone from the model, can you confirm it should be dropped?

zurdi15: I added it because at the beginning I was going to add the ra metadata per user-rom, but at the end I added the whole metadata to each user row, those are kind of leftovers.

We can drop it

backend/alembic/versions/0036_screenscraper_platforms_id.py
```diff
@@ -8,7 +8,7 @@
 
 import sqlalchemy as sa
 from alembic import op
-from handler.metadata.ss_handler import SLUG_TO_SS_ID
+from handler.metadata.ss_handler import SCREENSAVER_PLATFORM_LIST
```
zurdi15, line 11: SCREENSAVER everywhere 😆

backend/endpoints/platform.py
```diff
@@ -90,6 +90,8 @@ def get_supported_platforms(request: Request) -> list[PlatformSchema]:
             "created_at": now,
             "updated_at": now,
             "fs_size_bytes": 0,
+            "is_unidentified": False,
+            "is_identified": True,
```
zurdi15, line 94: why two flags? flag for partially identified?

backend/models/platform.py
```diff
@@ -53,6 +54,28 @@ class Platform(BaseModel):
...
+    def is_unidentified(self) -> bool:
+        return (
+            not self.igdb_id
+            and not self.moby_id
+            and not self.ss_id
+            and not self.launchbox_id
+        )
+
+    @property
+    def is_identified(self) -> bool:
```
zurdi15, line 66: why not negating ``is_unidentified`` directly?

backend/endpoints/platform.py
gantoine: `unidentified` means no metadata IDs, `is_identified` means some metadata IDs, `fully_identified` (which is not useful rn) means all metadata IDs

backend/models/platform.py
gantoine: is that not what's happening here?

zurdi15: Yes but I mean, negating the function call/property instead of having a separate function /property

gantoine: eh i suppose, though i think this is ok for now for consistency

PR #1523, opened 2025-01-21
Fix MySQL json_contains Error by Converting Non-String Values to JSON Strings

jamiecounsell:
Loving the project so far 😎! Thanks for adding MySQL support!

I am running into an issue using MySQL. When viewing a ROM, it yields the error:

```
(mysql.connector.errors.DataError) 3146 (22032): Invalid data type for JSON data in argument 1 to function json_contains; a JSON string or JSON type is required.
```

This is beacuse for mySQL, the driver expects a string here, and receives an int. This PR fixes it by calling json.dumps(value) before passing the value to json_contains. This doesn’t affect Postgres, which uses different logic (has_key/contains).

I have tested and verified this running Romm 3.7.2 with mySQL ~8.0~ 9.1.0 on Unraid 🚀

### DIFF
--- a/backend/utils/database.py
+++ b/backend/utils/database.py
@@ -16,6 +16,10 @@ def is_postgresql(conn: sa.Connection) -> bool:
     return conn.engine.name == "postgresql"
 
 
+def is_mysql(conn: sa.Connection) -> bool:
+    return conn.engine.name == "mysql"
+
+
 def json_array_contains_value(
     column: sa.Column, value: Any, *, session: Session
 ) -> ColumnElement:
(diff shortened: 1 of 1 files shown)


adamantike:
Thanks for the contribution, @jamiecounsell!

As you mentioned, this logic won't affect PostgreSQL, as it only runs for MariaDB/MySQL environments. MariaDB supports receiving an integer as second argument for `JSON_CONTAINS`, so I think we could have an implementation where casting the object is only needed for MySQL, to reduce the amount of work we do in Python, and rely on each engine's JSON support.

A few ideas after testing this change both in MariaDB and MySQL:

* The simplest option would be to add an `is_mysql` util function to this file, and have another branch here to run `JSON_CONTAINS` plus `json.dumps` only for MySQL. That way, MariaDB doesn't get the extra value cast.
* The native solution for MySQL would be to rely on the [`MEMBER OF` operator](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#operator_member-of) introduced in [MySQL 8.0.17](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-17.html#mysqld-8-0-17-feature). We would also need the `is_mysql` util as the operator doesn't exist in MariaDB, but we would avoid casting values using `json.dumps` at all.

I don't think SQLAlchemy provides the `MEMBER OF` operator, so we would need to build it ourselves.

jamiecounsell:
Updated!

adamantike (approved):
Looks great, thanks for your contribution!

PR #1561, opened 2025-01-27
fix: Remove sibling roms from Recently Added.

ItsKaa:
Small UI change.

The Recently Added element displays every sibling whereas the Platform pages do not, this streamlines that by making the roms under Recently Added behave similarly, it will only display the main sibling (or the first entry) per metadata id. Continue Playing will still show the individual siblings.

A cleaner solution would obviously be a dropdown in the UI to select the specific version (perhaps more so for Continue Playing) but I don't feel like going that far right now. Feel free to decline if this is intended behavior though.

Before:
Screenshot_20250127_205658
The displayed ids in order: `17, 16, 15, 13, 12, 10, 9`

After:
Screenshot_20250127_204937
The displayed ids in order: `17, 16, 15, 13`

### DIFF
--- a/frontend/src/stores/roms.ts
+++ b/frontend/src/stores/roms.ts
@@ -41,6 +41,23 @@ export default defineStore("roms", {
   },
 
   actions: {
+    _getGroupedRoms(roms: SimpleRom[]): SimpleRom[] {
+      // Group roms by external id.
+      return Object.values(
+        groupBy(
+          roms,
+          (game) =>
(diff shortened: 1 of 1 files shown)


adamantike:
Should this behavior only be applied when the "Group roms" configuration is set?

ItsKaa:
> Should this behavior only be applied when the "Group roms" configuration is set?

Whoops, yeah probably. Just updated that.

gantoine (approved):

PR #1562, opened 2025-01-29
[ROMM-994] Autogenerated collections

### LINKED ISSUE #994: [Feature] Auto-generated collections for franchises
It would be great to have collections that are automatically generated (possibly with this feature able to be toggled on or off for those who don't want it) based on franchise metadata. This way, we can have franchise games be even more easily discoverable within Romm.

gantoine:
### Description

Very excited about this time! This PR creates a virtual database table that collects and groups games based on certain IGDB metadata fields: collections, franchises, genres, companies and modes. The list of virtual `VirtualCollections` can be fetched and filtered by type, and displayed in the home screen and collections sidebar. Editing these virtual collections is not possible, so controls like the collection editor window are hidden.

As we add more metadata providers, we can update the virtual table to pull properties outside `igdb_metadata`.

### TODO

- [ ] Test migration with postgresql

## Related Issues

Closes #994 

## Checklist

Please check all that apply.

- [ ] All existing tests are passing

### Screenshots

| | |
|---|---|
|Screenshot 2025-01-28 at 7 00 51 PMScreenshot 2025-01-28 at 7 04 04 PM|Screenshot 2025-01-28 at 9 52 22 PM|


backend/alembic/versions/0034_virtual_collections_db_view.py
```diff
@@ -0,0 +1,306 @@
...
+                    SELECT * FROM franchises_collection
+                    UNION ALL
+                    SELECT * FROM collection_collection
+                    UNION ALL
+                    SELECT * FROM modes_collection
+                    UNION ALL
+                    SELECT * FROM companies_collection
+                ) combined
+                GROUP BY collection_type, collection_name
+                HAVING COUNT(DISTINCT rom_id) > 2
```
adamantike, line 133: Should this be `> 1` or `>= 2` instead?

backend/alembic/versions/0034_virtual_collections_db_view.py
```diff
@@ -0,0 +1,306 @@
...
+                    SELECT * FROM franchises
+                    UNION ALL
+                    SELECT * FROM collections
+                    UNION ALL
+                    SELECT * FROM modes
+                    UNION ALL
+                    SELECT * FROM companies
+                ) combined
+                GROUP BY collection_type, collection_name
+                HAVING COUNT(DISTINCT rom_id) > 2
```
adamantike, line 255: Same here.

backend/endpoints/collections.py
```diff
@@ -108,9 +110,30 @@ def get_collections(request: Request) -> list[CollectionSchema]:
     """
 
     collections = db_collection_handler.get_collections()
+
     return CollectionSchema.for_user(request.user.id, [c for c in collections])
 
 
+@protected_route(router.get, "/collections/virtual", [Scope.COLLECTIONS_READ])
```
adamantike: nit: Do not forget to rebase and remove the `/collections` prefix to the route.

backend/endpoints/collections.py
```diff
@@ -124,13 +147,31 @@ def get_collection(request: Request, id: int) -> CollectionSchema:
...
 
     collection = db_collection_handler.get_collection(id)
-
     if not collection:
         raise CollectionNotFoundInDatabaseException(id)
 
     return CollectionSchema.model_validate(collection)
 
 
+@protected_route(router.get, "/collections/virtual/{id}", [Scope.COLLECTIONS_READ])
```
adamantike: Same here

backend/alembic/versions/0034_virtual_collections_db_view.py
```diff
@@ -0,0 +1,306 @@
...
+                SELECT
+                    collection_name as name,
+                    collection_type as type,
+                    CONCAT('Autogenerated ', collection_name, ' collection') AS description,
+                    NOW() AS created_at,
+                    NOW() AS updated_at,
+                    JSON_ARRAYAGG(DISTINCT rom_id) as rom_ids,
+                    JSON_ARRAYAGG(DISTINCT path_cover_s) as path_covers_s,
+                    JSON_ARRAYAGG(DISTINCT path_cover_l) as path_covers_l
+                FROM
```
adamantike, line 242: Should these three fields always match their orderings? e.g. Second rom_id corresponds with second element for the cover paths in the other arrays.

If that's the case, I think we should enforce sorting using something like:
```sql
                    JSON_ARRAYAGG(DISTINCT rom_id ORDER BY rom_id) as rom_ids,
                    JSON_ARRAYAGG(DISTINCT path_cover_s ORDER BY rom_id) as path_covers_s,
                    JSON_ARRAYAGG(DISTINCT path_cover_l ORDER BY rom_id) as path_covers_l
```

backend/alembic/versions/0034_virtual_collections_db_view.py
```diff
@@ -0,0 +1,306 @@
...
+                )
+                SELECT
+                    collection_name as name,
+                    collection_type as type,
+                    'Autogenerated ' || collection_type || ' collection' AS description,
+                    NOW() AS created_at,
+                    NOW() AS updated_at,
+                    array_to_json(array_agg(rom_id)) as rom_ids,
+                    array_to_json(array_agg(path_cover_s)) as path_covers_s,
+                    array_to_json(array_agg(path_cover_l)) as path_covers_l
```
adamantike, line 120: This differs from MySQL, where we use `DISTINCT` within the aggregation. I don't think that's needed as I don't see a way for CTEs to repeat `rom_id` values. We should have the same logic for both engines.

backend/alembic/versions/0034_virtual_collections_db_view.py
gantoine: imo 2 games don't make a collection, so i set the minimum to 3 games

backend/alembic/versions/0034_virtual_collections_db_view.py
gantoine: same answer!

backend/alembic/versions/0034_virtual_collections_db_view.py
gantoine: the magic here is that it'll always be in order, even without `ORDER BY`!
the duplicates here are a postgres thing, but don't affect anything
Screenshot 2025-02-09 at 8 36 17 PM

backend/alembic/versions/0034_virtual_collections_db_view.py
gantoine: i can't seem to figure out how to `DISTINCT` in postgres like i did in mysql 🤷🏼

PR #1581, opened 2025-02-05
PS5 icon

Casuallynoted:
Adds ps5.ico for the PS5 platform. Updated attributions to provide proper attribution for the new icon.


gantoine (approved):
niiiice thanks!

PR #1586, opened 2025-02-06
Read-only kiosk mode for viewers

gantoine:
## Description

This PR add a new env var `KIOSK_MODE`, which when set to `true` will disable all editing for `VIEWER` roles. Options in the UI will be disabled/hidden, and the scopes on the backend are restricted to read-only ones. Login and setup screens will also be bypassed when accessing any authenticated page directly.

## Checklist

Please check all that apply.

### DIFF
--- a/frontend/src/components/common/Navigation/SettingsDrawer.vue
+++ b/frontend/src/components/common/Navigation/SettingsDrawer.vue
@@ -88,7 +88,7 @@ async function logout() {
       <v-list-item
         class="mt-1"
         rounded
-        :to="{ name: 'userInterface' }"
+        :to="{ name: 'user-interface' }"
         append-icon="mdi-palette"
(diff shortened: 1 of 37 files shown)


frontend/src/components/common/Navigation/SettingsDrawer.vue
```diff
@@ -96,7 +97,7 @@ async function logout() {
         class="mt-1"
         rounded
         append-icon="mdi-table-cog"
-        :to="{ name: 'libraryManagement' }"
+        :to="{ name: 'library-management' }"
```
zurdi15: Maybe we should use in this kind of links the ``ROUTES`` variable we have in ``router.ts``

zurdi15 (changes requested):

zurdi15 (approved):

PR #1609, opened 2025-02-14
fix: Migration 0034 compatibility with MySQL

adamantike:
## Description

`DISTINCT` cannot be used within `JSON_ARRAYAGG` in MySQL, so it was removed.

Also, the PostgreSQL query was simplified and now generates JSONB columns, to be consistent with all the other tables.

## Checklist

Please check all that apply.

- [ ] I've updated the wiki accordingly


gantoine (approved):
nice, we didnt need the distinct anyways

PR #1611, opened 2025-02-14
fix: Custom covers for collections

zurdi15:
## Description

Add again the ability to set custom covers for collections that were removed in #1562.
Also added the platform icon in the list view of a collection

## Checklist

Please check all that apply.

## Screenshots

image

### DIFF
--- a/frontend/src/stores/collections.ts
+++ b/frontend/src/stores/collections.ts
@@ -73,7 +73,13 @@ export default defineStore("collections", {
     isVirtualCollection(
       collection: Collection | VirtualCollection,
     ): collection is VirtualCollection {
-      return (collection as VirtualCollection).id !== undefined;
+      return (
+        collection &&
+        typeof collection === "object" &&
+        "id" in collection &&
+        "name" in collection &&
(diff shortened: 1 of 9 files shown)


frontend/src/stores/collections.ts
```diff
@@ -73,7 +73,13 @@ export default defineStore("collections", {
...
       collection: Collection | VirtualCollection,
     ): collection is VirtualCollection {
-      return (collection as VirtualCollection).id !== undefined;
+      return (
+        collection &&
+        typeof collection === "object" &&
+        "id" in collection &&
+        "name" in collection &&
+        "virtual" in collection
+      );
```
gantoine, line 80: i don't think this is totally accurate, but merge the PR and ill improve the response in another one

gantoine (approved):
lgtm after formatting is fixed

frontend/src/stores/collections.ts
zurdi15: Yeah you are probably right, I'm not so confident about it but it's the only way I made it work. You will for sure improve it

PR #1617, opened 2025-02-16
[ROMM-1616] Support custom icons for platform versions

gantoine:
# Pull request template

## Description

When using PLATFORM_VERSIONS, this PR sets the `platform.slug` to the "real" metadata slug, instead of the `fs_slug`. It also updates the `platform-icon` component to try and load the `fs_slug` first, then fallback to the `slug`, then finally the default icon. All of this allows us to display the custom `.ico` for bindings and versions, while allowing users to play the system with emulatorjs. Also fixes the logout button visibility on mobile.

## Screenshots

Screenshot 2025-02-16 at 11 53 20 AM

### DIFF
--- a/frontend/src/components/common/Platform/Icon.vue
+++ b/frontend/src/components/common/Platform/Icon.vue
@@ -1,28 +1,22 @@
 <script setup lang="ts">
-import storeConfig from "@/stores/config";
-import { storeToRefs } from "pinia";
-
 const props = withDefaults(
   defineProps<{
     slug: string;
     name?: string;
     size?: number;
     rounded?: number;
+    fsSlug?: string;
   }>(),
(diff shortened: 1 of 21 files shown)


frontend/src/components/common/Platform/Icon.vue
```diff
@@ -1,28 +1,22 @@
...
     size?: number;
     rounded?: number;
+    fsSlug?: string;
   }>(),
   { size: 40, rounded: 0 },
 );
-const configStore = storeConfig();
-const { config } = storeToRefs(configStore);
+
+console.log(props);
```
zurdi15: i guess this is a leftover?

zurdi15 (approved):

PR #1645, opened 2025-02-21
[HOTFIX] Fix collections migration in both directions

gantoine:
Also updates trunk plugins


backend/alembic/versions/0034_virtual_collections_db_view.py
```diff
@@ -52,6 +52,8 @@ def upgrade() -> None:
                 SELECT c.id, rom_id::INT, NOW(), NOW()
                 FROM collections c,
                 LATERAL jsonb_array_elements_text(c.roms) AS rom_id
+                LEFT JOIN roms r ON rom_id::INT = r.id
+                WHERE r.id IS NOT NULL;
```
adamantike, line 56: nit: For simplicity on this query and the next one, we could just use an inner join, and avoid the `WHERE`.

adamantike (approved):

PR #1650, opened 2025-02-22
misc: Strip environment variables related to third-party credentials

adamantike:
## Description

For client_id/client_secret combinations, and API keys, we are certain that they should not have leading/trailing whitespaces. We can strip them to avoid any potential issues for users who might have added them incorrectly in their Docker Compose configuration.

This change does NOT affect users/passwords, where leading/trailing whitespaces might be intentional and valid.

## Checklist

Please check all that apply.

- [ ] I've updated the wiki accordingly
- [ ] I've have updated relevant comments


gantoine (approved):
💯

PR #1732, opened 2025-03-15
True endpoint pagination

gantoine:
#### Description

#### Checklist


backend/handler/database/roms_handler.py
```diff
@@ -129,11 +129,9 @@ def get_roms(
         search_term: str | None = None,
         order_by: str = "name",
         order_dir: str = "asc",
-        limit: int | None = None,
-        offset: int | None = None,
         query: Query = None,
         session: Session = None,
-    ) -> Sequence[Rom]:
+    ) -> Query[Rom]:
```
adamantike, line 405: Considering this method doesn't run the query anymore, we can remove the `@begin_session` decorator and the `session` parameter, to reduce unneeded DB overhead. Same for `get_roms_user`.

For that to work, I think both `with_simple` and `with_details` must stop checking that `session` is provided.

backend/endpoints/rom.py
```diff
@@ -122,12 +125,12 @@ def get_roms(
     collection_id: int | None = None,
     virtual_collection_id: str | None = None,
     search_term: str | None = None,
-    limit: int | None = None,
-    offset: int | None = None,
+    limit: int = 72,
```
adamantike: According to the [docs](https://uriyyo-fastapi-pagination.netlify.app/learn/tutorial_user_guide/add_to_route/), we shouldn't need to specify `limit` and `offset` here, and can avoid manually instantiating `LimitOffsetParams`.

We can test if that's the case, and whether those params are still added to the OpenAPI spec.

backend/handler/database/roms_handler.py
gantoine: >  we can remove the @begin_session decorator and the session parameter

Unfortunately not since `_filter` needs a session to run subqueries

```
def _filter(
        self,
        data,
        platform_id: int | None,
        collection_id: int | None,
        virtual_collection_id: str | None,
        search_term: str | None,
        session: Session,
    ):
        if platform_id:
            data = data.filter(Rom.platform_id == platform_id)

        if collection_id:
            collection = (
                session.query(Collection)
                .filter(Collection.id == collection_id)
                .one_or_none()
            )
```

backend/endpoints/rom.py
gantoine: yep works!

PR #1754, opened 2025-03-18
Tinfoil support with endpoint for file download

gantoine:
#### Description

This PR adds 2 new endpoints to `/api/roms` that allow you to get a `RomFile` by ID and download that file by passing the ID (and a filename). This is required since Tinfoil doesn't support passing query params and uses the last part of the URL as the filename on download. Check the screenshot below for example response content from `/api/tinfoil/feed`.

cc @KaitoKid

#### Checklist

#### Screenshots

Screenshot 2025-03-18 at 9 50 09 AM


gantoine:
> I've also identified an edge case where adding/modifying files or paths within the folder of a rom doesn't get reflected in a quick scans - which likely affects all roms, not just switch/tinfoil.

This is expected behavior, and something we can address outside this PR. It if seems to work well as-is we should merge it in and build on top of it. Thoughts @KaitoKid?

KaitoKid:
Aligned - the PR looks good to me. Thanks @gantoine

gantoine:
Awesome, very excited to announce this is fixed in the upcoming release notes 🚀

PR #1834, opened 2025-04-13
Add support for running DOS "roms" using DosWasmX

ICTylor:
**Description**
Add support for running DOS "roms" using DosWasmX. Adds the play button to DOS platform roms and passes the download link from the rom to DosWasmX (calling the internal function that would be called in case of drag and drop).

Uses the original index.html from the dist folder of the https://github.com/ICTylor/DosWasmX project, which is a fork of https://github.com/nbarkhina/DosWasmX without any changes at the moment (except for having the dist folder as a release). The index.html file is loaded in an iframe.

This is just a quick hack that I have only tested minimally locally. I expect that adding DOS emulation support would probably require a very different implementation to be accepted, but I wanted to share it regardless in case it is somehow useful or serves as an inspiration.

**Checklist**


gantoine:
Hey neat project, thanks for the PR! Unfortunately I don't feel comfortable including code from untested and un-etted project with 0 stars and 2 contributors directly in our docker image. I understand that it's a fork of `nbarkhina/DosWasmX`, and if we were to move forward with this, we'd want to use the base project as the source of truth, or host our own fork that the team can vet internally.

ICTylor:
> Hey neat project, thanks for the PR! Unfortunately I don't feel comfortable including code from an untested and un-vetted repo with 0 stars and 2 contributors, directly in our docker image. I understand that it's a fork of `nbarkhina/DosWasmX`, and if we were to move forward with this, we'd want to use the base project as the source of truth, or host our own fork that the team can vet internally.

Of course, I didn't expect otherwise. It was just the first repo that I found that was more or less easy to integrate quickly. I guess https://github.com/dreamlayers/em-dosbox would be a much better option, given that it is used by Archive.org apparently. I am not sure how easy it would be to build it (and the last commit seems to be from 3 years ago, not sure if it is still maintained)

There is also this one https://github.com/caiiiycuk/js-dos maybe.

If there is interest on a similar pull request with any of those (or any other suggested repo), I could give it a try (although it might be a bit more complex and I am not sure if I will have time)

PR #1885, opened 2025-05-13
feat: Recover password

zurdi15:
**Description**
This PR adds a option in the login view to recover a forgotten password. Any user can now request a link to recover and set a new password. The link will be printed in the container logs, so any admin can provide the link to the user that requested the recovery.

**Checklist**


backend/endpoints/auth.py
```diff
@@ -270,3 +273,49 @@ async def auth_openid(request: Request):
...
+@router.post("/forgot-password")
+def request_password_reset(username: str = Body(..., embed=True)) -> MessageResponse:
+    """Request password reset by email.
+
+    Args:
+        email (str): User's email
+
+    Returns:
+        MessageResponse: Confirmation message
+    """
```
gantoine: this comment text isn't right

backend/handler/auth/base_handler.py
```diff
@@ -63,6 +65,93 @@ async def get_current_active_user_from_session(self, conn: HTTPConnection):
...
+        Set the new password for the user.
+        Args:
+            user (Any): The user object.
+            new_password (str): The new password to set.
+        """
+        from handler.database import db_user_handler
+
+        db_user_handler.update_user(
+            user.id, {"hashed_password": self.get_password_hash(new_password)}
+        )
```
gantoine, line 153: 👍🏼

frontend/src/plugins/router.ts
```diff
@@ -238,7 +259,7 @@ router.beforeEach(async (to, _from, next) => {
 
 router.afterEach(() => {
   // Scroll to top to avoid annoying behaviour on mobile
-  window.scrollTo({ top: 0, left: 0 });
+  // window.scrollTo({ top: 0, left: 0 });
```
gantoine: temp or permanent removal?

gantoine (changes requested):

backend/endpoints/auth.py
zurdi15: true! I was going to do it by email but reverted that deccision. Fixed!

frontend/src/plugins/router.ts
zurdi15: permanent! I forgot to add the logic to scroll to top on new page but keep the previous scroll if going back

gantoine (approved):

PR #1890, opened 2025-05-14
refactor: Assets filesystem path

zurdi15:
**Description**
This PR makes a change in the assets path where now we use the ``rom id`` in the path where the ``saves``, ``states`` and ``screenshots`` are stored:

 - Before: ``assets/users/{user_id}/(saves|states|screenshots)/{platform_slug}/[{emulator}]/{save_filename}``
 - After: ``assets/users/{user_id}/(saves|states|screenshots)/{platform_slug}/{rom_id}/[{emulator}]/{save_filename}``

This fix the overwriting issue where an asset was replaced by the new uploaded one if both had the same name, for different games of the same platform.

**Checklist**


frontend/src/views/Player/EmulatorJS/Base.vue
```diff
@@ -50,21 +50,34 @@ function onPlay() {
...
+      script.textContent = jsCode;
+      document.body.appendChild(script);
+    })
+    .catch(() => {
+      console.warn("Local EmulatorJS failed, falling back to CDN");
+      window.EJS_pathtodata = CDN_PATH;
+      const fallbackScript = document.createElement("script");
+      fallbackScript.src = `${CDN_PATH}loader.js`;
+      document.body.appendChild(fallbackScript);
+    });
```
gantoine, line 81: not sure what's happening here, will have to take a closer look before this is merged

gantoine (approved):

frontend/src/views/Player/EmulatorJS/Base.vue
zurdi15: It basically makes a double check that the local loader.js file is correct. It was happening to me that sometimes somehow it was being loaded but incorrectly and it didn't fallback to the CDN but instead it was raising an error.

PR #1902, opened 2025-05-21
fix: iOS full screen play

zurdi15:
**Description**
This PR is an attempt to fix te iOS issue that makes impossible to play fullscreen

**Checklist**

### DIFF
--- a/frontend/src/views/Player/EmulatorJS/Player.vue
+++ b/frontend/src/views/Player/EmulatorJS/Player.vue
@@ -20,6 +20,8 @@ import {
 } from "./utils";
 import type { Emitter } from "mitt";
 import type { Events } from "@/types/emitter";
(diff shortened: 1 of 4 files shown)


gantoine (approved):

frontend/src/views/Player/EmulatorJS/Player.vue
```diff
@@ -338,12 +345,6 @@ window.EJS_onGameStart = async () => {
   <div id="game" />
 </template>
 
-<style scoped>
-#game {
-  height: 100%;
-}
-</style>
```
gantoine, line 345: why was this removed?

zurdi15: I have been testing and with the autosrcoll to top on play and the little style it already has was not needed

PR #1917, opened 2025-05-27
feat: Added optional 3D tilt effect to cards

zurdi15:
**Description**
Added optional 3D tilt effect for cards when hover on them

**Checklist**

#### Screenshots

image

### DIFF
--- a/frontend/src/locales/de_DE/settings.json
+++ b/frontend/src/locales/de_DE/settings.json
@@ -2,6 +2,8 @@
   "continue-playing-as-grid": "Weiterspielen als Raster",
   "continue-playing-as-grid-desc": "Zeige die Weiterspielen-Karten als Raster auf der Startseite",
   "email": "E-Mail",
+  "enable-3d-effect": "3D-Effekt aktivieren",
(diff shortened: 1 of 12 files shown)


frontend/src/views/Gallery/Collection.vue
```diff
@@ -39,8 +40,16 @@ const {
...
 const emitter = inject<Emitter<Events>>("emitter");
+const isHovering = ref(false);
+const hoveringRomId = ref();
+const openedMenu = ref(false);
+const openedMenuRomId = ref();
+const storedEnable3DEffect = localStorage.getItem("settings.enable3DEffect");
+const enable3DEffect = ref(
+  isNull(storedEnable3DEffect) ? false : storedEnable3DEffect === "true",
+);
+let timeout: ReturnType<typeof setTimeout>;
```
gantoine, line 52: this is duplicated like 5 times, we should probably (not in this PR) make this a utility

zurdi15: Actually I wanted this to be merged because my plan is to create a pinia store for settings. That will avoid this being duplicated everywhere. Also the handlers can be extracted to a utility too.

PR #1924, opened 2025-05-29
misc: Optimized covers storing them as webp format

zurdi15:
**Description**
Now any kind of new artowrk (uploaded or fetched) will be converted to `.webp` format to improve performance

**Checklist**


gantoine:
Thinking about this more it's likely webp images won't be supported on muos or playnite, for when we decide to fetch cover art for this plugins. Going to close this PR given that case, and I'm sure there are other image perf changes we can make in the future.

PR #1931, opened 2025-06-05
Add playmatch hash-based integration

DevYukine:
**Description**
This adds basic support for matching roms via hashes (and filename + filesize) with [playmatch](https://github.com/RetroRealm/playmatch)

This works by adding a new metadata handler for Playmatch and then when scanning roms and before searching on IGDB we send a request with filename, filesize and the hashes to playmatch and check if there is anything in the database

My idea here is to make this fully optional, i've added a flag to disable this via env for this

also this code can probably be refactored and improved as im not a great python developer myself and from first looks i can already tell that it should only be run once hashes are calculated as in the current way is run before hashes are calculated and only matches by filename + filesize (which is also unique but just not as good as hashes when it comes to potential false positives)

i've tested this succesfully locally in a develop environment after i got this to work (only took me a few hours... *we love setting up an dev env*)

I've recently released a tool which basically does this as script (https://github.com/DevYukine/romm-hash-matcher) and with that also improved Playmatch performance and caching to ensure that it will survive all the traffic from RomM instances (Playmatch has Cloudflare caching + an own caching layer and the average response time for hash identify calls is ~50ms for non cached ones)

I'm interested in feedback of all kinds here 👍 

Some important information:

Playmatch currently only supports IGDB (more are planned) and Redump and No-Intro Dat files (which are the most popular) so we can right now probably just implement it for IGDB matching

**Checklist**

#### Screenshots
as this is only in the backend i can't show any nice UI but this is how the console log output looks right now
image


zurdi15:
Pretty solid PR so far 👌

DevYukine:
I've now added an env variable which allows playmatch to be disabled (like i suggested in the PR itself)

I've also changed request exceptions so that they get logged but then identify just silently fails instead throwing an error as Playmatch errors should not stop a Scan in my opinion

DevYukine:
I'm also still not happy where i call playmatch identify right now as this will not work great for new instances

right now if you start a new instance it will do the matching before the hash calculation and always call playmatch with only filename and filesize, in the long run this should probably be changed to run hash calculation before doing metadata matching (hasheous would have the same issues as playmatch here)

but right now this will work fine for older instances which already have run a scan before which triggered hash calculation but on new instances it will always use playmatch without hashes

I'm not sure if this is the scope of such a change but i'm interested if you think its worth changing here

DevYukine:
This PR requires https://github.com/rommapp/romm/pull/1981 to be merged

PR #1940, opened 2025-06-06
Cleanup lingeting t_thumb in cover and screenshto art

gantoine:
**Description**

Lingering `t_thumb` values in `url_cover` and `url_screenshots` will cause romm to overrite cover and screenshot art with thumbnails on partial scans. This replaces all uses of `t_thumb` in those fields with `t_1080p` for covers and `t_720p` for screenshots. Tested on both mariadb and postgres locally.

Some `cover_url` fields in `igdb_metadata` might still use `t_thumb` and would still be affected, but there's no easy way top clean those up through SQL.

**Checklist**

### DIFF
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -502,26 +502,22 @@ async def _identify_rom(
                     rom_id=_added_rom.id,
                     hash=ra_hash,
                 )
-                _added_rom.ra_id = ra_handler_rom.get("ra_id", "")
-                _added_rom.ra_metadata = ra_handler_rom.get("ra_metadata", {})
-                for a in _added_rom.ra_metadata.get("achievements", {}):
-                    # Store both normal and locked version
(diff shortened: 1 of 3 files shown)


backend/endpoints/sockets/scan.py
```diff
@@ -502,26 +502,22 @@ async def _identify_rom(
...
+                        badge_url_lock = a.get("badge_url_lock", None)
+                        badge_path_lock = a.get("badge_path_lock", None)
+                        if badge_url_lock and badge_path_lock:
+                            await fs_resource_handler.store_badge(
+                                badge_url_lock, badge_path_lock
+                            )
+                        badge_url = a.get("badge_url", None)
+                        badge_path = a.get("badge_path", None)
+                        if badge_url and badge_path:
+                            await fs_resource_handler.store_badge(badge_url, badge_path)
```
gantoine, line 520: This is just some defensive coding, behaviour should be the same

zurdi15 (approved):

PR #1956, opened 2025-06-09
fix: Use aiohttp for RetroAchievements API calls

adamantike:
**Description**
This change replaces the `httpx` client with `aiohttp` for the RetroAchievements API service.

The main reason for this change is that `httpx` has an unavoidable log line with `INFO` level, which includes the request full URL, containing the user's API key.

`httpx` has had an
[open discussion](https://github.com/encode/httpx/discussions/2765) regarding this security issue for almost two years.

The change to `aiohttp` is painless, and would allow us to migrate more of the codebase to it in the future, to avoid leaking sensitive information in logs.

**Checklist**


gantoine (approved):
💯

PR #1977, opened 2025-06-12
feat: Mark missed files from filesystem as "missing"

### LINKED ISSUE #1784: [Feature] Add a missing games tab
If Romm doesn't detect a game that has already been scanned, it automatically deletes its metadata, even if the user has written important notes or other items they don't want to lose. All it takes is for the user to accidentally change the name of a game's folder or to organize their library, and the metadata for that game will be deleted the next time Romm scans, without warning or backtracking!

The solution would be to do as most popular media management tools like AudiobookShelf do, by classifying all media that were not found in a "missing games" tab. This would allow the user to also delete them from the romm interface, to correctly rename their files or folders to match the metadata paths, or to attach the metadata to the new game paths.

zurdi15:
**Description**
This PR changes the scanner behaviour when any entity (rom, platform or firmware) is missing from the filesystem. Now instead of purging it from the database, will be marked as ``missing`` avoiding user data loss. If the file is restored (same path, same filename) and a scan is performed, will be unmarked as ``missing`` again.

**Checklist**

#### Screenshots

image

image

Closes #1784

### DIFF
--- a/backend/alembic/versions/0042_add_missing_rom_flag.py
+++ b/backend/alembic/versions/0042_add_missing_rom_flag.py
@@ -27,10 +27,50 @@ def upgrade():
             sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
         )
 
+    with op.batch_alter_table("rom_files", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
+        )
+
+    with op.batch_alter_table("firmware", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
+        )
+
+    with op.batch_alter_table("saves", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
+        )
+
+    with op.batch_alter_table("states", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
+        )
+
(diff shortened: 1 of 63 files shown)


backend/alembic/versions/0042_add_missing_rom_flag.py
```diff
@@ -0,0 +1,76 @@
...
+revision = "0042_add_missing_fields"
+down_revision = "0041_assets_t_thumb_cleanup"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("platforms", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("missing", sa.Boolean(), nullable=False, server_default="0")
```
gantoine: could these be called `missing_from_fs` or `missing_from_disk`? i find `missing` to be vague

backend/handler/database/firmware_handler.py
```diff
@@ -83,13 +83,14 @@ def purge_firmware(
...
                 and_(
                     Firmware.platform_id == platform_id,
                     Firmware.file_name.not_in(fs_firmwares_to_keep),
                 )
             )
+            .values(**{"missing": True})
             .execution_options(synchronize_session="evaluate")
         )
-        return purged_firmware
+        return missing_firmware
```
gantoine, line 96: reusing `purge_*` was a good idea 👍🏼

gantoine (changes requested):
thanks for keeping the implementation clean 🫡

backend/handler/database/firmware_handler.py
zurdi15: It was easy to find every place where to replace x)

PR #1981, opened 2025-06-13
Calculate hashes before fetching metadata

gantoine:
**Description**

This PR moves the calculation of hashes before metadata scans. While this does have an impact on performance, it's a requirement for implementing playmatch and hasheous, and brings RA matching more in line with our other matchers.

**Checklist**


backend/handler/database/roms_handler.py
```diff
@@ -787,16 +787,15 @@ def update_rom_file(self, id: int, data: dict, session: Session = None) -> RomFi
         return session.query(RomFile).filter_by(id=id).one()
 
     @begin_session
-    def missing_rom_files(
+    def purge_rom_files(
```
adamantike, line 790: Is this change expected, or a bad rebase after Zurdi's changes?

gantoine: it's expected, we want to purge missing files since a) they dont have metadata and b) we can rescan to repopulate them

PR #1997, opened 2025-06-18
Use hasheous as igdb proxy

gantoine:
**Description**

This PR uses hasheous as a metadata proxy for games with IGDB matches. This means that users won't have to setup their own IGDB API keys and account to fetch (most) metadata, as hasheous will proxy the important ones.

**Checklist**


backend/handler/metadata/igdb_handler.py
```diff
@@ -610,15 +610,15 @@ async def get_matched_roms_by_name(
             alternative_roms_ids = []
             for rom in alternative_matched_roms:
                 alternative_roms_ids.append(
-                    rom.get("game").get("id", "")
+                    pydash.get(rom, "game.id", "")
```
adamantike, line 613: I prefer to avoid the magical `pydash.get` function, as we lose typing information. In this case, Mypy won't be able to alert us if we are trying to append the wrong value type to this list.

gantoine: In general I would agree, but I'd like to fix/refactor all these in 1 go in a future PR and remove all of them at once in igdb_handler

PR #1998, opened 2025-06-18
Fix Retroachievements URL in README

Azelphur:
Nothing like a new user to break the docs :rofl:

Existing link is 404, I believe this is the correct link?


zurdi15:
You are totally right! Thank you for pointing it out. A couple of things:

1. Can you change the other provider links in the README? I just realized everything is outdated
2. Can you open the PR against ``master`` instead of ``release``?

Azelphur:
Sure :)

Should be all sorted now.

PR #2015, opened 2025-06-24
Update DEVELOPER_SETUP.md

Konscription:
Update DEVELOPER_SETUP.md to include all extras in poetry sync and explain pytest command

**Description**
 * added the --all-extras flag to the poetry sync command. 
 while following this guide, I had to dig into why my setup would not run tests correctly out of the box. 
 i found that by just adding this flag it would install everything needed to run tests without any fuss.
 while is could just be my unfamiliarity with poetry, for ease of use, i believe adding it here would be beneficial. 
* added an additional pytest example at the bottom following the vscode task format with a explanatory description above.
* 
**Checklist**

- [*] I've tested the changes locally
- [*] I've updated relevant comments


gantoine:
thanks for the PR! can you re-open this against the `master` branch? like checkout `master` and make your changes against it, there are some changes on there to this file already and rebasing isnt possible.

PR #2028, opened 2025-06-28
fix: Add debounce to RomUser changes

adamantike:
**Description**
Stop bombarding the API with updates when editing personal data, like notes (which currently trigger an API call on each keystroke), and status (e.g. changing values on the "Completion %" slider triggers an API call on each change).

This change introduces a debounce of 500ms before sending updates to the API.

**Checklist**


gantoine (approved):
beautiful

PR #2029, opened 2025-06-28
misc: Specify English language for all Markdown blocks

adamantike:
**Description**
`md-editor-v3` defaults to Simplified Chinese for `MdEditor` and `MdPreview` components. This change explicitly sets the language to English (`en-US`) for all instances of these components.

Ideally, this should take the user's language preference into account, and a common component can be created to avoid forgetting to set common settings on new instances.

**Checklist**


gantoine (approved):
good start

PR #2043, opened 2025-07-01
Fix for setting the correct forward proto when behind ssl proxy

### LINKED ISSUE #1737: [Bug] Tinfoil feed is not aware of HTTPS settings in reverse proxy
**RomM version**
v3.8.1

**Describe the bug**
Tinfoil integration is not aware of reverse proxy settings, thus creates feeds with `http` only. If the RomM expects users to use `https` only for serving files, Tinfoill will not work properly on game installation.

**To Reproduce**

1. Update the library with some switch games
2. Setup the instance with tinfoil integration according [official doc](https://docs.romm.app/latest/System-Setup/Tinfoil-Integration/)
3. Go to the tinfoil endpoint `/api/tinfoil/feed/` to see the list of files
4. See error in urls. All links seem to be in `http` regardless of reverse proxy settings of RomM

**Additional context**
Maybe have an environment variable as configuration item for creating urls in Tinfoil feed with `http` or `https`?

**Screenshots**

Image

starkmapper:
**Description**
Fixes #1737 

Feeds are generated using the "Forward protocol".
This was set to "http" by default in the nginx config, regardless if it is set or not.

This change will set it to `https`, if that's what it was set to already, or to `$scheme` if it's unset.

**Checklist**

- ~[ ] I've added unit tests that cover the changes~ Not applicable in my opinion.

### DIFF
--- a/docker/nginx/templates/default.conf.template
+++ b/docker/nginx/templates/default.conf.template
@@ -2,6 +2,12 @@
 # by using `envsubst` to replace the environment variables in the template with
 # their actual values.
 
+# Helper to get scheme regardless if we are behind a prox or not
+map $http_x_forwarded_proto $forwardscheme {
+    default $scheme;
+    https https;
+}
+
 server {
     root /var/www/html;
     listen ${ROMM_PORT};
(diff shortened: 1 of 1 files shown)


adamantike (approved):
Looks good, thanks for your contribution!

PR #2059, opened 2025-07-07
fix: Correctly use Python 3.13 in Docker image

adamantike:
**Description**
The `production-stage` stage was depending on the `python3` package from Alpine, which at the moment of writing is still Python 3.12.

To avoid relying on Alpine's package and releases, we now copy the Python installation directly from its official Docker image.

Other options were tested but did not work:
- Trying to install Python 3.13 using `apk`. As mentioned, Alpine does not support Python 3.13 yet.
- Installing Python 3.13 using `pyenv`. The installation worked, but the generated image was too large. Related `pyenv` discussion: https://github.com/orgs/pyenv/discussions/2868
- Installing Python 3.13 using `uv`. Only worked for `amd64` architecture, but the installation failed on `arm64`.

**Checklist**


gantoine (approved):
yooooo this solution is great, wouldnt have thought of it lol

PR #2210, opened 2025-08-05
[POC] Server side cheat code

wctang:
**Description**
If such a feature could be implemented, it would be very convenient for playing on different devices. Currently, cheat codes are stored in a note using a special format and are automatically loaded upon startup. It would be more suitable to use a database for this.
```
- cheat: Unlimited Health: 82020646+04FF
- cheat: Unlimited Super Bar: 8202064A+0400
```
**Checklist**


gantoine:
Cool idea! We are planning to allow users to set the emulatorjs config per game/platform before starting a game, so I don't think using the note is viable. If you want to give that feature a shot hop in the discord and ping the team, happy to talk it over.

PR #2216, opened 2025-08-05
fix: Debounce search input in app search bar

adamantike:
**Description**
Using `nextTick` was triggering the search as soon as the user typed, making unneeded requests and the frontend ignoring other characters added to the search term if the user typed too fast.

This change debounces the search input to reduce the number of requests to the backend, and cancels the debounce if the user presses Enter, ensuring that the search is performed immediately instead.

**Checklist**


adamantike:
Closed in favor of #2208.

PR #2220, opened 2025-08-06
[ROMM-2142] Custom SGDB title match

### LINKED ISSUE #2142: [Bug] Steam Grid DB matching (incorrectly) by name? (4.0.0)
**RomM version**

4.0.0

**Describe the bug**

Using the "Chef's Choice" metadata setup, some games match Hasheous + IGDB + RA by the game hash correctly but end up with a wrong cover (wrong game) from SteamGridDB.

I think SGDB searches must still be trying to match by name, because the bad results are often from games that kind of sound like the actual game but are subtly different. See the screenshot below for "Solar Fox" (Atari), which got the cover for a game called "Solar Flux". Another example is "Quadrun" matching to "Quadrant".

Moreover, there doesn't seem to be any way to _fix_ this issue in the UI, other than manually uploading a new cover image. If I edit the game, delete the cover, and search again, it _only_ shows results from SGDB, none of them matching the game; it will not allow me to fall back to the IGDB cover image which in many cases is perfectly fine.

**To Reproduce**

Import any of the games mentioned above. Can provide more examples if need be.

**Expected behavior**

Avoid using SGDB images if the game doesn't match; fall back to the cover image from the primary metadata provider in these cases.

Ideally, this should be completely automatic when importing games. The next best thing would be to use SGDB anyway, but flag/alert me when the name isn't a perfect match, and give some way to revert to IGDB/SS/other primary metadata in the cover editor.

**Screenshots**

Import success

Wrong Cover

Manual cover search

**Additional context**

Otherwise impressed with the speed and accuracy of the new Hasheous-based system in 4.0.0. So far the only thing that's misbehaving is SGDB.

gantoine:
**Description**

This PR changes SGDB matching to use a custom algorithm that mixes token overlap, sequence similarity and (custom) Levenshtein distance with different ratios to match the most likely game.

Fix #2142 

**Checklist**


backend/handler/metadata/sgdb_handler.py
```diff
@@ -32,7 +54,56 @@ class SGDBRom(TypedDict):
...
+        Returns a score between 0 and 1, where 1 is a perfect match.
+        """
+        game_normalized = self.normalize_search_term(game_name, remove_articles=False)
+
+        # Exact match gets the highest score
+        if search_normalized == game_normalized:
+            return 1.0
+
+        # Split into tokens for word-based matching
+        search_tokens = set(re.findall(r"\b\w+\b", search_normalized.lower()))
```
adamantike: nit: We can compile this regex, as we expect it to be called pretty frequently.

backend/handler/metadata/sgdb_handler.py
```diff
@@ -62,28 +133,19 @@ async def get_details_by_names(self, game_names: list[str]) -> SGDBRom:
...
-                game_name_normalized = self.normalize_search_term(
-                    game["name"], remove_articles=False
-                )
-                distance = levenshtein_distance(
-                    game_name_normalized, search_term_normalized
+                similarity_score = self._calculate_title_similarity(
+                    search_term, game["name"]
                 )
-                game_distances.append((game, distance))
+                game_scores.append((game, similarity_score))
```
adamantike, line 144: This could be in a separate PR, but getting a similarity score of `1.0` could trigger a break here, so there's no need to keep calculating for the rest of the list.

adamantike (approved):

PR #2224, opened 2025-08-06
feat: Migrate scheduler to native RQ process

adamantike:
**Description**
This change replaces our custom `scheduler.py` script with the `rqscheduler` command, allowing us to run the RQ scheduler as a separate, low-memory process, by avoiding the need to maintain the Python app in memory.

* Remove `scheduler.py` script.
* Move initialization of scheduled tasks to `worker.py`.
* Update `docker/init_scripts/init` to start the `rqscheduler` command instead of the custom script.
* Fix scheduled tasks' `func` paths to the new project structure.
* Temporarily use a fork of `rq-scheduler` to support username and SSL settings in the `rqscheduler` command.

**Checklist**
<sup>Please check all that apply.</sup>

- [x] I've tested the changes locally
- [x] I've updated relevant comments
- [x] I've assigned reviewers for this PR
- [ ] I've added unit tests that cover the changes

PR #2227, opened 2025-08-07
Migrate worker to native rq process

gantoine:
**Description**

Hot on the heels of #2224, we do the same thing but for the worker. Task initialization now happens in `main.py` since `worker.py` was purged.

**Checklist**

### DIFF
--- a/backend/watcher.py
+++ b/backend/watcher.py
@@ -1,11 +1,12 @@
 import os
 import sys
 from datetime import timedelta
-from typing import Literal
 
 import sentry_sdk
 from config import (
     ENABLE_RESCAN_ON_FILESYSTEM_CHANGE,
+    HASHEOUS_API_ENABLED,
+    LAUNCHBOX_API_ENABLED,
     LIBRARY_BASE_PATH,
     RESCAN_ON_FILESYSTEM_CHANGE_DELAY,
     SENTRY_DSN,
@@ -13,7 +14,12 @@ from config import (
 from config.config_manager import config_manager as cm
 from endpoints.sockets.scan import scan_platforms
 from handler.database import db_platform_handler
-from handler.scan_handler import ScanType
+from handler.metadata.igdb_handler import IGDB_API_ENABLED
+from handler.metadata.moby_handler import MOBY_API_ENABLED
+from handler.metadata.ra_handler import RA_API_ENABLED
+from handler.metadata.sgdb_handler import STEAMGRIDDB_API_ENABLED
+from handler.metadata.ss_handler import SS_API_ENABLED
+from handler.scan_handler import MetadataSource, ScanType
 from logger.formatter import CYAN
 from logger.formatter import highlight as hl
 from logger.logger import log
@@ -50,7 +56,6 @@ def on_any_event(
     src_path: str,
     _dest_path: str,
     event_type: str,
-    fs_obj: Literal["file", "directory"],
 ):
     if event_type not in valid_events:
         return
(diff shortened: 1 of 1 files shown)


.vscode/tasks.json
```diff
@@ -14,12 +14,6 @@
       "command": "cd backend && uv run python3 main.py",
       "problemMatcher": []
     },
-    {
-      "label": "Launch worker",
```
adamantike, line 18: We also need to remove dependency `Launch worker` from step `Launch full stack`.

backend/main.py
```diff
@@ -133,6 +140,19 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
...
+    if ENABLE_SCHEDULED_UPDATE_SWITCH_TITLEDB:
+        log.info("Starting scheduled update switch titledb")
+        update_switch_titledb_task.init()
+
+    if ENABLE_SCHEDULED_UPDATE_LAUNCHBOX_METADATA:
+        log.info("Starting scheduled update launchbox metadata")
+        update_launchbox_metadata_task.init()
+
     # Run migrations
     alembic.config.main(argv=["upgrade", "head"])
```
adamantike: I think it'll be safer to run migrations before scheduling tasks, so workers don't start processing tasks while the database is in a previous schema version.

docker/init_scripts/init
```diff
@@ -173,38 +174,54 @@ start_bin_scheduler() {
...
+	PYTHONPATH="/backend:${PYTHONPATH-}" rq worker \
+		--path /backend \
+		--pid /tmp/rq_worker.pid \
+		--url "${redis_url}" \
+		high default low &
+}
+
+start_bin_watcher() {
+	info_log "Starting watcher"
+	watchmedo shell-command \
```
adamantike, line 204: Just in case, let's add the `watchmedo` extra for `watchdog` to our dependency definition:
https://github.com/gorakhargosh/watchdog/blob/d197487fb6b51aef05569bea0554f3272cb79039/setup.py#L64

backend/watcher.py
```diff
@@ -23,99 +24,74 @@
...
 
-    def on_any_event(self, event: FileSystemEvent) -> None:
-        """Catch-all event handler.
 
-        Args:
-            event: The event object representing the file system event.
-        """
-        if not ENABLE_RESCAN_ON_FILESYSTEM_CHANGE:
-            return None
+def on_any_event(src_path: str, _dest_path: str, event_type: str, object: str):
```
adamantike:
```suggestion
def on_any_event(src_path: str, _dest_path: str, event_type: str, obj: Literal["file", "directory"]) -> None:
```

backend/watcher.py
```diff
@@ -23,99 +24,74 @@
...
 
-class EventHandler(FileSystemEventHandler):
-    """Filesystem event handler"""
+valid_events = [
+    DirCreatedEvent.event_type,
+    DirDeletedEvent.event_type,
+    FileCreatedEvent.event_type,
+    FileDeletedEvent.event_type,
+    FileSystemMovedEvent.event_type,
+]
```
adamantike:
```suggestion
valid_events = frozenset(
    (
        DirCreatedEvent.event_type,
        DirDeletedEvent.event_type,
        FileCreatedEvent.event_type,
        FileDeletedEvent.event_type,
        FileSystemMovedEvent.event_type,
    )
)
```

backend/watcher.py
```diff
@@ -23,99 +24,74 @@
...
-            # Otherwise trigger a rescan for the specific platform
-            log.info(f"Change detected in {hl(fs_slug)} folder, {rescan_in_msg}")
-            tasks_scheduler.enqueue_in(
-                time_delta,
-                scan_platforms,
-                [db_platform.id],
-                scan_type=ScanType.QUICK,
-            )
+    # Any change to a platform directory should trigger a full rescan
+    if object == "directory" and event_src.count("/") == 1:
```
adamantike:
```suggestion
    if obj == "directory" and event_src.count("/") == 1:
```

docker/init_scripts/init
```diff
@@ -173,38 +174,54 @@ start_bin_scheduler() {
...
+# Commands to start RQ worker
+start_bin_rq_worker() {
+	info_log "Starting rq worker"
+
+	# Build Redis URL properly
+	local redis_url
+	if [[ -n ${REDIS_USERNAME-} && -n ${REDIS_PASSWORD-} ]]; then
+		redis_url="redis${REDIS_SSL:+s}://${REDIS_USERNAME}:${REDIS_PASSWORD}@${REDIS_HOST:-127.0.0.1}:${REDIS_PORT:-6379}/${REDIS_DB:-0}"
+	elif [[ -n ${REDIS_PASSWORD-} ]]; then
+		redis_url="redis${REDIS_SSL:+s}://:${REDIS_PASSWORD}@${REDIS_HOST:-127.0.0.1}:${REDIS_PORT:-6379}/${REDIS_DB:-0}"
```
adamantike: We are not correctly handling the scenario when there's a user but no password. The simple approach for this would be:

1. If there's a password: Use `${REDIS_USERNAME}:${REDIS_PASSWORD}`, so even if there's no user, the auth will correctly set `:<password>`.
2. If there's a user: Use `${REDIS_USERNAME}`. We need to avoid adding a trailing `:` in this scenario.
2. Fallback to scenario without Basic authentication.

adamantike (approved):
🚀🚀🚀

PR #2228, opened 2025-08-07
misc: Replace emoji dependency with constants

adamantike:
**Description**
The `emoji` library has been removed, in favor of using constants for the few emojis used in the codebase. This reduces memory usage, and avoids calling `emojize` for Python to discover where to replace emojis in pre-defined strings.

**Checklist**


gantoine (approved):
🍋

PR #2250, opened 2025-08-11
fix: Make default ROMM_AUTH_SECRET_KEY consistent between workers

adamantike:
**Description**
When the `ROMM_AUTH_SECRET_KEY` environment variable is not set, and the Gunicorn application server spawns multiple worker processes, each worker was generating a different random secret key. This led to intermittent authentication failures, hard to debug and reproduce.

To fix this, we now set a default secret key that is consistent across all workers. For security reasons, it is still recommended to set a unique and secure key via the `ROMM_AUTH_SECRET_KEY` environment variable. When not set, the default key is now derived from the Docker's `HOSTNAME` environment variable, or a hardcoded fallback value.

**Checklist**

### DIFF
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -1,5 +1,4 @@
 import os
-import secrets
 from typing import Final
 
 import yarl
@@ -9,7 +8,7 @@ load_dotenv()
 
 
 def str_to_bool(value: str) -> bool:
-    return value.lower() in ("true", "1")
+    return value.strip().lower() in ("1", "true", "yes", "on")
 
 
 ROMM_BASE_URL = os.environ.get("ROMM_BASE_URL", "http://0.0.0.0")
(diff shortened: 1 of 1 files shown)


backend/config/__init__.py
```diff
@@ -98,8 +97,11 @@ def str_to_bool(value: str) -> bool:
 TGDB_API_ENABLED: Final = str_to_bool(os.environ.get("TGDB_API_ENABLED", "false"))
 
 # AUTH
-ROMM_AUTH_SECRET_KEY: Final = os.environ.get(
-    "ROMM_AUTH_SECRET_KEY", secrets.token_hex(32)
+# The secret key has a default value to not difer between concurrent workers.
+# For security reasons, you should change this to a unique and secure key.
+ROMM_AUTH_SECRET_KEY: Final[str] = os.environ.get(
+    "ROMM_AUTH_SECRET_KEY",
+    os.environ.get("HOSTNAME", "(H(3>X(V6n0~n@5>bzpe.'K^pG*T.q7_m0aETIs?X}aF#y{@z&"),
```
gantoine, line 104: Would it be possible instead to generate one in the `init` script and pass it through as a env var? IMO we should actually just throw an error if `ROMM_AUTH_SECRET_KEY` isn't set by the user.

This is also going to make Aikido very mad lol.

PR #2260, opened 2025-08-13
misc: Move cache initialization to startup script

adamantike:
**Description**
Guarantee that cache is initialized during startup, and only once, instead of every time a `MetadataHandler` object is instantiated.

Also, improve logic to determine `fixtures` paths.

**Checklist**


gantoine (approved):
yessss very nice

PR #2314, opened 2025-08-25
Console Mode

kodaxx:
## Summary
- No breaking changes for standard RomM UI use
- Update to the latest version of console mode, which includes:
  - Code cleanup, optimization, organization
  - Refined navigation experience
  - Subtle sound effects
  - Reduced resource usage
  - Color/theme tweaks

## Implementation Notes
- State: Pinia console/navigation store for selection index tracking  
- Input System: Gamepad/keyboard actions already mapped; dispatch layer now calls `sfxForAction()`.  
- Audio: Lazy-created `AudioContext` (only after first user gesture). No assets loaded—zero bundle bloat.  

## Testing / Verification Steps
1. Launch frontend, enable controller or keyboard navigation.
2. Open a platform, move selection several items deep, enter a game, etc.

## Future Work (Optional / Deferred)
- Add more supported system images
- Settings: Enable/disable SFX
- Distinct “invalid move” or boundary edge soft bump sound.
- Delete unused legacy scroll fields if confirmed not needed.
- Theming support

## Checklist


frontend/src/console/components/BackButton.vue
```diff
@@ -0,0 +1,38 @@
...
+
+<script setup lang="ts">
+import { toRefs } from "vue";
+import { useRouter } from "vue-router";
+
+const props = defineProps<{
+  text?: string;
+  onBack?: () => void;
+}>();
+const { text, onBack } = toRefs(props);
```
gantoine: whats the purpose of `toRefs` here? afaik props are already reactive

kodaxx: toRefs is used here to destructure props into text and onBack without losing reactivity. Props are already reactive, yes, but destructuring it normally would give plain values that don’t update if the parent changes. Wrapping with toRefs makes each property a ref keeping them reactive and in sync. There's probably 100 ways to do it, if you have a more elegant solution, I'm not tied to this one

docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       - ROMM_BASE_PATH=/app/romm
     ports:
       - "3000:3000" # Frontend dev server
-      - "${DEV_PORT:-5000}:5000" # Backend API
+      - "${DEV_PORT:-5050}:5050" # Backend API
```
adamantike: I would revert this change. If you need to change from the default port `5000`, that's why we have `DEV_PORT` as an option.

kodaxx: Yeah, this one was an accidental push anyway. Good eye

kodaxx: Although, on my Mac, Control Center is listening on port 5000. Starting with macOS Monterey, Apple uses ports 5000 (and 7000) for AirPlay Receiver functionality.

When AirPlay Receiver is enabled, the ControlCenter process binds to port 5000, making it unavailable for other local apps. 

So it may be a worthwhile change anyway

adamantike: Port `5000` was a very common config even before Apple decided to use it for AirPlay Receiver. I would keep port 5000, and give the option to set `DEV_PORT` if needed.

I've found myself completely disabling AirPlay Receiver for this same reason.

gantoine (approved):

ghost:
Thank you so much!!!!

PR #2322, opened 2025-08-27
Add options to gunicorn call

gantoine:
**Description**

**Checklist**


env.template
```diff
@@ -93,3 +90,12 @@ YOUTUBE_BASE_URL=https://www.youtube.com
 
 # Logging
 LOGLEVEL=DEBUG
+
+# Gunicorn (optional)
```
adamantike: I would avoid going back to `GUNICORN_` prefixed environment variables, in case we want to switch to a different web server, without breaking compatibility with existing installations (that's why we switched to `WEB_CONCURRENCY` in the first place).

PR #2334, opened 2025-08-28
Convert images to webp task

gantoine:
**Description**

This PR adds a new scheduled task to convert existing cover images to WebP format. When enabled, images are converted on scan, nightly, and on-demand is the task is manually called. The client will attempt to fetch webp images if the scheduled task is enabled (not a perfect solution...). Also standardizes loading base data (heartbeat, config, etc.) through pinia stores and in parallel.

**Checklist**


backend/tasks/scheduled/convert_images_to_webp.py
```diff
@@ -0,0 +1,221 @@
...
+            List of paths to image files that can be converted to WebP
+        """
+        if not self.resources_path.exists():
+            log.warning(f"Resources path does not exist: {self.resources_path}")
+            return []
+
+        image_files = [
+            p
+            for p in self.resources_path.rglob("**/cover/*")
+            if p.is_file() and p.suffix.lower() in ImageConverter.SUPPORTED_EXTENSIONS
```
adamantike: Should we call `p.is_file(follow_symlinks=False)` instead, to avoid symlinks that could be pointing to other images (and end up processing the same image multiple times)?

backend/tasks/scheduled/convert_images_to_webp.py
```diff
@@ -0,0 +1,221 @@
...
+        ]
+
+        return sorted(image_files)  # Sort for consistent processing order
+
+    def _process_single_image(self, image_path: Path) -> bool:
```
adamantike, line 130: nitpick: I like Go's design where a method like this could return `(bool, str)`, where the latter is the success/error message. That way, counter updates and `errors` are changed within `run()`, reducing the need for shared state between methods.

backend/config/__init__.py
```diff
@@ -151,6 +151,13 @@ def str_to_bool(value: str) -> bool:
     "SCHEDULED_UPDATE_LAUNCHBOX_METADATA_CRON",
     "0 5 * * *",  # At 5:00 AM every day
 )
+ENABLE_SCHEDULED_CONVERT_IMAGES_TO_WEBP: Final = str_to_bool(
+    os.environ.get("ENABLE_SCHEDULED_CONVERT_IMAGES_TO_WEBP", "false")
+)
+SCHEDULED_CONVERT_IMAGES_TO_WEBP_CRON: Final = os.environ.get(
+    "SCHEDULED_CONVERT_IMAGES_TO_WEBP_CRON",
+    "0 6 * * *",  # At 6:00 AM every day
```
adamantike: nitpick: Probably something for later, but we could put all tasks unrelated to Scan into the same hour.
I feel otherwise, as we add more tasks, it will make the system to have scheduled processes running for most part of the day (if using default values).

adamantike (approved):
:fire:

PR #2340, opened 2025-08-28
fix devcontainer launch

MauGx3:
**Description**
issue: devcontainer doesn't launch properly as is

problematic code: filepath for docker-compose.yml and not using the existing env.template file

fix: change filepath to refer to root and use env.template in the devcontainer for faster usage

**Checklist**

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,5 @@
 {
   "name": "ROMM Development",
-  // Use the repository root docker-compose.yml so the romm-dev service uses the project's root Dockerfile
   "dockerComposeFile": ["../docker-compose.yml"],
   "service": "romm-dev",
   "workspaceFolder": "/app",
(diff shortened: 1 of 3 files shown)


gantoine:
Using `env.template` is incorrect, you need to copy it to `.env` and use that file. If you remove that change but keep the `dockerComposeFile` change I'll merge this in!

gantoine (changes requested):
-

MauGx3 (commented):
@gantoine yeah guess I was too sleepy last night, sorry!

gantoine:
Nice, I'll test this when im back in town monday!

.devcontainer/devcontainer.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "ROMM Development",
-  "dockerComposeFile": "docker-compose.yml",
+  // Use the repository root docker-compose.yml so the romm-dev service uses the project's root Dockerfile
```
adamantike: Does devcontainer support JSON5? Otherwise, we can't have comments here.

gantoine (approved):
thanks for the PR!

PR #2345, opened 2025-08-30
misc: Create IGDB service adapter

adamantike:
**Description**
Add a new service adapter for the IGDB API, to separate concerns with RomM's handler for metadata. This adapter is agnostic to the handler and only provides methods to interact with the API, and correctly return typed responses.

The API authorization was also improved to not rely on decorating each method that makes requests, but instead using an `aiohttp` middleware to automatically add the required headers to each request.

Utils `mark_expanded` and `mark_list_expanded` where added to help narrow the types of IGDB's expandable fields when we know they are expanded, for `mypy` type checking.

**Checklist**

### DIFF
--- a/backend/adapters/services/igdb.py
+++ b/backend/adapters/services/igdb.py
@@ -1,0 +1,202 @@
+import asyncio
+import http
+import json
+from collections.abc import Sequence
+from functools import partial
+from typing import TYPE_CHECKING
+
+import aiohttp
+import yarl
+from adapters.services.igdb_types import Game
+from aiohttp.client import ClientTimeout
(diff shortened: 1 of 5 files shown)


gantoine (approved):
💯

adamantike:
@gantoine, can you enforce a merge when you get the time? Seems something broke in GitHub, and it's waiting for the `typecheck` action, which will never run as there are no changes in `frontend/`.

gantoine:
I've removed the check for `typecheck` action, now only `trunk check` is required to run. PR can be merged as-is. 👍🏼

PR #2350, opened 2025-08-30
fix: Invalidate fixture cache on file change

### LINKED ISSUE #2347: [Bug] Bios files are not correctly getting verification status (Outdated fixture cache in Redis)
**RomM version**
4.1.6

**Describe the bug**
Uploaded a known frimware through webui ("64dd"). Even though manually verified the file with data from project files, the bios is not getting verified tag in UI.

The json index files (psp_serial_index.json, known_bios_files.json etc.) are updated with RomM updates, but the redis cache has a ttl of never and thus never updated. This makes RomM keep using outdated data if Redis is not reconstructed or cleared manually after application update.

**To Reproduce**

1. Use an old version of RomM
2. Let RomM fully start to fill in the caches
3. Upgrade to a newer version of RomM where any of the said files are updated ("known_bios_files.json")
4. Upload good frimware and see frimware not correctly get verification status (as it was newly added to json)

**Expected behavior**
The bios files should have correct verfification status.

**Additional context**
The json map caches should be recreated when RomM version is changed (upgrade/downgrade). Maybe keep RomM version in Redis and forcefully recache json files if version does not match with current RomM during startup.

adamantike:
**Description**
The cache for fixture files was being set without a TTL, which meant that updates to the fixture files were not reflected in the cache.

This change saves the MD5 hash of the fixture file in a different key, and compares it to the current hash of the file before deciding whether to update the cache.

Fixes #2347

**Checklist**

### DIFF
--- a/backend/tests/utils/test_cache.py
+++ b/backend/tests/utils/test_cache.py
@@ -27,10 +27,35 @@ class TestConditionallySetCache:
         mock_pipeline.hset.assert_called()
         mock_pipeline.execute.assert_called_once()
 
-    async def test_cache_exists_skips_loading(self, mocker):
-        """Test skipping load when cache already exists."""
+    async def test_cache_exists_and_hash_does_not_match_loads_data(self, mocker):
+        """Test loading data when cache exists but file hash does not match."""
         mock_cache_exists = mocker.patch.object(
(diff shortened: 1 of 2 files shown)


gantoine (approved):
nice solution 👌🏼

PR #2352, opened 2025-08-31
feat: Add scheduled task to sync RetroAchievements progress

adamantike:
**Description**
Add a new scheduled task that syncs RetroAchievements progress for all users with a RetroAchievements username.

Environment variables:
- `ENABLE_SCHEDULED_RETROACHIEVEMENTS_PROGRESS_SYNC`: Enable or disable the task (default: `false`)
- `SCHEDULED_RETROACHIEVEMENTS_PROGRESS_SYNC_CRON`: Cron string to schedule the task (default: "0 4 * * *" - daily at 4 AM)

**Checklist**


zurdi15 (approved):

backend/handler/database/users_handler.py
```diff
@@ -2,13 +2,40 @@
...
+                func.lower(User.username).in_([u.lower() for u in usernames])
+            )
+        if emails:
+            query = query.filter(
+                func.lower(User.email).in_([e.lower() for e in emails])
+            )
+        if roles:
+            query = query.filter(User.role.in_(roles))
+        if has_ra_username is not None:
+            predicate = and_(User.ra_username != "", User.ra_username.isnot(None))
```
gantoine, line 33: these filters are exclusive right? so if i pass in usernames and emails, it'll only match users in both groups? Not that we'd ever use it that way but a) it's good to understand and b) might want a comment if that's the case.

gantoine (approved):
very nice 💯

backend/handler/database/users_handler.py
adamantike: Yes, at the moment all filters are chained using `AND`. If we need to add more flexibility in the future, we can include an `operator` kwarg which defaults to `and_`, but could be set to `or_`.

PR #2357, opened 2025-09-01
misc: Add PlayBtn component to unify button logic

adamantike:
**Description**
The Play button logic was duplicated in three different places, including when to run either EmulatorJS or Ruffle.

This change centralizes that logic in a new `PlayBtn` component, which internally decides which player to use, and binds to any arguments passed to it.

Now all three places also consider platform versions when checking for emulation support.

**Checklist**

### DIFF
--- a/frontend/src/components/common/Game/Card/ActionBar.vue
+++ b/frontend/src/components/common/Game/Card/ActionBar.vue
@@ -1,44 +1,35 @@
 <script setup lang="ts">
 import type { Emitter } from "mitt";
-import { storeToRefs } from "pinia";
 import { computed, inject, ref, watch } from "vue";
 import { useI18n } from "vue-i18n";
 import AdminMenu from "@/components/common/Game/AdminMenu.vue";
 import PlayBtn from "@/components/common/Game/PlayBtn.vue";
 import romApi from "@/services/api/rom";
 import storeAuth from "@/stores/auth";
-import storeConfig from "@/stores/config";
 import storeDownload from "@/stores/download";
-import storeHeartbeat from "@/stores/heartbeat";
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/Game/Card/ActionBar.vue
```diff
@@ -71,43 +62,15 @@ watch(menuOpen, (val) => {
...
-        "
-        icon="mdi-play"
-        variant="text"
-        rounded="0"
-        :aria-label="`Play ${rom.name}`"
-      />
-      <v-btn
-        v-if="ruffleEmulationSupported"
-        :disabled="rom.missing_from_fs"
+    <v-col v-if="isEmulationSupported" class="d-flex">
```
adamantike, line 56: I haven't found a better way to not display this `v-col` when empty (`play-btn` decides it must not render), than importing `isAnyEmulationSupported` and evaluating it here too.

zurdi15: You can send an event from to the parent component (from the PlayBtn.vue to ActionBar.vue in this case) and perform an action based on that event. We do that with game cards and galleries, with the defineEmits. This way you don't need to duplicate the isEmulationSupported logic:

parent component: https://github.com/rommapp/romm/blob/4d962ea02a1249fcdb2914fe2ed90aeb013847a1/frontend/src/views/Gallery/Platform.vue#L285
child component: https://github.com/rommapp/romm/blob/4d962ea02a1249fcdb2914fe2ed90aeb013847a1/frontend/src/components/common/Game/Card/Base.vue#L66

zurdi15 (changes requested):

gantoine (approved):

PR #2411, opened 2025-09-09
feat: Initial support for multi-value ROM filters

adamantike:
**Description**
This change converts the existing single-value ROM filter parameters to support multiple values. Users can now filter ROMs by multiple genres, franchises, collections, companies, age ratings, regions, and languages by repeating the respective query parameters in API requests.

At the moment, setting multiple values for a filter will return ROMs that match any of the provided values (logical OR). A future change can introduce parameters like `genres_all (boolean)` to allow filtering ROMs that match all specified values (logical AND).

The frontend has been updated to send single-value filters as arrays to maintain compatibility with existing UI components. At the moment, the UI does not support selecting multiple values, but this change lays the groundwork for it.

NOTE: Breaking API change, as the filter parameters have changed names.

**Checklist**


backend/handler/database/collections_handler.py
```diff
@@ -250,9 +250,32 @@ def get_smart_collection_roms(
...
+            return None
+
+        # Apply conversions
+        genres = convert_legacy_filter("genres", "selected_genre")
+        franchises = convert_legacy_filter("franchises", "selected_franchise")
+        collections = convert_legacy_filter("collections", "selected_collection")
+        companies = convert_legacy_filter("companies", "selected_company")
+        age_ratings = convert_legacy_filter("age_ratings", "selected_age_rating")
+        regions = convert_legacy_filter("regions", "selected_region")
+        languages = convert_legacy_filter("languages", "selected_language")
```
gantoine, line 269: 👍🏼

backend/handler/database/roms_handler.py
```diff
@@ -298,6 +311,18 @@ def filter_by_playable(self, query: Query, value: bool) -> Query:
...
+        self, query: Query, value: bool, user_id: int | None = None
+    ) -> Query:
+        """Filter based on whether the rom has a last played value for the user."""
+        if not user_id:
+            return query
+
+        predicate = RomUser.last_played.isnot(None)
+        if not value:
+            predicate = RomUser.last_played.is_(None)
+        return query.filter(predicate)
```
gantoine: 👍🏼

frontend/src/services/api/index.ts
```diff
@@ -4,7 +4,32 @@ import { debounce } from "lodash";
...
+            if (item !== undefined && item !== null) {
+              searchParams.append(key, String(item));
+            }
+          });
+        } else if (value !== undefined && value !== null) {
+          searchParams.append(key, String(value));
+        }
+      });
+
+      return searchParams.toString();
```
gantoine, line 29: 💯

gantoine (commented):
very clean implementation and easy to review 🤘🏼

backend/handler/database/collections_handler.py
zurdi15: Actually, now self revieweing again the code I removed the old parameters. Should we remove the convert_legavy_filter or do you think it's better to re-add the old single-value parameters?

gantoine: good question, not sure but do what you think is best and we'll test in staging using exisiting smart collections.

zurdi15: ili'll leave itt here to not mess with already existing smart collections

PR #2510, opened 2025-10-03
Prevent OIDC claim role escalation

gantoine:
**Description**

> The OpenID token's roles claim (OIDC_CLAIM_ROLES) is treated as authoritative and the handler will unconditionally update the local user's role to the mapped Role value (including ADMIN) if it differs from the stored role. Because role updates happen during the single callback flow, an attacker who can present an OIDC token whose roles include the configured admin role will be promoted to ADMIN in a single request.

@MaienM Can you take a look at this?

**Checklist**

### DIFF
--- a/backend/handler/auth/base_handler.py
+++ b/backend/handler/auth/base_handler.py
@@ -278,6 +278,16 @@ class OAuthHandler:
 
 
 class OpenIDHandler:
+    def _get_role_priority(self, role) -> int:
+        from models.user import Role
+
+        role_priorities = {
+            Role.ADMIN: 3,
+            Role.EDITOR: 2,
+            Role.VIEWER: 1,
+        }
+        return role_priorities.get(role, 0)
+
     async def get_current_active_user_from_openid_token(self, token: Any):
         from handler.database import db_user_handler
         from models.user import Role, User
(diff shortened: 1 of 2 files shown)


MaienM:
> an attacker who can present an OIDC token whose roles include the configured admin role

Can you explain what this potential attack looks like in more detail? As far as I can tell, this attacker needs to have a token that has been signed by the configured OIDC provider that contains the configured roles, right? This isn't just some field that a user can arbitrarily change or make up, this data comes from the OIDC provider.

If you can't trust the contents of this token then there are security issues even without role mapping as an attacker that can provide a token with an admin email address will then be logged into an admin account.

The roles from the OIDC provider overwriting the roles configured in Romm is in fact the entire point of my recent PR which introduces this, I want the roles configured in my OIDC provider to be authoritative. This change (if I understand it correctly) makes the role mapping entirely pointless IMHO, as you will need to still manually change the users roles in Romm.

gantoine:
@MaienM I've been trying zeropath's autofixer for "security vulnerabilities" and this was one of the PRs it generated. Given your comment I think we can close this PR as "not actually a security risk", seeing as trusting the OIDC's token is necessary for other actions anyways. Thanks for the review!

PR #2570, opened 2025-10-19
fix: type error when role claim is null

Tarow:
**Description**

In my example, using Authelia + LLDAP and `OIDC_CLAIM_ROLES=groups`, when a user is part of no group, the claim will be returned as `null`.
image

Which results in a type error in RomM:
```
File "/backend/endpoints/auth.py", line 256, in auth_openid
await oidc_handler.get_current_active_user_from_openid_token(token)
File "/backend/handler/auth/base_handler.py", line 336, in get_current_active_user_from_openid_token
if OIDC_ROLE_ADMIN and OIDC_ROLE_ADMIN in roles:
                                    ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
 ```

**Checklist**

### DIFF
--- a/backend/handler/auth/base_handler.py
+++ b/backend/handler/auth/base_handler.py
@@ -332,7 +332,7 @@ class OpenIDHandler:
 
         role = Role.VIEWER
         if OIDC_CLAIM_ROLES and OIDC_CLAIM_ROLES in userinfo:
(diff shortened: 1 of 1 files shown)


gantoine (approved):
👍🏼

PR #2602, opened 2025-10-27
Feat: Add SEVEN_ZIP_TIMEOUT

JamieeBond:
**Description**
<sup>Add the ability to change the hardcoded timeout for 7z via a env variable, on big archives 60 seconds is not enough on when archiving on low performance devices like a NAS. The default value still remains at 60, but this merge request allows it to be overridden.</sup>

**Checklist**


JamieeBond:
@gantoine It looks like i have no permission to assignee a reviewer, so I've tagged you

PR #2660, opened 2025-11-15
Improve HLTB API handling by attempting to automatically fetch the current endpoint

death-droid:
**Description**
HLTB seems to actively change their API endpoint breaking the scraping functionality. This change attempts to make this experience smoother by automatically grabbing the script and working out what should be the correct search endpoint.

Additionally, a change was made so that  HLTB will use the roms defined "Name" instead of the filename. In my testing, this yielded a higher hit rate.

**Checklist**

- [X ] I've tested the changes locally

### DIFF
--- a/backend/handler/metadata/hltb_handler.py
+++ b/backend/handler/metadata/hltb_handler.py
@@ -1,6 +1,7 @@
 import json
 import re
 from typing import Final, NotRequired, TypedDict
+import hashlib
 
 import httpx
 from fastapi import HTTPException, status
@@ -224,7 +225,7 @@ class HLTBHandler(MetadataHandler):
             log.warning("Unexpected error discovering HLTB endpoint from site: %s", e)
 
 
-    def fetch_hltb_app_script(self) -> str:
+    def fetch_hltb_app_script(self) -> str | None:
         """Fetch the HLTB app script from the site."""
         try:
             with httpx.Client() as client:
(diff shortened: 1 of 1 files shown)


gantoine:
interesting approach, but this seems much more complicated then just updating the URL on github when it changes. and if the `*_app.js` file changes or is renamed, we'll need a deploy to fix it. thoughts?

il'l trigger an alert in discord when the URL changes, on top of the email i already send myself, so i can fix this sooner.

death-droid:
More complicated but less reliant on the Github repository remaining correct, especially with how often HLTB changes that API endpoint. The "_app.js" should remain somewhat stable due to the way NextJS works.

My alternate idea was to write a GitHub Action to automatically update a file with the latest API endpoint. Or to utilize a Cloudflare worker to provide a dynamic web location that tracks the latest API endpoint. Both might be good alternatives if you want to avoid having to re-deploy the container if there are ever any drastic changes.

gantoine:
@death-droid love the idea of a github action, what's your take on this? https://github.com/rommapp/romm/pull/2683/files
still needs to be triggered manually, but anyone on the team can do that even when AFK

death-droid:
> @death-droid love the idea of a github action, what's your take on this? https://github.com/rommapp/romm/pull/2683/files still needs to be triggered manually, but anyone on the team can do that even when AFK

Looks good to me and seems like a solution that should be easier to maintain down the track.

PR #2700, opened 2025-11-20
feat: Add metadata locale selection for IGDB and ScreenScraper

tvdu29:
**Description**

Having the possibility to pull metadata in my language and proper jacket for my version of the games is something I really want :)

This PR adds the possibility to choose a locale for compatible metadata sources. It is a bit crude for now but I find it to be a nice first step.

- Add provider-specific locale configuration for IGDB and ScreenScraper
- IGDB: Support for ja-JP, ko-KR, zh-CN, zh-TW, and EU regions
- ScreenScraper: Support for en, fr, de, es, it, pt languages
- Settings UI with locale dropdowns in Metadata Sources section
- Fallback warnings when requested locale is not available

> This PR was written primarily by Claude Code.

**Checklist**

#### Screenshots
{0DC08048-F618-47E4-84FC-CC0E002FEA2D}
{190C5F0E-6369-4F36-BEC4-1D2337E5EE6D}
{990C93E6-A83C-4830-93BB-C4D4FF4F38B1}

### DIFF
--- a/backend/handler/metadata/igdb_handler.py
+++ b/backend/handler/metadata/igdb_handler.py
@@ -213,13 +213,13 @@ def extract_metadata_from_igdb_rom(self: MetadataHandler, rom: Game) -> IGDBMeta
 # Mapping from scan.priority.region codes to IGDB game_localizations region identifiers
 # IGDB's game_localizations provides regional titles and cover art, but NOT localized descriptions
 REGION_TO_IGDB_LOCALE: dict[str, str | None] = {
-    "us": None,        # United States - use default (no localization needed)
-    "wor": None,       # World - use default
-    "eu": "EU",        # Europe region
-    "jp": "ja-JP",     # Japan
-    "kr": "ko-KR",     # Korea
-    "cn": "zh-CN",     # China (Simplified Chinese)
-    "tw": "zh-TW",     # Taiwan (Traditional Chinese)
+    "us": None,  # United States - use default (no localization needed)
+    "wor": None,  # World - use default
+    "eu": "EU",  # Europe region
+    "jp": "ja-JP",  # Japan
+    "kr": "ko-KR",  # Korea
+    "cn": "zh-CN",  # China (Simplified Chinese)
+    "tw": "zh-TW",  # Taiwan (Traditional Chinese)
 }
 
 
@@ -240,7 +240,7 @@ def get_igdb_preferred_locale() -> str | None:
         if igdb_locale is not None:
             return igdb_locale
 
-    return None  # Default - no localization
+    return None
 
 
 def extract_localized_data(rom: Game, preferred_locale: str | None) -> tuple[str, str]:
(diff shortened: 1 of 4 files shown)


gantoine:
excited to try this! can you post an example of the value of `provider_locales` in `config.yml`/

tvdu29:
> excited to try this! can you post an example of the value of `provider_locales` in `config.yml`/

Hey! Here's an example of how `provider_locales` looks in `config.yml`:

```yaml
scan:
  provider_locales:
    igdb: "ja-JP"
    ss: "fr"
```

**Currently supported locales:**

**IGDB** (region codes - there may be more available, default is US version):
- `ja-JP` - Japanese
- `ko-KR` - Korean  
- `zh-CN` - Chinese Simplified
- `zh-TW` - Chinese Traditional
- `EU` - Europe

**ScreenScraper** (language codes):
- `en` - English
- `fr` - Français
- `de` - Deutsch
- `es` - Español
- `it` - Italiano
- `pt` - Português

gantoine:
right so we already support languages under `scan.priority.region`, if you'd like i can take what you built here and merge the two, mapping to the IGDB ones. how does that sound?

tvdu29:
> right so we already support languages under `scan.priority.region`, if you'd like i can take what you built here and merge the two, mapping to the IGDB ones. how does that sound?

Thanks for the offer! I have to admit I was pretty confused about the distinction between region and locale while building this, so apologies if I've mixed things up.

Here's what I understand now about how they're currently used:

**ScreenScraper:**
- `scan.priority.region` selects which regional game name and artwork to prefer (e.g., "Resident Evil" with US box art vs "Biohazard" with JP box art)
- `provider_locales.ss` selects which language for synopsis/description (e.g., French vs English)
- Both filter client-side after receiving all data from the API

**IGDB:**
- `provider_locales.igdb` selects from `game_localizations` API field, which provides region-specific localized names and covers (e.g., "ja-JP" → Japanese title and cover)
- Also client-side filtering after receiving all localizations

**The difference:**
For ScreenScraper, region and language serve different purposes (name/artwork vs description language). For IGDB, they're combined in `game_localizations`.

If you think there's a cleaner way to structure this by mapping `scan.priority.region` to IGDB's region identifiers, I'm definitely open to it! Just want to make sure we preserve the ability to set ScreenScraper's language separately from its region preference.

gantoine:
> If you think there's a cleaner way to structure this by mapping scan.priority.region to IGDB's region identifiers, I'm definitely open to it!

I think you nailed it here, we co-opt `scan.priority.region` (or `scan.priority.lang`) and map it to IGDB's expected locales. you can pick either one, and lmk if you want a hand with the implementation.

tvdu29:
You're right - I've reimplemented IGDB to use `scan.priority.region` as you suggested. 

I realize now that ScreenScraper was already using `scan.priority.language` - I should have checked master more carefully before adding the provider-specific override. My apologies for that oversight!

### Changes

**IGDB Handler:**
- Now uses `scan.priority.region` to determine locale (maps `jp` → `ja-JP`, `eu` → `EU`, etc.)
- Fetches localized game titles and cover art from IGDB's `game_localizations` API
- Automatically falls back to default when locale unavailable

**ScreenScraper Handler:**
- Added locale fallback warning for better observability

**Configuration:**
```yaml
scan:
  priority:
    region: ["jp", "eu", "us"]  # Used by IGDB for regional variants
    language: ["fr", "en"]       # Used by ScreenScraper for descriptions
```

PR #2766, opened 2025-12-14
[ROMM-2697] Use colocated es-de folders to importart not present in gamelist.xml

### LINKED ISSUE #2697: [Bug] Scraper for ES-DE gamelist.xml does not properly scrape artwork
**RomM version**
4.4.0 container

**Describe the bug**
With 4.4.0 RomM introduced the ability to scrape the `gamelist.xml` file created by ES-DE. The data RomM expects isn't the behavior that ES-DE actually provides. ES-DE has moved the `gamelist.xml` out of the systems directories by default and hides the option to place them there under the `LegacyGamelistFileLocation` boolean in the `es_settings.xml`. Even if you switch this option to true and place all the artwork in the systems directory (which can be done by setting `MediaDirectory` to the same directory as the roms), it will not pick up the artwork. It seems that RomM's scraper expects the artwork location data to be stored in the `gamelist.xml`, but according to ES-DE devs this was never the case and instead ES-DE stores the artwork in `MediaDirectory` and references the artwork by filename.

In addition, all the directories such as miximages, covers, fanart, etc... that ES-DE created for its art were scanned as rom files in RomM. Perhaps when using `gamelist.xml` as a source it should ignore these directories that ES-DE creates, or maybe a note in the docs could be provided on how to setup RomM to ignore them? I'm assuming in the config.yaml:

```yaml
exclude:
  roms:
    multi_file:
      names:
      - 3dboxes
      - backcovers
      - covers
      - fanart
      - manuals
      - marquees
      - miximages
      - physicalmedia
      - screenshots
      - titlescreens
      - videos
    single_file:
      names:
      - systeminfo.txt
      - gamelist.xml
```

**To Reproduce**

1. Install ES-DE
2. Populate a rom directory with some roms
3. Have ES-DE scrape those roms
4. Check the resulting `gamelist.xml` and observe that no artwork information is stored
5. Check the `ES-DE/downloaded_media` directory and observe that all files are sorted by system, type, and filename.

**Expected behavior**
A way to solve this would be to change the expectation from the path metadata being in the XML, to perhaps providing a variable to point RomM to the correct `MediaDirectory` and RomM would walk the directory and match by filenames.

I don't think this is relevant, but just in case

- OS: Fedora 42 derivative
- Browser: Firefox
- Version: 145

**Additional context**
Tested using ES-DE 3.4.0 AppImage installed with Gear Lever. Troubleshooting was done with the help of arcaneadsada on Discord for RomM also, ant and LeonSe on the ES-DE discord.

gantoine:
**Description**

This PR discovers and imports colocated ES-DE media folders when entries aren't present in `gamelist.xml`. It also comments out the entire `example.config.yml` so people stop copying it as-is and opening tickets...

Fixes #2697

**Checklist**

### DIFF
--- a/examples/config.example.yml
+++ b/examples/config.example.yml
@@ -64,12 +64,12 @@
 #   # [your custom platform folder name]: [RomM platform name]
 #   # In this example if you have a 'gc' folder, RomM will treat it like the 'ngc' folder
 #   platforms:
-#     - gc: ngc
-#     - ps1: psx
+#     gc: ngc
+#     ps1: psx
 
 #   # Asociate one platform to it's main version (IGDB only)
 #   versions:
-#     - naomi: arcade
+#     naomi: arcade
 
 # The folder name where your roms are located
 # filesystem:
--- a/backend/handler/metadata/gamelist_handler.py
+++ b/backend/handler/metadata/gamelist_handler.py
@@ -2,7 +2,7 @@ import glob
 import os
 import uuid
 from pathlib import Path
-from typing import NotRequired, TypedDict
+from typing import Final, NotRequired, TypedDict
 from xml.etree.ElementTree import Element  # trunk-ignore(bandit/B405)
 
 import pydash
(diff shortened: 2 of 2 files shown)


examples/config.example.yml
```diff
@@ -2,48 +2,74 @@
 # Rename this file to `config.yml`, copy it to a `config` folder, and mount that folder as per the docker-compose.example.yml
```
zurdi15, line 2: 👌

PR #2771, opened 2025-12-15
feat: Add fs_slug filter in get roms

phscezario:


zurdi15:
Can you detail what's this PR for and how can adding that to the get_roms endpoint help?

If you only make changes in the backend and not also the frontend the change will not be useful at least for official RomM frontend. Also, the naming for the parameter should be ``platform_fs_slug``, not only ``fs_slug`` because that can be missleading with the ``rom_fs_slug``

phscezario:
This change introduces support for using fs_slug as an additional filter.
The purpose of this modification is to enable filtering of custom platforms that do not exist in external search repositories. In these cases, the platform does not yet have an igdbId, which makes it impossible to properly import ROMs when using frontends such as Playnite.
By allowing filtering through fs_slug, users can still organize and access their ROM collections even when the platform is not recognized by IGDB or other databases. This ensures that custom platforms remain usable and that ROMs can be imported correctly despite the absence of an external identifier

gantoine:
> which makes it impossible to properly import ROMs when using frontends such as Playnite

I see what you're trying to do but I don't think this is the right approach. can you open an issue in the [playnite-plugin repo](https://github.com/rommapp/playnite-plugin/issues) instead? We can take care of this from the plugin's side of things.

phscezario:
Okay, I can open an issue there without any problems. I believe we need to make changes on both sides, hence this PR. I would be very grateful if you could help. Later, I'll take a look at the plugin's code and see what I can understand from their side, and if possible, open a PR there as well.

PR #2777, opened 2025-12-16
[ROMM-816] User settings cross browser/device

### LINKED ISSUE #816: [UI/UX] Make user settings cross-browser/device
Storing user settings in the database will allow users to have their settings shared across multiple browsers/devices

zurdi15:
**Description**

This PR makes the user settings cross browser/device storing them into the database

**Checklist**

Closes #816

### DIFF
--- a/backend/alembic/versions/0060_user_ui_settings.py
+++ b/backend/alembic/versions/0060_user_ui_settings.py
@@ -1,6 +1,6 @@
 """user_ui_settings
 
-Revision ID: 0058_user_ui_settings
+Revision ID: 0060_user_ui_settings
 Revises: 0059_rom_version_tag
 Create Date: 2025-12-16 21:02:52.394533
 
@@ -11,7 +11,7 @@ from alembic import op
 from sqlalchemy.dialects import postgresql
 
 # revision identifiers, used by Alembic.
-revision = "0058_user_ui_settings"
+revision = "0060_user_ui_settings"
 down_revision = "0059_rom_version_tag"
 branch_labels = None
 depends_on = None


backend/alembic/versions/0060_user_ui_settings.py
```diff
@@ -0,0 +1,37 @@
...
+    with op.batch_alter_table("users", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "ui_settings",
+                sa.JSON().with_variant(
+                    postgresql.JSONB(astext_type=sa.Text()), "postgresql"
+                ),
+                nullable=True,
+            )
+        )
```
gantoine, line 31: i was gonna write something about `settings` vs `ui_settings`, but seeing as this is the first time we add cross-browser settings to the DB it should be fairly easy to migrate if/when we want to refactor in the future.

frontend/src/composables/useUISettings.ts
```diff
@@ -0,0 +1,179 @@
...
+  };
+}
+
+export function useUISettings() {
+  // Return singleton instance
+  if (!uiSettingsInstance) {
+    uiSettingsInstance = createUISettings();
+  }
+  return uiSettingsInstance;
+}
```
gantoine, line 179: 👌🏼

backend/alembic/versions/0060_user_ui_settings.py
zurdi15: Yeah makes sense. As I saw it we are too far of making the ui settings too complicated. Refactor and improvements will happen in the future 👌

PR #2784, opened 2025-12-18
[ROMM-1371] Add library structure management on setup wizard

### LINKED ISSUE #1371: [Feature] For first time installs, pre-create the required directory structure
For first time installs to simplify setup, pre-create the entire directory structure for all supported platforms.

An option using a check-box could also be provided during initial setup to turn this function on (as not every new install would be "new").

zurdi15:
**Description**
Improved the setup wizard with an extra step to detect and manage the folder structure, making a better UX for the user on RomM setup

**Checklist**

image

Closes #1371

### DIFF
--- a/backend/tests/handler/filesystem/test_platforms_handler.py
+++ b/backend/tests/handler/filesystem/test_platforms_handler.py
@@ -4,7 +4,7 @@ from unittest.mock import patch
 import pytest
 
 from config.config_manager import LIBRARY_BASE_PATH, Config
-from handler.filesystem.platforms_handler import FSPlatformsHandler
+from handler.filesystem.platforms_handler import FSPlatformsHandler, LibraryStructure
 
 
 class TestFSPlatformsHandler:
@@ -325,7 +325,7 @@ class TestFSPlatformsHandler:
                 mock_exists.return_value = True
 
                 result = handler.detect_library_structure()
-                assert result == "struct_a"
+                assert result == LibraryStructure.A
                 mock_exists.assert_called_once_with(roms_path)
 
     def test_detect_library_structure_structure_b(
@@ -358,7 +358,7 @@ class TestFSPlatformsHandler:
                         mock_exists.side_effect = exists_side_effect
 
                         result = handler.detect_library_structure()
-                        assert result == "B"
+                        assert result == LibraryStructure.B
 
     def test_detect_library_structure_none(self, handler: FSPlatformsHandler, config):
         """Test detect_library_structure returns None when no structure detected"""


backend/endpoints/heartbeat.py
```diff
@@ -155,3 +164,176 @@ async def metadata_heartbeat(source: str) -> bool:
...
+
+    # Check authentication - only allow public access if no admin users
+    # If admin users exist, this would need authentication (but won't be called during setup)
+
+    # Auto-detect structure type
+    # Structure A: /library/roms/{platform}
+    # Structure B: /library/{platform}/roms
+    # If there are admin users already, enforce the USERS_WRITE scope.
+    if (
+        Scope.PLATFORMS_READ not in request.auth.scopes
```
gantoine, line 193: `request.auth.scopes` is the scopes the user has access to?

backend/endpoints/heartbeat.py
```diff
@@ -155,3 +164,176 @@ async def metadata_heartbeat(source: str) -> bool:
...
+
+    try:
+        # Detect structure type to determine if we need to create the roms folder
+        detected_structure = fs_platform_handler.detect_library_structure()
+
+        # If no structure detected, create structure A (roms folder)
+        if detected_structure is None:
+            cnfg = cm.get_config()
+            roms_path = os.path.join(LIBRARY_BASE_PATH, cnfg.ROMS_FOLDER_NAME)
+            os.makedirs(roms_path, exist_ok=True)
```
gantoine: i'm fine keeping this folder creation login in here for now since its a 1-time run, but in the future we might move this into the filesystem handler

backend/handler/filesystem/platforms_handler.py
```diff
@@ -22,17 +22,60 @@ def _exclude_platforms(self, platforms: list):
...
+            "A" for Structure A (roms/{platform})
+            "B" for Structure B ({platform}/roms)
+            None if no structure detected
+        """
+        cnfg = cm.get_config()
+
+        # Check if the roms folder exists (Structure A indicator)
+        roms_path = os.path.join(LIBRARY_BASE_PATH, cnfg.ROMS_FOLDER_NAME)
+        if os.path.exists(roms_path):
+            return "A"
```
gantoine: these should probably return an enum, or something like it:
```
class LibraryStructure(enum.StrEnum):
  A = "struct_a"
  B = "struct_b"
```

backend/handler/filesystem/platforms_handler.py
```diff
@@ -22,17 +22,60 @@ def _exclude_platforms(self, platforms: list):
...
+        structure = self.detect_library_structure()
+        if structure == "A":
+            return cnfg.ROMS_FOLDER_NAME
+        if structure == "B":
+            return ""
+
+        # Fallback to config hint when detection is inconclusive
         return (
             cnfg.ROMS_FOLDER_NAME
             if os.path.exists(cnfg.HIGH_PRIO_STRUCTURE_PATH)
```
gantoine, line 65: are these checks still necessary?

backend/utils/platforms.py
```diff
@@ -0,0 +1,105 @@
...
+                "url_logo": igdb_platform.get("url_logo")
+                or tgdb_platform.get("url_logo")
+                or "",
+            }
+        )
+
+        platform = Platform(**platform_attrs)
+        supported_platforms.append(PlatformSchema.model_validate(platform))
+
+    return supported_platforms
```
gantoine, line 105: nice 👍🏼

frontend/src/services/api/setup.ts
```diff
@@ -0,0 +1,34 @@
+import api from "@/services/api";
+import type { Platform } from "@/stores/platforms";
+
+export type LibraryStructure = "A" | "B" | null;
```
gantoine: if/when you switch to an enum you _might_ be able to get it in the frontend

gantoine (changes requested):

backend/endpoints/heartbeat.py
zurdi15: Yeah it's the same [logic](https://github.com/rommapp/romm/blob/d76cd3412b05233d00e0cdc77d4f6b37c093e947/backend/endpoints/user.py#L61) we have to create the first admin user

backend/endpoints/heartbeat.py
zurdi15: You are right, that is cleaner. I'm moving it

frontend/src/services/api/setup.ts
zurdi15: I changes them to ``struct_x`` because I think it's more clear but we need to keep here as a type because we wil have a lot lof issues with enums and null in js

backend/handler/filesystem/platforms_handler.py
zurdi15: removed

PR #2817, opened 2025-12-30
fix: Notes flag

zurdi15:
**Description**
Re-added notes flag in gallery and migrated ``notes dialog`` to ``multi notes manager``

**Checklist**

### DIFF
--- a/frontend/src/__generated__/models/SimpleRomSchema.ts
+++ b/frontend/src/__generated__/models/SimpleRomSchema.ts
@@ -81,6 +81,6 @@ export type SimpleRomSchema = {
     rom_user: RomUserSchema;
     merged_screenshots: Array<string>;
     merged_ra_metadata: (RomRAMetadata | null);
-    has_notes: boolean;
+    has_notes?: boolean;
 };
 


backend/endpoints/responses/rom.py
```diff
@@ -329,15 +329,27 @@ def sort_comparator(self) -> str:
...
     def from_orm_with_request(cls, db_rom: Rom, request: Request) -> SimpleRomSchema:
+        user_id = request.user.id
         db_rom = cls.populate_properties(db_rom, request)
+
+        # Calculate has_notes from already-loaded notes relationship
+        # Check if there are any notes for this ROM (public or user's own)
+        notes = getattr(db_rom, "notes", [])
+        db_rom.has_notes = any(  # type: ignore
+            note.is_public or note.user_id == user_id for note in notes
+        )
```
gantoine: would this be cleaner as an `@property` on the model instead?

```
@property
def has_notes(self):
  return any(note.is_public or note.user_id == user_id for note in self.notes)
```

gantoine (changes requested):

backend/endpoints/responses/rom.py
zurdi15: Yeah good one. I was having issues with that because of the lazy load, but adding ``selectinload(Rom.notes)`` (check the last commit) to the database handler when fetching roms did the trick

zurdi15: Nevermind, I reverted the change because I needed the user context to check notes visibility. It's not as easy as changing it to be a property of the model it seems 😕

gantoine (approved):

PR #2845, opened 2026-01-02
[ROMM-2844] Use webp images for game covers in list view

### LINKED ISSUE #2844: [Bug] List view always requests PNG even when WebP is available and downloaded
**RomM version**
4.5.0

**Describe the bug**
The box art views (both small and large) seem to request the best file format available; If you've done the WebP conversion, this is WebP; but regardless if you've done this and then switch to the list view, the original version will always try to be requested. 

This became very apparent when I did the WebP conversion and then purged all the original files. This broke the list view images.

**To Reproduce**

1. Do the webP conversion task
2. Remove your original images
3. go to either box view then go to list view
4. See the images are missing in list view

Alternatively:
1. Do the webP conversion task
2. Open up your network dev tools in chrome
3. Go to the cover art view (small or large); notice that the small and big images are downloaded in webp
4. Then go to list view and observe how it redownloads the 'small' PNG 

**Expected behavior**
IF WebP images are available, I'd expect the list view to either download or use the 'small' WebP file.

**Screenshots**
Image
Image
Image

- OS: MacOS 26.2 (25C56)
- Browser: Chrome 143.0.7499.170

gantoine:
**Description**

Fixes #2844 

**Checklist**

### DIFF
--- a/frontend/src/components/common/Game/Card/Base.vue
+++ b/frontend/src/components/common/Game/Card/Base.vue
@@ -11,6 +11,7 @@ import {
   useTemplateRef,
 } from "vue";
 import { useDisplay } from "vuetify";
+import type { VImg } from "vuetify/lib/components/VImg/VImg.js";
 import type { BoxartStyleOption } from "@/components/Settings/UserInterface/Interface.vue";
 import ActionBar from "@/components/common/Game/Card/ActionBar.vue";
 import Flags from "@/components/common/Game/Card/Flags.vue";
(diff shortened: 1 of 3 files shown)


zurdi15:
What if the scheduled task is not enabled but ran at least once manually?

Ronald-Diemicke:
> What if the scheduled task is not enabled but ran at least once manually?

IMHO - that might be a different issue entirely because I noticed I had to enable the task - without enabling it, I couldn't run it manually nor would it run scheduled.

I'd argue managing this via a switch is ~~bad~~ sub-optimal overall and 'yet another' reason that the conversion should almost be _forced_ (like I was mentioning in discord) so that you can predictably use the best format instead of making this via configuration, it could be via convention.

PR #2872, opened 2026-01-05
feat: Added player count metadata from SS with display and filter

DevOldSchool:
Added player count metadata from SS. Displays on game detail screen and added a filter under search for player counts.

**Description**
> **AI Assistance Notice**
> JetBrains Junie free version was used to understand the code base and implement the initial change based on my analysis of the metadata from SS.
>
> I reviewed all changes and rewrote most of them to follow the existing code style.

This change brings in the player count (joueurs) field from the SS endpoint. This data was already available but was not being saved in the ss_metadata field.

The player count field is displayed on the game detail page as "Players" with a clickable filter for the value.

An additional search filter was added for player count to filter games by the selected values.

**Note; the player count is defaulted to 1 when no data exists. (An existing library will need a scan to update this value)**

**Checklist**

Detail screen:
image

Search filter:
image

image

**Questions for reviewers**

- Do I need to add best effort translations for other locales or add the English value in other locales or leave them empty?

### DIFF
--- a/backend/alembic/versions/0063_roms_metadata_player_count.py
+++ b/backend/alembic/versions/0063_roms_metadata_player_count.py
@@ -102,6 +102,7 @@ def upgrade():
                         NULLIF(r.manual_metadata ->> 'player_count', '1'),
                         NULLIF(r.ss_metadata ->> 'player_count', '1'),
                         NULLIF(r.igdb_metadata ->> 'player_count', '1'),
+                        NULLIF(r.gamelist_metadata ->> 'player_count', '1'),
                         '1'
                     ) AS player_count,
 
@@ -286,6 +287,7 @@ def upgrade():
                             NULLIF(JSON_UNQUOTE(JSON_EXTRACT(r.manual_metadata, '$.player_count')), '1'),
                             NULLIF(JSON_UNQUOTE(JSON_EXTRACT(r.ss_metadata, '$.player_count')), '1'),
                             NULLIF(JSON_UNQUOTE(JSON_EXTRACT(r.igdb_metadata, '$.player_count')), '1'),
+                            NULLIF(JSON_UNQUOTE(JSON_EXTRACT(r.gamelist_metadata, '$.player_count')), '1'),
                             '1'
                         ) AS player_count,
 
--- a/backend/handler/metadata/igdb_handler.py
+++ b/backend/handler/metadata/igdb_handler.py
@@ -177,15 +177,11 @@ def extract_metadata_from_igdb_rom(
     for mm in multiplayer_modes:
         platform_data = mm.get("platform")
 
+        igdb_id = -1
+        name = ""
         if isinstance(platform_data, dict):
-            igdb_id = platform_data.get("id", 0)
+            igdb_id = platform_data.get("id", -1)
             name = platform_data.get("name", "")
-        elif isinstance(platform_data, int):
-            igdb_id = platform_data
-            name = ""
-        else:
-            igdb_id = 0
-            name = ""
 
         multiplayer_modes_metadata.append(
             IGDBMetadataMultiplayerMode(


backend/handler/database/roms_handler.py
```diff
@@ -462,6 +462,16 @@ def filter_by_languages(
         op = json_array_contains_all if match_all else json_array_contains_any
         return query.filter(op(Rom.languages, values, session=session))
 
+    def filter_by_player_counts(
+            self,
+            query: Query,
+            *,
+            session: Session,
+            values: Sequence[str],
+            match_all: bool = False,
```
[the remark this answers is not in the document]
DevOldSchool: Ignoring this as line 692 passes in the parameter so it's expected to be there.

```
query, session=session, values=values, match_all=(logic == "all")
```

backend/handler/database/roms_handler.py
```diff
@@ -462,6 +462,16 @@ def filter_by_languages(
...
 
+    def filter_by_player_counts(
+        self,
+        query: Query,
+        *,
+        session: Session,
+        values: Sequence[str],
+        match_all: bool = False,
+    ) -> Query:
+        return query.filter(RomMetadata.player_count.in_(values))
```
gantoine, line 473: was the choice of adding `player_count` to the `roms_metadata` db view taken to use it here as a filter? was it not possible with `ss_metadata` column?

backend/alembic/versions/0063_roms_metadata_player_count.py
```diff
@@ -0,0 +1,390 @@
...
+                                AND r.launchbox_metadata ->> 'esrb' != ''
+                            THEN
+                                jsonb_build_array(r.launchbox_metadata ->> 'esrb')
+                            ELSE
+                                '[]'::jsonb
+                        END,
+                        '[]'::jsonb
+                    ) AS age_ratings,
+
+                    COALESCE(r.ss_metadata ->> 'player_count', '1') AS player_count,
```
gantoine: for a single source, we don't add it to the metadata view and instead use it directly from `rom.ss_metadata.get('player_count', None)` or the TS equivalent

gantoine (changes requested):
good start! can you also integrate IGDB's multiplayer API (https://api-docs.igdb.com/#multiplayer-mode)? es-de also supports a `<players>` tag with an integer and should be included in `gamelist_handler.py`

backend/handler/database/roms_handler.py
DevOldSchool: I was just following other filters such as genre as a base for what to do. I'll try this out.

Edit, doesn't look like the ss_metadata column is returned.
Considering you would like to expand to multiple sources, should we keep this in the view so we can handle those as well?

backend/alembic/versions/0063_roms_metadata_player_count.py
gantoine: ignore this is it'll use data from multiple providers!

backend/handler/database/roms_handler.py
gantoine: > Considering you would like to expand to multiple sources, should we keep this in the view so we can handle those as well?

👍🏼

DevOldSchool:
> good start! can you also integrate IGDB's multiplayer API (https://api-docs.igdb.com/#multiplayer-mode)? es-de also supports a `<players>` tag with an integer and should be included in `gamelist_handler.py`

Looks like the gamelist_handler already correctly handles the <players> tag from gamelist.xml.

IGDB data doesn't seem very good.
Here is an example for GTA V where there is no player count for PC listed.

```json
[
  {
    "id": 134709,
    "name": "Grand Theft Auto V"
  },
  {
    "id": 1020,
    "multiplayer_modes": [
      {
        "id": 1045,
        "offlinecoop": false,
        "offlinemax": 0,
        "onlinecoop": true,
        "onlinemax": 0,
        "platform": {
          "id": 9,
          "name": "PlayStation 3"
        }
      },
      {
        "id": 12794,
        "offlinecoop": false,
        "offlinemax": 0,
        "onlinecoop": true,
        "onlinemax": 30,
        "platform": {
          "id": 49,
          "name": "Xbox One"
        }
      },
      {
        "id": 12795,
        "offlinecoop": false,
        "onlinecoop": true,
        "platform": {
          "id": 6,
          "name": "PC (Microsoft Windows)"
        }
      }
    ],
    "name": "Grand Theft Auto V"
  },
  {
    "id": 239064,
    "multiplayer_modes": [
      {
        "id": 24400,
        "offlinecoop": false,
        "onlinecoop": true,
        "onlinemax": 30,
        "platform": {
          "id": 49,
          "name": "Xbox One"
        }
      },
      {
        "id": 24401,
        "offlinecoop": false,
        "onlinecoop": true,
        "onlinemax": 30,
        "platform": {
          "id": 169,
          "name": "Xbox Series X|S"
        }
      }
    ],
    "name": "Grand Theft Auto V"
  }
]
```

It also introduces online/offline multiplayer concepts which I had not considered as my library is retro consoles.

How should we handle this?

gantoine:
> How should we handle this?

Map `offlinecoop: true` and `onlinecoop: true` to "2 players", `offlinemax` of 3 to 3 players, and `offlinemax` > 4 to 4 player. does that make sense? it's not perfect but should be close enough for our needs.

On that topic, seems mobygames supports "Number of Players Supported" via the API, but we've more or less deprecated it so don't worry about supporting it at the moment.

DevOldSchool:
> > How should we handle this?
> 
> Map `offlinecoop: true` and `onlinecoop: true` to "2 players", `offlinemax` of 3 to 3 players, and `offlinemax` > 4 to 4 player. does that make sense? it's not perfect but should be close enough for our needs.
> 
> On that topic, seems mobygames supports "Number of Players Supported" via the API, but we've more or less deprecated it so don't worry about supporting it at the moment.

Ended up doing this except for the > 4 logic. I just let it use the max value.

DevOldSchool:
@gantoine, can you review again?

gantoine:
@DevOldSchool ill have a look tomorrow, great work on this PR!

backend/handler/metadata/igdb_handler.py
```diff
@@ -146,12 +165,48 @@ def extract_metadata_from_igdb_rom(self: MetadataHandler, rom: Game) -> IGDBMeta
...
+
+        if isinstance(platform_data, dict):
+            igdb_id = platform_data.get("id", 0)
+            name = platform_data.get("name", "")
+        elif isinstance(platform_data, int):
+            igdb_id = platform_data
+            name = ""
+        else:
+            igdb_id = 0
+            name = ""
```
gantoine: a comment here explain why platform_data could be a dict or an int would be useful (if you know). also this might be able to be simplified as:

```
        igdb_id = platform_data or -1
        name = ""
        if isinstance(platform_data, dict):
            igdb_id = platform_data.get("id", 0)
            name = platform_data.get("name", "")
```

backend/handler/metadata/igdb_handler.py
```diff
@@ -210,6 +269,53 @@ def extract_metadata_from_igdb_rom(self: MetadataHandler, rom: Game) -> IGDBMeta
...
+            mm.get("onlinecoopmax", 0),
+        )
+
+        max_players = max(
+            max_players,
+            mm.get("offlinemax", 0),
+            mm.get("onlinemax", 0),
+        )
+
+    return f"1-{max_players}" if max_players > 1 else "1"
```
gantoine, line 312: cleanly written 👍🏼

backend/alembic/versions/0063_roms_metadata_player_count.py
```diff
@@ -0,0 +1,400 @@
...
+                        END,
+                        '[]'::jsonb
+                    ) AS age_ratings,
+
+                    COALESCE(
+                        NULLIF(r.manual_metadata ->> 'player_count', '1'),
+                        NULLIF(r.ss_metadata ->> 'player_count', '1'),
+                        NULLIF(r.igdb_metadata ->> 'player_count', '1'),
+                        '1'
+                    ) AS player_count,
```
gantoine, line 107: only thing left here is player_count from `gamelist_metadata` (an for the other db type below)

backend/handler/metadata/igdb_handler.py
DevOldSchool: I'm not actually sure why platform_data can be. This was a type guard to resolve the trunk check issues:

backend/handler/metadata/igdb_handler.py:179:19
 179:19  high  Item "None" of "Platform | int | None" has no attribute "get"  mypy/union-attr
 179:19  high  Item "int" of "Platform | int | None" has no attribute "get"   mypy/union-attr
 180:16  high  Item "None" of "Platform | int | None" has no attribute "get"  mypy/union-attr
 180:16  high  Item "int" of "Platform | int | None" has no attribute "get"   mypy/union-attr

backend/alembic/versions/0063_roms_metadata_player_count.py
DevOldSchool: That's been added.

Issue #2874, opened 2026-01-05
[Feature] `updated_since` query parameter to Platform and Rom endpoints


BrandonKowalski:
**Is your feature request related to a problem? Please describe.**
This is a follow up to #2810 after spending more time with the problem.

Network calls are the bottleneck for Grout on cheap handhelds.

I've started implementing a basic JSON file cache as I explore the problem space. All signs are pointing to the need for a local SQLite database to store this data sanely.

The problem is keeping this cache in sync with RomM.

**Describe the solution you'd like**
Add a new query parameter `updated_since` (or another name that makes sense) to the Platform and Rom endpoints. Additionally, it looks like an index will need to be added to both the `platforms` and `roms` tables for the `updated_at` column.

**Describe alternatives you've considered**
I started with the solution proposed in #2810 to do a limit 1 query ordered by `updated_at` descending. This does help identify a stale cache. However, refreshing the cache with this scheme is an expensive operation as it requires an entire pull, even if only one thing changed.

**Additional context**
This won't solve everything. The first sync will still require pulling everything. I'm hoping for huge collections #2811 will help with the performance of a full pull. After the first pull these requested changes will bring efficiency. Additionally, I believe other client projects would benefit from this change.

The only gap with this solution is it does not assist with deletes. I'm guessing a logical delete isn't in the cards so a new endpoint that enumerates a current list of ids (platforms and roms) would be helpful.

PR #2898, opened 2026-01-13
[ROMM-2874] Feature/updated-after-query-param

BrandonKowalski:
**Description**
For #2874 

Adds `updated_after` as a query parameter to the `/api/platforms` and `/api/roms` endpoints.

Additionally, added a DB migration to add indexes to the `updated_at` field on the respective tables.

It has been a while since I've python-ed. I tried to follow existing examples (especially for the migration stuff as I've never used Alembic).

This will allow client applications to perform more efficient updates to their local caches by only pulling a delta.

**Checklist**

- [x] I've updated relevant comments (in the OpenAPI spec)

### DIFF
--- a/backend/endpoints/rom.py
+++ b/backend/endpoints/rom.py
@@ -407,7 +407,9 @@ def get_roms(
     ] = "asc",
     updated_after: Annotated[
         datetime | None,
-        Query(description="Filter roms updated after this datetime (ISO 8601 format with timezone information)."),
+        Query(
+            description="Filter roms updated after this datetime (ISO 8601 format with timezone information)."
+        ),
     ] = None,
 ) -> CustomLimitOffsetPage[SimpleRomSchema]:
     """Retrieve roms."""


backend/handler/database/platforms_handler.py
```diff
@@ -65,9 +66,12 @@ def get_platform(
     @with_firmware
     def get_platforms(
         self,
+        updated_after: datetime | None = None,
```
gantoine, line 69: thoughts on `updated_after` vs `updated_since`?

gantoine (changes requested):
try and run `npm run generete` to see if the openapi files change

backend/handler/database/platforms_handler.py
BrandonKowalski: I've been flipping back and forth on that. In the original title of the feature issue I used `updated_since`.

I'm fine with either, a quick Google search is a toss up on what people go with. If this gets switched to since I think the query will need to be switched to inclusive (either case the inclusivity / exclusivity needs to be added to the OpenAPI spec).

gantoine: let's leave it as is, long as you thought about it im good

PR #2917, opened 2026-01-18
Add device-based save synchronization

tmgast:
## Description

Adds device-based save synchronization to enable multi-device save management. Devices (handhelds, PCs, etc.) can register with the server and track which saves they've synced, enabling conflict detection when a device tries to upload stale data.

### Features

- **Device Registration**: Devices register once and receive a persistent UUID
- **Sync Tracking**: Each device tracks when it last synced each save file
- **Conflict Detection**: Uploading stale data returns 409 with details; bypass with `overwrite=true`
- **Track/Untrack**: Devices can opt out of syncing specific saves
- **Slot-Based Saves**: Optional `slot` parameter for categorizing saves into named slots
- **Datetime Tagging**: Filenames auto-tagged with `[YYYY-MM-DD_HH-MM-SS]` when slot provided
- **Autocleanup**: Optional `autocleanup` + `autocleanup_limit` to prune old slot saves
- **Save Summary**: Aggregated slot information via `/saves/summary`

> [!NOTE]
> This is a foundational PR to set up an initial structure and API implementation for devices and save syncing in preparation for client application usage and eventual alternative sync modes and use-cases.

### Known Gaps

- no registered device recovery on duplicate/repeat registration
- sync_mode defaults to API for all devices
- no batch update mechanic
- file-transfer and push-pull modes not handled (out of scope)
- features not exposed to the front-end

---

### Device

| Field            | Type        | Description                                          |
| ---------------- | ----------- | ---------------------------------------------------- |
| `id`             | `str` (PK)  | UUID assigned on registration                        |
| `user_id`        | `int` (FK)  | Owner of the device                                  |
| `name`           | `str?`      | User-friendly name (e.g., "Odin3", "Pocket Mini v2") |
| `platform`       | `str?`      | OS/platform (e.g., "android", "linux")               |
| `client`         | `str?`      | Client app name (e.g., "argosy", "grout")            |
| `client_version` | `str?`      | Client version                                       |
| `ip_address`     | `str?`      | Last known IP                                        |
| `mac_address`    | `str?`      | MAC address                                          |
| `hostname`       | `str?`      | Device hostname                                      |
| `sync_mode`      | `SyncMode`  | Enum: `api`, `file_transfer`, `push_pull`            |
| `sync_enabled`   | `bool`      | Whether sync is enabled (default: true)              |
| `last_seen`      | `datetime?` | Last activity timestamp                              |

### DeviceSaveSync

| Field            | Type           | Description                                   |
| ---------------- | -------------- | --------------------------------------------- |
| `device_id`      | `str` (PK, FK) | References `devices.id`                       |
| `save_id`        | `int` (PK, FK) | References `saves.id`                         |
| `last_synced_at` | `datetime`     | When device last synced this save             |
| `is_untracked`   | `bool`         | Device opted out of syncing this save         |
| `is_current`     | `bool`         | Computed: device has the current save version |

### DeviceSchema Sample

```json
{
  "id": "abc-123-uuid",
  "user_id": 1,
  "name": "Steam Deck",
  "platform": "linux",
  "client": "RetroArch",
  "client_version": "1.17.0",
  "ip_address": "192.168.1.100",
  "mac_address": "AA:BB:CC:DD:EE:FF",
  "hostname": "steamdeck",
  "sync_mode": "api",
  "sync_enabled": true,
  "last_seen": "2025-01-18T14:30:00Z",
  "created_at": "2025-01-15T10:00:00Z",
  "updated_at": "2025-01-18T14:30:00Z"
}
```

### SaveSchema Sample

```json
{
  "id": 42,
  "rom_id": 100,
  "user_id": 1,
  "file_name": "pokemon_emerald [2025-01-18_14-00-00].sav",
  "file_name_no_tags": "pokemon_emerald",
  "file_name_no_ext": "pokemon_emerald [2025-01-18_14-00-00]",
  "file_extension": "sav",
  "file_path": "/saves/gba/100",
  "file_size_bytes": 131072,
  "full_path": "/saves/gba/100/pokemon_emerald [2025-01-18_14-00-00].sav",
  "download_path": "/api/saves/42/content",
  "missing_from_fs": false,
  "created_at": "2025-01-10T08:00:00Z",
  "updated_at": "2025-01-18T14:00:00Z",
  "emulator": "mgba",
  "slot": "Main Playthrough",
  "screenshot": null,
  "device_syncs": [
    {
      "device_id": "abc-123-uuid",
      "device_name": "Steam Deck",
      "last_synced_at": "2025-01-17T10:00:00Z",
      "is_untracked": false,
      "is_current": false
    }
  ]
}
```

### SaveSummarySchema Sample

```json
{
  "total_count": 6,
  "slots": [
    {
      "slot": null,
      "count": 3,
      "latest": { "...SaveSchema..." }
    },
    {
      "slot": "Main Playthrough",
      "count": 2,
      "latest": { "...SaveSchema..." }
    },
    {
      "slot": "Nuzlocke Run",
      "count": 1,
      "latest": { "...SaveSchema..." }
    }
  ]
}
```

---

### Devices

**POST** `/api/devices`

- Register a new device

```json
REQUEST
{
  "name": "Steam Deck",
  "platform": "linux",
  "client": "RetroArch",
  "client_version": "1.17.0",
  "ip_address": "192.168.1.100",
  "mac_address": "AA:BB:CC:DD:EE:FF",
  "hostname": "steamdeck"
}

RESPONSE
{
  "device_id": "abc-123-uuid",
  "name": "Steam Deck",
  "created_at": "2025-01-18T12:00:00Z"
}
```

**GET** `/api/devices`

- List all devices for the current user

```json
RESPONSE
[DeviceSchema, ...]
```

**GET** `/api/devices/{device_id}`

- Get device details

```json
RESPONSE
DeviceSchema
```

**PUT** `/api/devices/{device_id}`

- Update device properties

```json
REQUEST
{
  "name": "Steam Deck (Docked)",
  "sync_enabled": true
}

RESPONSE
DeviceSchema
```

**DELETE** `/api/devices/{device_id}`

- Delete a device and its sync records

```
RESPONSE: 204 No Content
```

### Saves

**GET** `/api/saves/summary?rom_id={rom_id}`

- Get aggregated slot information for a ROM
- Returns total count, list of slots with counts, and latest save per slot

```json
RESPONSE
{
  "total_count": 6,
  "slots": [
    {
      "slot": null,
      "count": 3,
      "latest": SaveSchema
    },
    {
      "slot": "Main Playthrough",
      "count": 2,
      "latest": SaveSchema
    }
  ]
}
```

### Save Sync Operations

**POST** `/api/saves/{id}/track`

- Re-enable sync tracking for a save on this device

```json
REQUEST
{
  "device_id": "abc-123-uuid"
}

RESPONSE
SaveSchema
```

**POST** `/api/saves/{id}/untrack`

- Opt out of syncing this save on this device

```json
REQUEST
{
  "device_id": "abc-123-uuid"
}

RESPONSE
SaveSchema
```

**POST** `/api/saves/{id}/downloaded`

- Confirm download completed (for non-optimistic sync)

```json
REQUEST
{
  "device_id": "abc-123-uuid"
}

RESPONSE
SaveSchema
```

---

## Updated API Endpoints

**POST** `/api/saves`

- Added `device_id` query param to activate sync features
- Added `slot` query param for slot categorization (optional)
- Added `autocleanup` query param (default: false)
- Added `autocleanup_limit` query param (default: 10)
- When `slot` provided, filename auto-tagged with `[YYYY-MM-DD_HH-MM-SS]`
- Existing datetime tags in filename are replaced, not appended
- Returns 409 on conflict when device has stale sync

```json
409 RESPONSE (without slot - destructive conflict)
{
  "detail": {
    "error": "conflict",
    "message": "Save has been updated since last sync and will be overwritten",
    "save_id": 42,
    "current_save_time": "2025-01-18T14:00:00Z",
    "device_sync_time": "2025-01-17T10:00:00Z"
  }
}

409 RESPONSE (with slot - out of sync deviation)
{
  "detail": {
    "error": "out_of_sync",
    "message": "Save has been updated since last sync",
    "save_id": 42,
    "current_save_time": "2025-01-18T14:00:00Z",
    "device_sync_time": "2025-01-17T10:00:00Z"
  }
}
```

**GET** `/api/saves`

- Added `device_id` query param
- Added `slot` query param to filter by slot
- Returns `device_syncs[]` with sync status when `device_id` provided

**GET** `/api/saves/{id}`

- Added `device_id` query param
- Returns `device_syncs[]` with sync status when `device_id` provided

**GET** `/api/saves/{id}/content`

- Added `device_id` query param
- Added `optimistic` query param (default: true)
- When `optimistic=true`, updates sync record on download

---

## New Scopes

- **devices.read**: Read device information
- **devices.write**: Register/modify/delete devices

---

## Database Changes

- **Migration**: `0068_save_sync.py`

---

### Device Registration (First Launch)

User login -> register new device -> store device ID for save sync API calls

### Discover Save Slots (Game Selected)

Fetch save summary to populate slot picker UI

**GET** `/api/saves/summary?rom_id={rom_id}`

- Returns all slots with counts and latest save per slot
- Use to build slot selection UI:
  - "Main Playthrough" (3 saves, last: Jan 30)
  - "Nuzlocke Run" (1 save, last: Jan 15)
  - No slot (5 saves, last: Jan 28)

### Start Game Flow

**If loading specific slot:**

**GET** `/api/saves?rom_id={rom_id}&slot={slot_name}&device_id={device_id}`

- Get saves for selected slot only
- Check `device_syncs[0].is_current` on latest
  - `true` -> local save is current, safe to play
  - `false` -> download latest before playing

**GET** `/api/saves/{id}/content?device_id={device_id}`

- Downloads save file
- With `optimistic=true` (default), sync record updates automatically

**If loading latest across all slots:**

Use summary's `latest` from each slot to find most recent globally

### End Game Flow

Check sync status -> upload new save file

**GET** `/api/saves/{id}?device_id={device_id}`

- Check `device_syncs[0].is_current` before uploading
  - `true` -> safe to upload
  - `false` -> warn user: server has newer save

**POST** `/api/saves?rom_id={rom_id}&device_id={device_id}&slot={slot_name}`

- Upload save to specific slot
- Filename auto-tagged with datetime (e.g., `save [2025-01-18_14-30-00].sav`)
- Each upload creates new timestamped save in slot

**With autocleanup (recommended for autosave slots):**

**POST** `/api/saves?rom_id={rom_id}&device_id={device_id}&slot=autosave&autocleanup=true&autocleanup_limit=10`

- Keeps only 10 most recent saves in "autosave" slot
- Older saves automatically pruned

> [!NOTE]
> To force overwrite on conflict:
> **POST** `/api/saves?rom_id={rom_id}&device_id={device_id}&overwrite=true`

---

### Datetime Tagging

When uploading with a `slot` parameter:

- Filename is automatically tagged: `filename [YYYY-MM-DD_HH-MM-SS].ext`
- If filename already has a datetime tag, it is replaced (not double-appended)
- Saves without `slot` parameter are not tagged (backward compatible)

### Autocleanup

When uploading with `autocleanup=true`:

- Applies only when `slot` is also provided
- After upload, counts saves in that slot
- Deletes oldest saves beyond `autocleanup_limit` (default: 10)
- Useful for autosave slots to prevent unbounded growth

### Summary Endpoint

**GET** `/api/saves/summary?rom_id={rom_id}` returns:

- `total_count`: Total saves for this ROM
- `slots[]`: Array of slot summaries, each with:
  - `slot`: Slot name (or `null` for unslotted saves)
  - `count`: Number of saves in this slot
  - `latest`: Full SaveSchema of most recent save in slot

---

> AI Disclosure
> Planning and review assisted by Claude Code

## Checklist

### DIFF
--- a/backend/endpoints/device.py
+++ b/backend/endpoints/device.py
@@ -2,7 +2,7 @@ import uuid
 from datetime import datetime, timezone
 
 from fastapi import HTTPException, Request, Response, status
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 
 from decorators.auth import protected_route
 from endpoints.responses.device import DeviceCreateResponse, DeviceSchema
@@ -26,9 +26,15 @@ class DeviceCreatePayload(BaseModel):
     ip_address: str | None = None
     mac_address: str | None = None
     hostname: str | None = None
-    allow_existing: bool = False
+    allow_existing: bool = True
     allow_duplicate: bool = False
     reset_syncs: bool = False
+
+    @model_validator(mode="after")
+    def _duplicate_disables_existing(self) -> "DeviceCreatePayload":
+        if self.allow_duplicate:
+            self.allow_existing = False
+        return self
 
 
 class DeviceUpdatePayload(BaseModel):
--- a/backend/tests/endpoints/test_device.py
+++ b/backend/tests/endpoints/test_device.py
@@ -282,7 +282,7 @@ class TestDeviceUserIsolation:
 
 
 class TestDeviceDuplicateHandling:
-    def test_duplicate_mac_address_returns_409(
+    def test_duplicate_mac_address_returns_existing(
         self, client, access_token: str, admin_user: User
     ):
         db_device_handler.add_device(
@@ -303,12 +303,12 @@ class TestDeviceDuplicateHandling:
             headers={"Authorization": f"Bearer {access_token}"},
         )
 
-        assert response.status_code == status.HTTP_409_CONFLICT
-        data = response.json()["detail"]
-        assert data["error"] == "device_exists"
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()
         assert data["device_id"] == "existing-mac-device"
+        assert data["name"] == "Existing Device"
 
-    def test_duplicate_hostname_platform_returns_409(
+    def test_duplicate_hostname_platform_returns_existing(
         self, client, access_token: str, admin_user: User
     ):
         db_device_handler.add_device(
@@ -331,10 +331,37 @@ class TestDeviceDuplicateHandling:
             headers={"Authorization": f"Bearer {access_token}"},
         )
 
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()
+        assert data["device_id"] == "existing-hostname-device"
+        assert data["name"] == "Existing Device"
+
+    def test_duplicate_with_allow_existing_false_returns_409(
+        self, client, access_token: str, admin_user: User
+    ):
+        db_device_handler.add_device(
+            Device(
+                id="reject-duplicate-device",
+                user_id=admin_user.id,
+                name="Existing Device",
+                mac_address="FF:EE:DD:CC:BB:AA",
+            )
+        )
+
+        response = client.post(
+            "/api/devices",
+            json={
+                "name": "New Device",
+                "mac_address": "FF:EE:DD:CC:BB:AA",
+                "allow_existing": False,
+            },
+            headers={"Authorization": f"Bearer {access_token}"},
+        )
+
         assert response.status_code == status.HTTP_409_CONFLICT
         data = response.json()["detail"]
         assert data["error"] == "device_exists"
-        assert data["device_id"] == "existing-hostname-device"
+        assert data["device_id"] == "reject-duplicate-device"
 
     def test_allow_existing_returns_existing_device(
         self, client, access_token: str, admin_user: User


backend/alembic/versions/0068_save_sync.py
```diff
@@ -0,0 +1,96 @@
...
+        sa.Column("client_version", sa.String(50), nullable=True),
+        sa.Column("ip_address", sa.String(45), nullable=True),
+        sa.Column("mac_address", sa.String(17), nullable=True),
+        sa.Column("hostname", sa.String(255), nullable=True),
+        sa.Column(
+            "sync_mode",
+            sa.Enum("API", "FILE_TRANSFER", "PUSH_PULL", name="syncmode"),
+            nullable=False,
+            server_default="API",
+        ),
```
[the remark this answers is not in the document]
tmgast: fixed in 08ee684

backend/endpoints/saves.py
```diff
@@ -124,43 +201,174 @@ async def add_save(
...
+            device_id=device.id,
+            save_id=save.id,
+            synced_at=save.updated_at,
+        )
+        db_device_handler.update_last_seen(device_id=device.id, user_id=request.user.id)
+
+    return FileResponse(path=str(file_path), filename=save.file_name)
+
+
+@protected_route(router.post, "/{id}/downloaded", [Scope.ASSETS_READ])
```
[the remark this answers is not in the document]
tmgast: I'm conflicted on this as I think, realistically, it's a combination of device and asset permissions being required (why it was the way I had it, originally). I'm going to keep these comments open but I've applied to suggested fixes. We can revert them after some discussion if need be.

backend/endpoints/saves.py
```diff
@@ -2,19 +2,68 @@
...
+from models.assets import Save
+from models.device import Device
+from models.device_save_sync import DeviceSaveSync
 from utils.router import APIRouter
 
+
+def _normalize_datetime(dt: datetime) -> datetime:
+    if dt.tzinfo is not None:
+        return dt.replace(tzinfo=None)
+    return dt
```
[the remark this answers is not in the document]
tmgast: the input dates all should come from the db, so it should be a non-issue, but I've updated this to force utc anyway in f0b2158

BrandonKowalski:
I have been tinkering with a naive save sync in Grout for about a month now and look forward to implementing what you've shared here. Right now I just use timestamps but this scheme will offer much more control.

I think this would also be the right time to take a look at adding the concept of "playthroughs" (name WIP). Right now, saves are essentially just a big bucket tied to game/user. A little extra organization will go a long way and I think will help make this syncing system even more robust.

I just saw another user mentioned it a few months ago on the Sync RFC: https://github.com/rommapp/romm/discussions/2199#discussioncomment-14607030

Instead of registering saves we'd register playthroughs. Playthroughs could store a rolling list of saves (configurable maybe with a default of 5 or so) and will always return the metadata for the most recent save file associated with it. This would allow for the ability to rollback to be added in the future. I realize while this is technically achievable already with the single bucket of saves but I think being more explicit will prevent trouble down the road.

Also I think there needs to be a sync audit log from the start (even if it isn't exposed anywhere user facing to begin).

Just my two cents and happy to help where needed!

tmgast:
It's already loosely possible to create playthrough with this, but I think it could be solidified in this spec as well. Including the history in RomM would take some of the burden off of the client for storage (though I still think it's a good idea to have a local mirror of the history for offline rollback). I also like the idea of auditing changes with a log that could double as the rollback state selection list. The hard part is updating the UI so the historic saves aren't all dumped into the UI. Seems like I'll probably have to address that to move forward with the improvements.

gantoine:
> would take some of the burden off of the client for storage

No directly related to that conversion, but given how _limited_ some of the targeted devices can be I think it's useful to shift the burden onto the server as much as possible. We can build modules in isolation for sync without affecting the wider system.

tmgast:
Would we be better off overhauling current saves or, my preference, break synced playthroughs off to their own thing so we aren't muddying up the existing UI and model structures for save files?

gantoine:
> Would we be better off overhauling current saves or, my preference, break synced playthroughs off to their own thing so we aren't muddying up the existing UI and model structures for save files?

I think that's a conversation we need to include @zurdi15 in. IMO the current save system isn't very practical, since it requires manually uploading or interfacing with the API, and the proposal would handle things like dropping/syncing folders and ingesting those saves. All that to say, a single system for handling saves would be ideal from a user POV and simpler in the code base. I haven't looked as the PR closely yet so I'll comment further when that's done.

tmgast:
Appreciate the input. I'd definitely like to get his opinion as well. I'm happy to put in the effort to make it happen. I'm going to work on a spec built on top of this branch over the next few days to see what would work best. I'll keep lower spec devices in mind too...

zurdi15:
I'm aligned with what Arcane is exposing here. A single saves/states system it's the ideal way of feeling it really as a central point where all your saves and satetes are not only managed but also that you can use any of them, anywhere.

The UI side should remain the same in terms of merging the new sync work (or just rework what we already have and add the sync part). The user should feel everything is the same and not to have two different systems for different purposes. I also agree that right now using the API to upload/download saves is not too practical, but the details about that are way too long to be discussing it here now in a comment, we should use the RFC for that

gantoine:
@tmgast i've got time this weekend to review this PR, anything you want to change before i dive in?

tmgast:
I wanna add save revisions/history, but I have been too busy with the app to revisit this.

gantoine:
Maybe that happens in another PR and we limit this one to what you've built so far? Would make it easier to review on our end.

tmgast:
I'm good with that... though I'm debating my current implementation of naming here intended for save slots. Let me take another look over it today and make a few adjustments.

tmgast:
I'm wrapping up some hashing stuff now, since it's basically just moving over the process from Argosy to be handled by the server. I think this will be more reliable than having the clients deal with it and ensures the strategy is done correctly and consistently.

The main concern is with hashing compressed content since just hashing the container would always be different, even when compressing the same contents. This solution hashes the inner content and combines that into a single MD5 for comparison and storage.

I'll push the commit for review after a run some test files through it.

backend/alembic/versions/0068_save_sync.py
```diff
@@ -0,0 +1,102 @@
...
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.ForeignKeyConstraint(["device_id"], ["devices.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["save_id"], ["saves.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("device_id", "save_id"),
+    )
```
gantoine, line 74: these are very robust 👌🏼

backend/endpoints/device.py
```diff
@@ -0,0 +1,143 @@
...
+        platform=platform,
+        client=client,
+        client_version=client_version,
+        ip_address=ip_address,
+        mac_address=mac_address,
+        hostname=hostname,
+        last_seen=now,
+    )
+
+    db_device = db_device_handler.add_device(device)
```
gantoine, line 113: there should be a check here if someone tries to register the same device twice, where it returns something like `409 Conflict`

backend/endpoints/device.py
```diff
@@ -0,0 +1,143 @@
...
+    device_id: str,
+    name: str | None = Body(None, embed=True),
+    platform: str | None = Body(None, embed=True),
+    client: str | None = Body(None, embed=True),
+    client_version: str | None = Body(None, embed=True),
+    ip_address: str | None = Body(None, embed=True),
+    mac_address: str | None = Body(None, embed=True),
+    hostname: str | None = Body(None, embed=True),
+    sync_enabled: bool | None = Body(None, embed=True),
+) -> DeviceSchema:
```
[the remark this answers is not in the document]
gantoine: this is good feedback, i'd implement it

backend/handler/database/saves_handler.py
```diff
@@ -42,12 +42,28 @@ def get_save_by_filename(
...
+        )
+
     @begin_session
     def get_saves(
         self,
         user_id: int,
         rom_id: int | None = None,
         platform_id: int | None = None,
+        slot: str | None = None,
+        order_by_updated_at_desc: bool = False,
```
gantoine: would it make more sense to pass in an `order_by` field?

backend/handler/scan_handler.py
```diff
@@ -817,11 +820,41 @@ async def fetch_sgdb_details() -> SGDBRom:
...
+
+
+def _compute_content_hash(file_path: str) -> str | None:
+    try:
+        if zipfile.is_zipfile(file_path):
+            return _compute_zip_hash(file_path)
+        return _compute_file_hash(file_path)
+    except Exception as e:
+        log.debug(f"Failed to compute content hash for {file_path}: {e}")
+        return None
```
gantoine: these should probably be in filesystem/asset_handler.py

backend/endpoints/saves.py
```diff
@@ -1,21 +1,110 @@
...
 from models.assets import Save
+from models.device import Device
+from models.device_save_sync import DeviceSaveSync
 from utils.router import APIRouter
 
+
+def _to_utc(dt: datetime) -> datetime:
+    if dt.tzinfo is None:
+        return dt.replace(tzinfo=timezone.utc)
+    return dt.astimezone(timezone.utc)
```
gantoine: this should be moved to a util file

backend/endpoints/saves.py
```diff
@@ -2,19 +2,68 @@
...
+            )
+        )
+
+    save_data = {
+        key: getattr(save, key)
+        for key in SaveSchema.model_fields
+        if key != "device_syncs" and hasattr(save, key)
+    }
+    save_data["device_syncs"] = device_syncs
+    return SaveSchema.model_validate(save_data)
```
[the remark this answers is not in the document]
gantoine: yes do this ☝🏼

gantoine (changes requested):
damn this is noice \*_chef kiss_\*

backend/handler/database/saves_handler.py
tmgast: Yeah, `order_by` with a default would probably be best. I'd only expect us to request them sorted newest first (desc)... but the process of pruning oldest would probably request it the other way if we ever wanted to screen upcoming save deletions (maybe I'm the only crazy person interested in that though)

backend/endpoints/device.py
tmgast: I actually debated on this because; maybe they'd submit the same device in the event they have to rebuild their system (I'm thinking about future use-case or even app uninstall/reinstall). If they send via the same IP address or HMAC or something, we'd either want to just send back that existing record after matching it or allow them the choice to configure as a new device.

Thoughts?

tmgast: I've added a proposal that defaults to conflict but allows applications/clients to decide how to handle device conflicts:

- reject (409: default returns error)
- recover (200: returns existing entry)
  - `allow_existing` to enable
  - `reset_syncs` (optional) clear sync statuses
- parallel (201: creates a new instance)
  - `allow_duplicate` to enable

backend/handler/database/saves_handler.py
gantoine: > pruning oldest would probably request it the other way if we ever wanted to screen upcoming save deletions

this makes perfect sense 👍🏼

backend/endpoints/device.py
gantoine: i'd say we send back the existing device, but there should be a mechanism (in a future PR) to reset the state of a device in the UI and/or via the API for this case, though we can worry about that in the future.

backend/endpoints/device.py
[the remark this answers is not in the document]
tmgast: Addressed in 8f4402729

backend/handler/database/saves_handler.py
tmgast: Addressed in 220d7531e

backend/handler/scan_handler.py
tmgast: Addressed in bf8cb92e9

backend/endpoints/saves.py
tmgast: Addressed in 132e0ad25

backend/endpoints/saves.py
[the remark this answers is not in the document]
tmgast: Addressed in e3642523f

backend/endpoints/device.py
tmgast: Addressed in a6ecfb738 and 34f48e58a.

backend/handler/filesystem/assets_handler.py
```diff
@@ -1,11 +1,44 @@
...
+
+
+def compute_content_hash(file_path: str) -> str | None:
+    try:
+        if zipfile.is_zipfile(file_path):
+            return compute_zip_hash(file_path)
+        return compute_file_hash(file_path)
+    except Exception as e:
+        log.debug(f"Failed to compute content hash for {file_path}: {e}")
+        return None
```
gantoine, line 39: note to self to refactor this into/with existing code that calculates hashes in filesystem handlers

backend/endpoints/responses/assets.py
```diff
@@ -37,6 +42,25 @@ class SaveSchema(BaseAsset):
...
+            return data
+        result = {}
+        for field_name in cls.model_fields:
+            if field_name in state.unloaded:
+                continue
+            try:
+                result[field_name] = getattr(data, field_name)
+            except InvalidRequestError:
+                continue
+        return result
```
gantoine, line 62: note to self: this is handy and could be extended to other response schemas

PR #2930, opened 2026-01-21
Fix files possibly not found in 7z archives

Delgan:
**Description**

Some archives have an empty "Attributes" field, i.e. `Attributes = `. Here is an example of valid ROM within a `7z` archive;

```
Path = Persona 4 Arena (Europe) (En,Ja).iso
Size = 6669598720
Packed Size = 6690577597
Modified = 
Attributes = 
CRC = CCD39BB9
Encrypted = -
Method = LZMA:24:lc4
Block = 0


```

But because in the code we call `strip()` on each line, the line becomes `Attributes =` and the `elif` branch was never taken, resulting with `largest_file` being `None` after processing.

As a workaround, I just use `lstrip()` because `strip()` is re-called for each value anyway afterwards. The parser could be made more robust, but I don't know enough about the 7z specs. 

**Checklist**

### DIFF
--- a/backend/utils/archive_7zip.py
+++ b/backend/utils/archive_7zip.py
@@ -45,7 +45,6 @@ def process_file_7z(
             line = line.lstrip()
             if line.startswith("Path = "):
                 current_file = line.split(" = ", 1)[1]
-                current_size = 0
             elif line.startswith("Size = "):
                 try:
                     current_size = int(line.split(" = ")[1].strip())


backend/utils/archive_7zip.py
```diff
@@ -42,7 +42,7 @@ def process_file_7z(
         current_size = 0
 
         for line in lines:
-            line = line.strip()
+            line = line.lstrip()
             if line.startswith("Path = "):
                 current_file = line.split(" = ")[1].strip()
```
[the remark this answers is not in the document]
Delgan: That sounds about right.

Delgan:
Interestingly there is an other theoretical edge case, that is if the archived file ends with a whitespace, then `current_file` value will be incorrect. Therefore I think the `strip()` should be removed (I verified that the output of `7z l` does contain the whitespaces).

backend/utils/archive_7zip.py
```diff
@@ -42,9 +42,10 @@ def process_file_7z(
         current_size = 0
 
         for line in lines:
-            line = line.strip()
+            line = line.lstrip()
             if line.startswith("Path = "):
-                current_file = line.split(" = ")[1].strip()
+                current_file = line.split(" = ", 1)[1]
+                current_size = 0
```
gantoine: why is this line required? we set `current_size = 0` in L42

Delgan: It was part of the commit suggested by Gemini (see above), which contained another interesting change. Gemini isn't wrong: it makes the parser a bit more robust. Whether its worth it depends on how much we "trust" the output of `7zz l -slt -ba` (I don't know if it's stable). But I agree that `Size` is likely guaranteed to exist, which makes this line redundant. Besides, there are more important edge cases to consider if we truly wanted to make the parser foolproof (eg. missing "Attributes" or mis-ordered fields).

I reverted the change.

PR #3057, opened 2026-03-02
POC: raw SQL bypass for GET /api/roms (39s to ~2s at limit=10000)

BrandonKowalski:
**Description**

This is a rough proof of concept to demonstrate the performance gains, not a proposal for what the code should look like. The dict-building is admittedly ugly. It exists because bypassing the ORM means we lose the computed properties on the Rom model (path_cover_small, is_unidentified, platform_display_name, etc.) and have to reconstruct them by hand.

The two things that matter for performance:
  1. Raw SQL skips hydrating 10 unused JSON metadata blobs per row
  2. Siblings (loaded via the sibling_roms VIEW) are skipped entirely; they aren't needed for list views

Of these, skipping siblings alone accounts for most of the gain (39s to ~6s). The raw SQL on top brings it from ~6s to ~2s.

**Checklist**

### DIFF
--- a/backend/endpoints/responses/rom.py
+++ b/backend/endpoints/responses/rom.py
@@ -348,6 +348,31 @@ class SiblingRomSchema(BaseModel):
 
 
 class SimpleRomSchema(RomSchema):
+    # Defaults for fields not needed in list views
+    summary: str | None = None
+    alternative_names: list[str] = []
+    youtube_video_id: str | None = None
+    igdb_metadata: RomIGDBMetadata | None = None
+    moby_metadata: RomMobyMetadata | None = None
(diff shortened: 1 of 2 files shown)


gantoine:
I think I see what you're going for here! FWIW `sibling_roms` is needed in the list view to aggregate/merge ROMs when "group roms" is enabled in the UI settings. I think there is some optimization we can do re: hydration of unused JSON metadata blobs, but i'll have to investigate separately. And uh, I don't think a raw SQL approach is best right now lol.

BrandonKowalski:
I'll leave it to you then! Thanks!

PR #3059, opened 2026-03-02
Updated search page to perform a full search when initially opened

Kalekdan:
**Description**
<sup>On opening the search page, performs an immeidate search of all games to show the full library.</sup>

**Checklist**


zurdi15:
this will wipe the query param, so no matter what query param you have, it will always load the full gallery. Something like:

http://romm.instance/search?search=mario won't load a mario search

zurdi15 (changes requested):
Try to check how the query param is loaded and only filter empty by default if the ``search`` query param is empty

gantoine:
Closing as won't do/already implemented.

PR #3102, opened 2026-03-09
Fix OAuth token expiry and implement refresh token rotation

HydroSulphide:
**Description**
The current state of OAuth token implementation is that they are always valid, because of a missing check against their expiration date. I implemented this check in `get_current_active_user_from_bearer_token()`. I also implemented refresh token rotation.

This PR changes the OAuth token flow so expired access tokens are rejected correctly, refresh tokens are consumed and rotated on refresh, and old refresh tokens can no longer be reused. The `/token` endpoint now uses `consume_refresh_token()` for the refresh flow, creates a new refresh token on every successful refresh, and returns `refresh_token`, `expires_in`, and `refresh_expires_in` in the response. I also updated the `TokenResponse` type accordingly.

Additionally, I added a follow-up fix for bearer authentication in `hybrid_auth.py`: if `get_current_active_user_from_bearer_token()` raises an `HTTPException` with status code `401`, it is now handled gracefully instead of causing an internal server error during API calls with expired access tokens.

I tested the changes locally with a token rotation test flow that verifies valid access tokens work before expiry, expired access tokens are rejected by the server, refresh tokens produce a new token pair, old refresh tokens fail after rotation, and expired refresh tokens are rejected.

Most importantly: I fixed a typo in a comment.

**Checklist**

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -1,10 +1,14 @@
+from datetime import datetime, timedelta, timezone
+
 import alembic.config
 import pytest
+from joserfc import jwt
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
 from config.config_manager import ConfigManager
 from handler.auth import auth_handler
+from handler.auth.base_handler import ALGORITHM, oct_key
 from handler.database import (
     db_platform_handler,
     db_rom_handler,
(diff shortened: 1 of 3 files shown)


gantoine:
@HydroSulphide i'll have a look once the comments and failing tests are fixed!

PR #3122, opened 2026-03-11
default to querying all roms

camhorn:
**Description**
The current behavior was that random would only consider roms that were not missing, verified, or with retro achievements.  Since `getRoms` is used exclusively by random buttons, a more sensible default behavior would seem to be to consider all roms.

**Checklist**


zurdi15:
lgtm!

PR #3147, opened 2026-03-21
EmulatorJS-SFU Release Candidate Update.

TechnicallyComputers:
**Description**
I am introducing server based netplay for EmulatorJS by forking EmulatorJS with support for my own SFU relay built on nodejs and mediasoup.

Users that do not enable variable `SFU_NETPLAY_ENABLED=true` in the romm docker container will download and execute EmulatorJS original build, and users that enable the variable will need to setup the additional variables and docker compose edits to configure and run the romm-sfu server, as described in documentation.  You can see which version of EmulatorJS you are running in the hamburger button context menu in the bottom right of the navbar on EmulatorJS.  It will report EmulatorJS, or EmulatorJS-SFU, with links to ROMM discord, and the EmulatorJS-SFU github.

I just tested, and you can literally just change `SFU_NETPLAY_ENABLED` from true to false, and it launches the official or the fork depending on the version in your compose file when you run `compose up -d`

This way, users can actually select which version of EmulatorJS to run.  I might have hardcoded EmulatorJS original behavior a little when I was trying to get my fork working at the start, so netplay might be hard disabled on the original EmulatorJS (P2P netplay version), even if it's enabled in the config.yml.  I do plan to revisit that later since I think we should allow users to switch between the two versions of the software, and even maintain a single config that works with netplay on both versions and just toggle a flag between them.

I am also proposing either forking, or taking over the romm-sfu project to own the server code under the community branding, but if you don't want to I'll keep it and maintain it.  But it's an offer I'm open to discuss, cause I licensed it MIT and I will reuse the code privately on commercial projects probably.

This server still uses WebRTC and still relies on STUN heavily, but if it's configured properly, it should virtually never require TURN except in cases where users are genuinely trapped behind aggressive firewalls that only allow https.  

Server netplay vs P2P netplay enables gameplay with 4 players, even 8 or so with full support of the PS1 multitap range, but I haven't implemented that yet.  

I've tested netplay functions pretty thoroughly.  I've cut out a lot of code from the core emulator.js file to keep it lighter and use my modules for netplay mostly, but that means rebasing emulatorJS code is more manual, and I must pick the commits I want and can't easily rebase, too many changes to rebase anymore.  

In my code changes to Romm, I've added support for a "NetplayID" in the profile menu for all users to set their own NetplayID as an alias in EmulatorJS netplay - where users have persistent identities based on romm authentication, and SFU server interactions are secured by auth read tokens and write tokens issued by the romm API that I wrote - the API tokens include the identity of the ROMM user in them directly for deeper level of user verification.  

So I've done other things like updated locale files for those new data fields, I've mostly built up the API endpoint that I need, and NGINX configuration modifications to accomodate the SFU API workflows.

There is some commits there like I added support to recognize my custom cores when I was attempting a delaysync netplay compatible core build, but it was deeply complicated and not a good focus.  

I also backtracked on a plan to more liberally use the redis database.  I don't think I use it at all anymore come to think of it.  I think I found a way to infer that auth keys are valid and haven't been consumed already without the redis server, and that the relay server is capable of storing a fair amount of metadata itself well enough that I don't need to use the redis workflow, and so previous testing on the redis database configuration has been reverted and left in the initial locked-down localhost secret-key only mode.

I've had to catch up to all the other commits in the time I've been building this, so I had to reorganize my alembic code and I think I set it up right so that the alembic update for netplay ID user data is loading after all the others that already exist, so it doesn't fail migration.  That's worth quickly reviewing or testing on your own, but mine works.

**Checklist**

- [x] I've uploaded the code, and containers, and tested a clean build with new images pulled live

Screenshot_20260320_234010
Screenshot_20260320_234209
photo_2026-03-20_23-45-32
photo_2026-03-20_23-45-32 (2)

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -53,7 +53,6 @@ jobs:
         uses: actions/checkout@v4.3.0
 
       - name: Install mariadb connectors
-        if: matrix.db == 'mariadb'
         run: |
           sudo apt-get update
           sudo apt-get install -y libmariadb3 libmariadb-dev
--- a/DEVELOPER_SETUP.md
+++ b/DEVELOPER_SETUP.md
@@ -76,9 +76,8 @@ sudo apt install libmariadb3 libmariadb-dev libpq-dev
 # Users on macOS can skip this step as RAHasher is not supported
 git clone --recursive https://github.com/RetroAchievements/RALibretro.git
 cd ./RALibretro
-git checkout 1.8.0
+git checkout 1.8.3
 git submodule update --init --recursive
-sed -i '22a #include <ctime>' ./src/Util.h
 make HAVE_CHD=1 -f ./Makefile.RAHasher
 cp ./bin64/RAHasher /usr/bin/RAHasher
 ```
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,14 +43,9 @@ RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | b
 ENV PATH="$NVM_DIR/versions/node/v24.13.1/bin:$PATH"
 
 # Build and install RAHasher (optional for RA hashes)
-RUN git clone --recursive --branch 1.8.1 --depth 1 https://github.com/RetroAchievements/RALibretro.git /tmp/RALibretro
+RUN git clone --recursive --branch 1.8.3 --depth 1 https://github.com/RetroAchievements/RALibretro.git /tmp/RALibretro
 WORKDIR /tmp/RALibretro
-RUN sed -i '22a #include <ctime>' ./src/Util.h \
-    && sed -i '6a #include <unistd.h>' \
-      ./src/libchdr/deps/zlib-1.3.1/gzlib.c \
-      ./src/libchdr/deps/zlib-1.3.1/gzread.c \
-      ./src/libchdr/deps/zlib-1.3.1/gzwrite.c \
-    && make HAVE_CHD=1 -f ./Makefile.RAHasher \
+RUN make HAVE_CHD=1 -f ./Makefile.RAHasher \
     && cp ./bin64/RAHasher /usr/bin/RAHasher
 RUN rm -rf /tmp/RALibretro
 
--- a/backend/adapters/services/rahasher.py
+++ b/backend/adapters/services/rahasher.py
@@ -65,6 +65,7 @@ PLATFORM_SLUG_TO_RETROACHIEVEMENTS_ID: dict[UPS, int] = {
     UPS.WASM_4: 72,
     UPS.SUPERVISION: 63,
     UPS.WIN: 102,
+    UPS.WII: 19,
     UPS.WONDERSWAN: 53,
     UPS.WONDERSWAN_COLOR: 53,
 }
@@ -96,7 +97,7 @@ class RAHasherService:
             return ""
 
         return_code = await proc.wait()
-        if return_code != 1:
+        if return_code != 0:
             if proc.stderr is not None:
                 stderr = (await proc.stderr.read()).decode("utf-8")
             else:
--- a/backend/handler/metadata/ra_handler.py
+++ b/backend/handler/metadata/ra_handler.py
@@ -464,6 +464,7 @@ RA_PLATFORM_LIST: dict[UPS, SlugToRAId] = {
         "name": "Watara/QuickShot Supervision",
     },
     UPS.WIN: {"id": 102, "name": "Windows"},
+    UPS.WII: {"id": 19, "name": "Wii"},
     UPS.WONDERSWAN: {"id": 53, "name": "WonderSwan"},
     UPS.WONDERSWAN_COLOR: {"id": 53, "name": "WonderSwan Color"},
 }
--- a/backend/tests/adapters/services/test_rahasher.py
+++ b/backend/tests/adapters/services/test_rahasher.py
@@ -51,7 +51,7 @@ class TestRAHasherService:
     async def test_calculate_hash_success(self, service):
         """Test successful hash calculation."""
         mock_proc = AsyncMock()
-        mock_proc.wait.return_value = 1  # RAHasher returns 1 on success
+        mock_proc.wait.return_value = 0  # RAHasher returns 0 on success
         mock_proc.stdout.read.return_value = b"a1b2c3d4e5f6789012345678901234ab\n"
         mock_proc.stderr = None
 
@@ -136,7 +136,7 @@ class TestRAHasherService:
     async def test_calculate_hash_with_extra_output(self, service):
         """Test when RAHasher returns hash with extra text."""
         mock_proc = AsyncMock()
-        mock_proc.wait.return_value = 1
+        mock_proc.wait.return_value = 0
         mock_proc.stdout.read.return_value = (
             b"Processing file... Hash: a1b2c3d4e5f6789012345678901234ab Done.\n"
         )
@@ -181,7 +181,7 @@ class TestRAHasherService:
 
         for platform_id, file_path, platform_slug in test_cases:
             mock_proc = AsyncMock()
-            mock_proc.wait.return_value = 1
+            mock_proc.wait.return_value = 0
             mock_proc.stdout.read.return_value = b"a1b2c3d4e5f6789012345678901234ab\n"
             mock_proc.stderr = None
 
@@ -289,7 +289,7 @@ class TestRAHasherServicePerformance:
     async def test_concurrent_hash_calculations(self, service):
         """Test multiple concurrent hash calculations."""
         mock_proc = AsyncMock()
-        mock_proc.wait.return_value = 1
+        mock_proc.wait.return_value = 0
         mock_proc.stdout.read.return_value = b"a1b2c3d4e5f6789012345678901234ab\n"
         mock_proc.stderr = None
 
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5936,9 +5936,9 @@
       }
     },
     "node_modules/flatted": {
-      "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/flatted/-/flatted-3.3.4.tgz",
-      "integrity": "sha512-3+mMldrTAPdta5kjX2G2J7iX4zxtnwpdA8Tr2ZSjkyPSanvbZAcy6flmtnXbEybHrDcU9641lxrMfFuUxVz9vA==",
+      "version": "3.4.2",
+      "resolved": "https://registry.npmjs.org/flatted/-/flatted-3.4.2.tgz",
+      "integrity": "sha512-PjDse7RzhcPkIJwy5t7KPWQSZ9cAbzQXcafsetQoD7sOJRQlGikNbx7yZp2OotDnJyrDcbyRq3Ttb18iYOqkxA==",
       "dev": true,
       "license": "ISC"
     },
--- a/frontend/src/components/Settings/UserInterface/ThemeOption.vue
+++ b/frontend/src/components/Settings/UserInterface/ThemeOption.vue
@@ -10,7 +10,7 @@ withDefaults(defineProps<{ text: string; icon: string }>(), {
 });
 </script>
 <template>
-  <v-item v-slot="{ isSelected, toggle }">
+  <v-item v-slot="{ isSelected, toggle }" :value="text">
     <v-card
       :color="isSelected ? 'primary' : 'romm-gray'"
       class="d-flex align-center"
--- a/frontend/src/plugins/vuetify.ts
+++ b/frontend/src/plugins/vuetify.ts
@@ -3,8 +3,7 @@ import { useLocalStorage } from "@vueuse/core";
 import { createVuetify } from "vuetify";
 import { VDateInput } from "vuetify/labs/VDateInput";
 import "vuetify/styles";
-import { themes, dark, light, autoThemeKey } from "@/styles/themes";
-import { isKeyof } from "@/types";
+import { dark, light } from "@/styles/themes";
 
 const mediaMatch = window.matchMedia("(prefers-color-scheme: dark)");
 mediaMatch.addEventListener("change", (event) => {
@@ -12,13 +11,10 @@ mediaMatch.addEventListener("change", (event) => {
 });
 
 function getTheme() {
-  const storedTheme = useLocalStorage("settings.theme", autoThemeKey);
+  const storedTheme = useLocalStorage("settings.theme", "auto");
 
-  if (
-    storedTheme.value !== autoThemeKey &&
-    isKeyof(storedTheme.value, themes)
-  ) {
-    return themes[storedTheme.value];
+  if (storedTheme.value === "dark" || storedTheme.value === "light") {
+    return storedTheme.value;
   }
 
   return mediaMatch.matches ? "dark" : "light";


.devcontainer/devcontainer.json
```diff
@@ -34,7 +34,8 @@
         "ms-python.pylint",
         "ms-python.black-formatter",
         "vue.volar",
-        "ms-vscode.vscode-typescript-next"
+        "ms-vscode.vscode-typescript-next",
+        "Continue.continue"
```
gantoine, line 38: this should be removed

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["continue.continue"]
+}
```
gantoine, line 3: same with this, it's not required

frontend/src/locales/en_GB/settings.json
```diff
@@ -165,7 +166,15 @@
...
   "username": "Username",
+  "netplay-id": "Netplay ID",
+  "netplay-id-length": "Netplay ID must be between 3 and 32 characters",
+  "netplay-id-chars": "Netplay ID can only contain letters, numbers, emojis, symbols, underscores, and dashes",
+  "netplay-id-hint": "You must create a unique netplay ID to use netplay.",
   "variant-folder": "Variant folder",
+  "netplay-id": "Netplay ID",
+  "netplay-id-length": "Netplay ID must be between 3 and 32 characters",
+  "netplay-id-chars": "Netplay ID can only contain letters, numbers, emojis, symbols, underscores, and dashes",
+  "netplay-id-hint": "You must create a unique netplay ID to use netplay.",
```
gantoine, line 177: you introduced these twice for all the languages

frontend/src/locales/pt_BR/.settings.json.kate-swp
gantoine, line 1: remove this file

backend/alembic/versions/0073_add_netplayid.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Add netplayid column to users table if it does not exist."""
+    conn = op.get_bind()
+    inspector = inspect(conn)
+    columns = [c["name"] for c in inspector.get_columns("users")]
+    if "netplayid" not in columns:
```
gantoine, line 25: you dont need this check, its a new column and should hard fail if the column already exists

backend/config/__init__.py
```diff
@@ -109,17 +128,18 @@ def _get_env(var: str, fallback: str | None = None) -> str | None:
 HLTB_API_ENABLED: Final[bool] = safe_str_to_bool(_get_env("HLTB_API_ENABLED"))
 
 # AUTH
-ROMM_AUTH_SECRET_KEY: Final[str] = _get_env("ROMM_AUTH_SECRET_KEY", "")
-if not ROMM_AUTH_SECRET_KEY:
+ROMM_AUTH_SECRET_KEY: Final[str] = _get_env(
+    "ROMM_AUTH_SECRET_KEY", "test" if IS_PYTEST_RUN else ""
+)
+if not ROMM_AUTH_SECRET_KEY and not IS_PYTEST_RUN:
```
gantoine, line 134: we don't need this, `pytest.ini` overwrites `ROMM_AUTH_SECRET_KEY`

backend/config/__init__.py
```diff
@@ -109,17 +128,18 @@ def _get_env(var: str, fallback: str | None = None) -> str | None:
...
-
-OAUTH_REFRESH_TOKEN_EXPIRE_SECONDS: Final[int] = safe_int(
-    _get_env("OAUTH_REFRESH_TOKEN_EXPIRE_SECONDS"), 7 * 24 * 60 * 60
-)  # 7 days, in seconds
+# SFU internal API auth
+# Dedicated shared secret for SFU -> RomM internal endpoints.
+# SECURITY: this must be distinct from ROMM_AUTH_SECRET_KEY in real deployments.
+ROMM_SFU_INTERNAL_SECRET: Final[str] = _get_env(
+    "ROMM_SFU_INTERNAL_SECRET", "test-sfu-internal" if IS_PYTEST_RUN else ""
+)
```
gantoine, line 142: > this must be distinct from ROMM_AUTH_SECRET_KEY in real deployments

why?

backend/config/__init__.py
```diff
@@ -137,6 +157,12 @@ def _get_env(var: str, fallback: str | None = None) -> str | None:
 INVITE_TOKEN_EXPIRY_SECONDS: Final[int] = safe_int(
     _get_env("INVITE_TOKEN_EXPIRY_SECONDS"), 10 * 60
 )
+OAUTH_ACCESS_TOKEN_EXPIRE_SECONDS: Final[int] = safe_int(
+    _get_env("OAUTH_ACCESS_TOKEN_EXPIRE_SECONDS"), 15 * 60
+)  # 15 minutes
+OAUTH_REFRESH_TOKEN_EXPIRE_SECONDS: Final[int] = safe_int(
+    _get_env("OAUTH_REFRESH_TOKEN_EXPIRE_SECONDS"), 7 * 24 * 60 * 60
+)  # 7 days
```
gantoine, line 165: why were these added?

backend/config/config_manager.py
```diff
@@ -359,6 +365,22 @@ def _get_ejs_controls(self) -> dict[str, EjsControls]:
...
+        """Resolve SFU netplay enabled from config, with env override.
+
+        SFU_NETPLAY_ENABLED (preferred) or EJS_NETPLAY_ENABLED (backwards compat)
+        override config.yml emulatorjs.netplay.enabled when set.
+        """
+        base = pydash.get(
+            self._raw_config, "emulatorjs.netplay.enabled", False
+        )
+        env_val = os.environ.get("SFU_NETPLAY_ENABLED") or os.environ.get(
+            "EJS_NETPLAY_ENABLED"
```
gantoine, line 378: `EJS_NETPLAY_ENABLED` is not an env var

backend/endpoints/responses/config.py
```diff
@@ -28,3 +30,5 @@ class ConfigResponse(TypedDict):
     SCAN_REGION_PRIORITY: list[str]
     SCAN_LANGUAGE_PRIORITY: list[str]
     SCAN_MEDIA: list[str]
+    SFU_HOST: str
+    SFU_PORT: int
```
gantoine, line 34: use the long name for this, `SFU` doesnt' mean anything in this context

backend/endpoints/configs.py
```diff
@@ -62,6 +64,9 @@ def get_config(request: Request) -> ConfigResponse:
         SCAN_REGION_PRIORITY=cfg.SCAN_REGION_PRIORITY,
         SCAN_LANGUAGE_PRIORITY=cfg.SCAN_LANGUAGE_PRIORITY,
         SCAN_MEDIA=cfg.SCAN_MEDIA,
+        # Add SFU configuration
+        SFU_HOST=getenv("SFU_HOST", getenv("EMULATORJS_SFU_HOST", 'localhost')),
+        SFU_PORT=int(getenv('SFU_PORT', getenv('EMULATORJS_SFU_PORT', '3001'))),
```
gantoine, line 69: it it's pulled from `env` it shouldn't be in the config response...

backend/endpoints/user.py
```diff
@@ -367,6 +368,55 @@ async def update_user(
...
+            # Prevent netplay ID from matching any existing username
+            username_collision = db_user_handler.get_user_by_username(form_data.netplayid)
+            if username_collision:
+                msg = f"Netplay ID cannot match an existing username: {form_data.netplayid}"
+                log.error(msg)
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail=msg,
+                )
+            cleaned_data["netplayid"] = form_data.netplayid or None
```
gantoine, line 418: this is a lot of logic for the endpoint, move it somewhere more appropriate

backend/models/user.py
```diff
@@ -67,6 +67,9 @@ class User(BaseModel, SimpleUser):
     ui_settings: Mapped[dict[str, Any] | None] = mapped_column(
         CustomJSON(), default=dict
     )
+    netplayid: Mapped[str | None] = mapped_column(
```
gantoine, line 70:
```suggestion
    netplay_id: Mapped[str | None] = mapped_column(
```

backend/handler/redis_handler.py
```diff
@@ -18,7 +18,13 @@ class QueuePrio(Enum):
...
 
 
-redis_client = Redis.from_url(REDIS_URL)
+if IS_PYTEST_RUN:
+    # Only import fakeredis when running tests, as it is a test dependency.
+    from fakeredis import FakeRedis
+
+    redis_client = FakeRedis(version=7)
+else:
+    redis_client = Redis.from_url(REDIS_URL)
```
gantoine, line 27: why? valkey is available when running pytest locally and in CI

backend/main.py
```diff
@@ -97,8 +98,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
         secret=ROMM_AUTH_SECRET_KEY,
         exempt_urls=[
             re.compile(r"^/api/token.*"),
-            re.compile(r"^/api/client-tokens/exchange"),
-            re.compile(r"^/api/client-tokens/pair/.+/status"),
+            re.compile(r"^/api/sfu/token.*"),
+            re.compile(r"^/api/sfu/internal/verify"),
```
gantoine, line 102: why are these exempt?

docker/init_scripts/docker-entrypoint.sh
```diff
@@ -38,6 +38,13 @@ done
 # support the default value syntax `${VAR:-default}`.
 export ROMM_BASE_PATH=${ROMM_BASE_PATH:-/romm}
 export ROMM_PORT=${ROMM_PORT:-8080}
+# SFU server host and port for netplay audio/video communication.
+# Preferred variables: SFU_HOST / SFU_PORT
+# Backwards-compatible aliases: EMULATORJS_SFU_HOST / EMULATORJS_SFU_PORT
+export SFU_HOST=${SFU_HOST:-${EMULATORJS_SFU_HOST:-localhost}}
+export SFU_PORT=${SFU_PORT:-${EMULATORJS_SFU_PORT:-3001}} # Traditional mediasoup default port 3000 in use by Vite
```
gantoine, line 45: again in this context `SFU` is vague and the full name should be used

docker/init_scripts/init
```diff
@@ -165,11 +165,19 @@ start_bin_valkey-server() {
 	set +o errexit
 
 	while ((retry < max_retries)); do
-		# Attempt to check if valkey TCP port is open
+		# First check if TCP port is open, then verify Valkey accepts Redis protocol (PING)
```
gantoine, line 168: explain this?

docker/init_scripts/init
```diff
@@ -301,6 +309,8 @@ fi
 # Start Valkey server if REDIS_HOST is not set (which would mean user is using an external Redis/Valkey)
 if [[ -z ${REDIS_HOST} ]]; then
 	watchdog_process_pid valkey-server
+	# Export REDIS_HOST so Python startup and workers reliably connect to internal Valkey
+	export REDIS_HOST=127.0.0.1
```
gantoine, line 313: this will break some setups...

docker/Dockerfile
```diff
@@ -25,20 +25,47 @@ ARG UV_VERSION=0.8.24
...
+FROM docker.io/library/python:${PYTHON_VERSION}-alpine${ALPINE_VERSION}@sha256:${PYTHON_ALPINE_SHA256} AS python-alias
 
 
 # FRONTEND BUILD
-FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION}@sha256:${NODE_ALPINE_SHA256} AS frontend-build
+FROM docker.io/library/node:${NODE_VERSION}-alpine${ALPINE_VERSION}@sha256:${NODE_ALPINE_SHA256} AS frontend-build
 WORKDIR /front
 
 COPY ./frontend/package*.json ./
+RUN npm config set engine-strict false
```
gantoine, line 36: why?

docker/Dockerfile
```diff
@@ -25,20 +25,47 @@ ARG UV_VERSION=0.8.24
...
+export default mediasoupClient;
+EOF
+
+RUN npx esbuild /build/entry.js \
+      --bundle \
+      --format=iife \
+      --global-name=mediasoupClient \
+      --define:process.env.NODE_ENV='"production"' \
+      --minify \
+      --outfile=/out/mediasoup-client-umd.js
```
gantoine, line 66: you're gonna have to explain what all this does exactly

docker/Dockerfile
```diff
@@ -164,7 +213,12 @@ RUN apk add --no-cache \
     libpq \
     7zip \
     tzdata \
-    valkey
+    valkey \
+    redis
+
+# Note: we install both `valkey` and `redis` packages.
+# - `valkey` provides the server used by the all-in-one container image.
+# - `redis` provides `redis-cli` which can be useful for debugging.
```
gantoine, line 221: valkey has no cli?

TechnicallyComputers:
Thanks I appreciate it I will look at those more closely and clean up or make better notes.  I have some stuff this weekend but I can start today. 

With my research on this I could offer optimizations to the p2p netplay performance by pushing simple optimizations for the video encoding, capture, and separating audio and video streams, and such.  

But P2P only offers 2p, with more users the host must tax his own system with multiple video and audio channels open for each user and that introduces latency itself when the host has to manage 3 streams.

With a server back-end for multiplayer, this can be migrated into a dedicated server role, or remain a self hosted server, and connected to others in whatever method we see appropriate.  

It can support an excessive number of active livestream clients because it allows the host to control for only one stream of everything, and the relay server manages mirroring media streams and routing them with more efficiency.  

I offered a server based netplay solution concept but it wasn't the direction EmulatorJS wanted to go.  If things change, or down the road, I'd consider adapting a more open model of the server, it wouldn't have authentication support or persistent player identities, but it would still offer a lot gameplay wise.

In this model with integration of the emulator, SFU, and romm at the code level, the product allows for persistent identification of authenticated romm users in netplay, and we can possibly enable retroachievements in EmulatorJS and show profile info for other users with said achievement data in the emulator netplay UI.

The deep integration of the three pieces of software unlocks a lot.  Features like live text chats, voice communication, livestream up to an arbitrary number of clients up to server/mesh capacity.  Highly scalable networked experience and feature support compared to what P2P can manage.  I can build a 4p lobby with spectators where spectators might have the option to request to "pass the controller" and switch places with a player to hotswap quickly and cleanly with a specific person to manage any kind of party gameplay with larger groups.  The restrictions of browser netplay are basically livestream is the most viable and it has some latency, but it's not terrible now and users might enjoy coop games and action RPGs that don't have extreme precision of platformers, since it is very optimized for low video latency by default now.

I'd also note that I'm not the first one to fork emulatorjs and build a netplay server back-end and integrate it into their site, in order to provide a more scalable and managed netplay experience, I've seen implementations using dedicated servers the players select when starting the netplay engine.

docker/Dockerfile
TechnicallyComputers: EmulatorJS fork depends on mediasoup-client-umd.js, a package provided by the mediasoup libraries that the mediasoup SFU server is built around.  Client software needs to have the mediasoup client package in order to work with mediasoup based systems.

This makes sure the clients are always running the latest binary of the mediasoup client since the libraries are actively maintained, and this way I don't have to update the code it will always use the latest libraries.  Better than simply uploading a static binary of it to the project code somewhere.

The server should behave the same with it's own mediasoup libraries.

docker/init_scripts/init
TechnicallyComputers: I was previously planning to integrate more with redis and expand on running it externally and with a separate access control, but that's all really not necessary in the end for my features, and so I will review the valkey and redis related stuff cause none of it really needs changing at all.  This was for healthchecking more advanced redis configs, but I think actually leaving it as is, running locked down for secret key access by romm itself on the localhost only, is the best.  

I'll make a commit that cleans up and remove the unnecessary valkey/redis stuff.

docker/Dockerfile
TechnicallyComputers: I'm gonna remove that cause it doesn't need to be there, but for my own work I did add that originally but I could have just added the valkey-cli explicitly instead..

Valkey cli is a separate package though, and it doesn't come bundled with that valkey package.  I have to run this every time I launch the server in order to verify the contents of the redis server, so I just installed that at runtime.

podman exec -it romm apk add valkey-cli

backend/config/__init__.py
TechnicallyComputers: Because the ROMM_AUTH_SECRET_KEY gives access to redis and the SQL db I assume, and the SFU internal secret is a less privileged key for use by the SFU server(s).

ROMM's SFU api endpoint uses JIT token semantics to mint authentication tokens for ROMM users, it validates the user requesting a cookie for netplay is a romm user authenticated and logged in to ROMM, the token is minted with the secret key, and a cookie is passed with the minted token and the user's Netplay ID for persistent netplay user identification, and secure authentication of netplay users.  

This shared secret is used by the SFU server(s), which might be running externally on another machine or offsite, and as the networking of the servers scales up, some form of tokenized authentication will persist, and a separate minting key from the one used to unlock all your personal stuff and admin control of your server from the one that is potentially used by multiple SFU nodes just reduces threat to the main site if the server is compromised.  It is required, if running with "SFU_NETPLAY_ENABLED=true"

backend/config/__init__.py
TechnicallyComputers: JIT Tokens for Netplay server access - 7 day read token - users can access lobby listings and view lobby metadata for 7 days before refreshing their browser - low threat level so I just gave it a long expiry to reduce network noise.

15 minute token is a token authorizing joining/creating a lobby, and has JWT semantics that the SFU server managers on it's own by simply inferring if it's been used already or not based on it's own built in record keeping.  

The EmulatorJS automatically requests these from the API endpoint, uses them to derive the user's own netplay ID for username display in the UI and passes it to the SFU server where it uses the shared secret to validate the tokens and unpacks the player data tied to the token for it's own listings.

frontend/src/locales/en_GB/settings.json
TechnicallyComputers: Bad user error during the rebase and merge, thanks, I'll clean that up also

backend/config/__init__.py
gantoine: right but what do they have to do with OAUTH?

gantoine:
Right, I'm going to make an executive decision and reject this PR. I'm not comfortable with how much code has to change, how we'll need to continue supporting this feature, or that it relies on a fork created by someone we don't know and can't vet. If I and the team don't understand or see the value in this feature I can't in good conscience accept the burden of maintaining it.

As per our license you're more then welcome to fork the repo and maintain a copy that includes the SFU server.

TechnicallyComputers:
I appreciate the feedback.  I'll be around, and I can understand the logic there.  I think I'll give it some more time, and maybe try another PR that is much much less obtrusive, strip down the features that rely on integration, and offer a version of all of this that requires very little code to maintain in romm itself.  With the goal that it all becomes much more modular and opt-in without adding complexity or burden.

But, let's give it some time.  I'll be around and I'm going to let this sit for a while for now.

PR #3189, opened 2026-03-27
Add outbound HTTP proxy support for backend requests

dafal:
## Summary
- make RomM's shared backend `aiohttp` and `httpx` clients honor standard outbound proxy environment variables
- route the HLTB bootstrap HTTP calls through the same proxy-aware client helpers and add focused tests for proxy env detection
- document supported `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` variables, including lowercase variants, in the Docker Compose example

### DIFF
--- a/backend/utils/context.py
+++ b/backend/utils/context.py
@@ -5,8 +5,9 @@ from typing import TypeVar
 
 import aiohttp
 import httpx
-from config import has_proxy_env
 from fastapi import Request, Response
+
+from config import has_proxy_env
 
 _T = TypeVar("_T")
 
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -21,20 +21,6 @@ def _get_env(var: str, fallback: str | None = None) -> str | None:
     return val.strip() if val else val
 
 
-def has_proxy_env() -> bool:
-    return any(
-        _get_env(var)
-        for var in (
-            "HTTP_PROXY",
-            "HTTPS_PROXY",
-            "NO_PROXY",
-            "http_proxy",
-            "https_proxy",
-            "no_proxy",
-        )
-    )
-
-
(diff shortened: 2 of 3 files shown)


gantoine:
can you give some concrete examples?

zurdi15:
I think this is for services that runs behind a proxy or a restricted net that won't allow requests against 3rd party sites unless you configure the backend to go through the proxy?

dafal:
I run romM connected on internal docker network without direct internet access.

examples/docker-compose.example.yml
```diff
@@ -15,6 +15,12 @@ services:
       - DB_NAME=romm # Should match MARIADB_DATABASE in mariadb
       - DB_USER=romm-user # Should match MARIADB_USER in mariadb
       - DB_PASSWD= # Should match MARIADB_PASSWORD in mariadb
+      - HTTP_PROXY= # Optional outbound proxy for server-side HTTP requests
+      - HTTPS_PROXY= # Optional outbound proxy for server-side HTTPS requests
+      - NO_PROXY= # Optional comma-separated hosts that should bypass the proxy
+      - http_proxy= # Lowercase variants are also supported
+      - https_proxy=
+      - no_proxy=
```
gantoine: remove these, they can live in the docs

backend/utils/context.py
```diff
@@ -13,6 +14,32 @@
...
+        os.environ.get(var)
+        for var in (
+            "HTTP_PROXY",
+            "HTTPS_PROXY",
+            "NO_PROXY",
+            "http_proxy",
+            "https_proxy",
+            "no_proxy",
+        )
+    )
```
gantoine: this should live in `backend/config/__init__.py`

gantoine (changes requested):

dafal:
Applied the requested changes in this branch. If you want, I can also open a separate docs PR with concrete proxy configuration examples.

PR #3204, opened 2026-04-01
feat: emulator streaming via LinuxServer containers (PCSX2 first)

LoneAngelFayt:
## What is this?

This is the start of a framework for streaming emulator gameplay directly in RomM using LinuxServer Docker containers as the backend. PCSX2 is the first one wired up but the goal is to expand this to support other emulators and containers over time.

When a streaming container is configured, a play button shows up on the game card. Clicking it opens an in-browser player that connects to a small HTTP broker running inside the container, which handles launching the ROM and managing the emulator process. The player requires HTTPS since the broker connection won't work over plain HTTP due to browser mixed-content restrictions.

## Changes

Added a new /streaming API router for session management (create, get config, release) that talks to the broker inside the container. Added a streaming section to the config schema (broker_host, broker_port, secret) along with updates to the config example. On the frontend there's a new stream player view, a Pinia store for session state, a route entry, and a conditional play button on the game card. Also updated docker-compose.yml and env.template to show how to run the container alongside RomM.

## Feedback welcome

Happy to rethink the config shape, API design, or anything else based on your input.

AI disclosure: some files and code generated by Claude, but testing and corrections by me


LoneAngelFayt:
here is the broker that goes into the pcsx2 container if you want to take a look at the api inside

https://github.com/LoneAngelFayt/pcsx2-romm-integration

I also cleaned up my commit method so it might be a little easier to read through

Spinnich:
Assuming these docker containers will need to be configured individually by the administrator and then tied to RomM via the config.yml or something like that?

LoneAngelFayt:
correct, they are standalone (so they don't need romm to be used) and they are just configured in romm as part of the config.yml

gantoine:
@LoneAngelFayt i dont know what happened with this PR/branch but it's unreviewable in this state. can you fix it in a new PR?

LoneAngelFayt:
weird, i'm not sure what happened either, i'll make another one as soon as i get a chance

PR #3225, opened 2026-04-06
Add server-side ROM patching endpoint

gantoine:
## Summary

- Adds `POST /api/roms/{id}/patch` endpoint that applies patch files to ROM files server-side using [RomPatcher.js](https://github.com/marcrobledo/RomPatcher.js/)
- Enables third-party apps (mobile clients, handhelds, desktop tools) to request ROM patching using games + patches already in the library, without needing client-side JavaScript
- Supports all 9 patch formats: IPS, UPS, BPS, PPF, RUP, APS, BDF, PMSR, VCDIFF

### How it works

1. A Node.js helper script (`backend/utils/patcher.js`) loads RomPatcher.js from the frontend's `node_modules` (already installed in the Docker image)
2. The FastAPI endpoint looks up ROM and patch files by their `RomFile` IDs, validates paths via `fs_rom_handler.validate_path()`, then invokes the patcher via `asyncio.create_subprocess_exec`
3. The patched ROM is streamed back as a file download; temp files are cleaned up via `BackgroundTask`

### API usage

```bash
# Apply patch (file_id=42) to ROM (file_id=10)
curl -X POST /api/roms/10/patch \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{"patch_file_id": 42}' \
  --output patched.bin

# With custom output filename
curl -X POST /api/roms/10/patch \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{"patch_file_id": 42, "output_file_name": "My Patched ROM"}' \
  --output patched.bin
```

## Test plan

- [ ] Verify the endpoint applies IPS/BPS/UPS patches correctly against known-good ROM + patch combinations
- [ ] Verify error handling for missing files, unsupported formats, and invalid patches
- [ ] Verify temp file cleanup after successful and failed requests
- [ ] Verify authentication (`roms.read` scope required)
- [ ] Test with third-party client using client tokens

https://claude.ai/code/session_01HS6ZvAiBjmLPVB3gGw8eEt


Dockerfile
```diff
@@ -54,6 +54,11 @@ COPY frontend/package.json /app/frontend/
 WORKDIR /app/frontend
 RUN npm install
 
+# Install backend Node helpers (server-side ROM patching)
+COPY backend/utils/rom_patcher/package.json /app/backend/utils/rom_patcher/
+WORKDIR /app/backend/utils/rom_patcher
+RUN npm install
```
gantoine, line 60: **This build path leaves the patcher broken at runtime.** `npm install` here installs the dependency to `backend/utils/rom_patcher/node_modules/rom-patcher/rom-patcher-js`, but `patcher.js` requires it from `path.resolve(__dirname, "rom-patcher-js")` — a sibling of `patcher.js`, not under `node_modules`. `docker/Dockerfile` makes this work with an explicit `COPY ... /node_modules/rom-patcher/rom-patcher-js /backend/utils/rom_patcher/rom-patcher-js` (line ~192), but this `Dockerfile` has no equivalent relocation, so any image built from it would 500 on every patch request (`Cannot find module .../rom-patcher-js`).

Either add the same relocation step here, or make `patcher.js` resolve from the real install path (`require.resolve("rom-patcher/rom-patcher-js/...")` against the local `node_modules`) so both build flows work without a manual copy.

---

backend/utils/rom_patcher/patcher.py
```diff
@@ -0,0 +1,52 @@
...
+    proc = await asyncio.create_subprocess_exec(
+        "node",
+        str(PATCHER_SCRIPT),
+        str(rom_path),
+        str(patch_path),
+        str(output_path),
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+    _, stderr = await proc.communicate()
```
gantoine: No timeout on the subprocess: `await proc.communicate()` waits indefinitely. A pathological patch/ROM (or a RomPatcher.js edge case) that hangs `node` will tie up the request and leak a node process for the life of the worker. Wrap the call in `asyncio.wait_for(...)` and `proc.kill()` on `TimeoutError`, raising `PatcherError` so the endpoint returns a clean 500 instead of hanging.

Related: there's no size cap or concurrency limit anywhere in this path. RomPatcher.js loads the entire ROM into memory in Node, and the endpoint is reachable by any `roms.read` caller (including client tokens), so several concurrent large-ROM patches can OOM the container or fill the temp dir. Worth at least a size guard on `rom_file`/`patch_file` and/or a small semaphore around `apply_patch`.

---

backend/utils/rom_patcher/patcher.js
```diff
@@ -0,0 +1,102 @@
...
+
+  // Parse the patch format
+  const patch = RomPatcher.parsePatchFile(patchFile);
+  if (!patch) {
+    throw new Error("Unsupported or invalid patch format");
+  }
+
+  // Apply patch
+  const patchedRom = RomPatcher.applyPatch(romFile, patch, {
+    requireValidation: false,
```
gantoine, line 84: `requireValidation: false` means a patch is applied even when its embedded source checksum doesn't match the ROM (relevant for BPS/UPS/RUP, which carry source CRCs). The user then gets a file reported as a successful patch that is actually garbage, with no signal that the ROM/patch pairing was wrong. If that's intentional (don't block on mismatch), consider at least surfacing the validation result in the response so callers can warn; otherwise flipping this to `true` and returning a 400 on mismatch would be a better default.

---

backend/endpoints/roms/patch.py
```diff
@@ -0,0 +1,147 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=str(e),
+        ) from e
+    except Exception as e:
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+        log.error(f"Patching error: {e}")
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=f"Unexpected patching error: {e}",
```
gantoine: Minor: this echoes the raw exception text (and the `PatcherError` path returns the node script's stderr) straight into the HTTP response `detail`. Those can include absolute server filesystem paths from `node`/RomPatcher.js. Authed-only, so low severity, but it'd be cleaner to log the detail server-side and return a generic "patching failed" message to the client.

---

gantoine (commented):
Reviewed the server-side patching feature (the endpoint, the Python `apply_patch` wrapper, the Node `patcher.js`, the two Dockerfiles, and the removal of the old client-side patcher). The large diff is mostly the clean removal of the v1/v2 client-side Patcher views — I confirmed there are no dangling references (router, nav button, `rompatcher.d.ts`) left behind.

The core security posture is good:

- **No command injection** — `create_subprocess_exec` is used with an argv array, not a shell, so odd characters in paths are inert.
- **No path traversal** — both the ROM and patch paths come from DB `RomFile` lookups by integer id and go through `fs_rom_handler.validate_path`; nothing user-supplied reaches the filesystem raw.
- **`output_file_name` is sanitized** via `Path(...).stem`, which strips directories and extension, so a malicious name can't escape the temp dir.
- **Temp cleanup** is handled on both the success path (`BackgroundTask`) and the error paths (`shutil.rmtree`).

Issues worth addressing, left inline:

- **Build inconsistency (functional bug):** the root `Dockerfile` runs `npm install` but never relocates `rom-patcher-js` to the sibling path `patcher.js` requires — only `docker/Dockerfile` does. Images built from the root `Dockerfile` would 500 on every patch request. Fix the relocation there, or have `patcher.js` resolve from the real `node_modules` install path so both flows work.
- **No subprocess timeout** in `apply_patch` — a hanging `node` ties up the worker and leaks the process indefinitely. Add `asyncio.wait_for` + `proc.kill()`.
- **No size/concurrency guard** — RomPatcher.js loads the whole ROM into memory, and the endpoint is reachable by any `roms.read` caller; concurrent large patches can OOM the container.
- **`requireValidation: false`** silently produces garbage output on a mismatched ROM/patch pairing — consider surfacing the validation result or defaulting to validate.
- **Minor:** raw exception/stderr text is returned in the response `detail` (can leak server paths).

The test plan in the description is all unchecked and the PR notes it hasn't been verified end-to-end — given this shells out to Node and depends on a Docker-build-time file relocation, an actual run against a known-good ROM+patch (and a check that the root Dockerfile image works, if it's still a supported build) is the most important thing before merge.

---

zurdi15:
The API archutecture is really well done :fire: 
But what if someone wants to patch a ROM and just download it? Shouldn't we keep the patcher as a standalone tool and add the on-the-fly functionality?

gantoine:
> But what if someone wants to patch a ROM and just download it? Shouldn't we keep the patcher as a standalone tool and add the on-the-fly functionality?

ye you can just download it from that tab instead of uploading it

Screenshot 2026-06-24 at 12 55 03 PM

zurdi15:
Ohh okay, missed that one. I was focusing in that you remove the standalone view, it lives in the details view instead :ok_hand:

zurdi15 (approved):

PR #3263, opened 2026-04-15
Fix race condition in collection and favorite rom membership updates

### LINKED ISSUE #3262: [Bug] Race condition in collection/favorites causes games to silently drop when adding rapidly
**RomM version**
v4.8.1

**Describe the bug**
Adding games to collections or marking games as favorites is unreliable. Changes appear to succeed based on the UI response (success toast is shown) but some games silently revert after a page refresh. The problem is most common when adding multiple games in quick succession.

**To Reproduce**
- Navigate to a platform with multiple games
- Rapidly click the star [favorite] icon on many games (more is better, maybe dozens) with minimal pausing between clicks
- Refresh the page
- Some games that appeared to be favorited will have lost their star

The same behavior occurs when adding games to a regular collection.

**Expected behavior**
All games that showed a success toast when favorited or added to a collection should remain in that collection after a page refresh.

- OS: UnRAID
- Version 7.2.4

- Device: Desktop
- OS: Windows 10/11
- Browser: Brave
- Version: 1.89.132

**Additional context**
The root cause appears to be a race condition in the frontend. Every add/remove operation sends the complete rom_ids list as a full replacement to PUT /api/collections/{id}. When multiple requests are in-flight simultaneously, whichever arrives at the server last wins and overwrites the others resulting in silently dropping games with no error shown. A fix has been prepared and will be available as a PR soon.

Spinnich:
Closes #3262

## Problem

Every add/remove operation sent the complete `rom_ids` list as a full replacement 
via `PUT /api/collections/{id}`. When two requests were in-flight simultaneously 
(e.g. rapidly clicking favorites), whichever arrived at the server last would 
overwrite the other — silently dropping games with no error shown.

## Changes

**Backend**
- Added `POST /collections/{id}/roms` — atomically inserts only the specified ROM 
  IDs into the junction table, skipping duplicates
- Added `DELETE /collections/{id}/roms` — atomically removes only the specified 
  ROM IDs from the junction table
- Both endpoints bump `updated_at` on the Collection row (keeps sync system consistent)
- Fixed missing `session.flush()` in `add_rom_user()` so the RomUser record is 
  guaranteed committed before being returned to the caller

**Frontend**
- `useFavoriteToggle.ts` — replaced optimistic mutate + full replace with a single 
  atomic endpoint call; store updated from confirmed server response only
- `FabOverlay.vue` — bulk add/remove favorites now use atomic endpoints; 
  `addToFavorites` now correctly updates the store from the server response 
  (previously it didn't)
- `AddRoms.vue` / `RemoveRoms.vue` — use atomic endpoints instead of full replace

**Tests**
- Added `backend/tests/endpoints/test_collection.py` covering collection CRUD, 
  the new atomic endpoints, permission checks, idempotency, and a regression test 
  that directly reproduces the original race condition

## AI Disclosure

The bug analysis, implementation, and tests in this PR were generated by 
**Claude Code (Claude Sonnet 4.6)**. I (the PR author) identified the issue 
through personal use, directed the investigation, reviewed all changes, and 
verified the fix manually through end-to-end testing in a local Docker environment.

### DIFF
--- a/frontend/src/components/common/Collection/Dialog/AddRoms.vue
+++ b/frontend/src/components/common/Collection/Dialog/AddRoms.vue
@@ -7,7 +7,6 @@ import CollectionListItem from "@/components/common/Collection/ListItem.vue";
 import RAvatarCollection from "@/components/common/Collection/RAvatar.vue";
 import RomListItem from "@/components/common/Game/ListItem.vue";
 import RDialog from "@/components/common/RDialog.vue";
-import type { UpdatedCollection } from "@/services/api/collection";
 import collectionApi from "@/services/api/collection";
 import storeCollections from "@/stores/collections";
 import storeRoms, { type SimpleRom } from "@/stores/roms";
@@ -18,7 +17,7 @@ const { mdAndUp } = useDisplay();
 const show = ref(false);
 const romsStore = storeRoms();
 const collectionsStore = storeCollections();
-const selectedCollection = ref<UpdatedCollection>();
+const selectedCollection = ref<ReturnType<typeof collectionsStore.getCollection>>();
 const roms = ref<SimpleRom[]>([]);
 const emitter = inject<Emitter<Events>>("emitter");
 emitter?.on("showAddToCollectionDialog", (romsToAdd) => {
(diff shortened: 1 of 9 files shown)


frontend/src/components/common/Collection/Dialog/AddRoms.vue
```diff
@@ -18,7 +17,8 @@ const { mdAndUp } = useDisplay();
 const show = ref(false);
 const romsStore = storeRoms();
 const collectionsStore = storeCollections();
-const selectedCollection = ref<UpdatedCollection>();
+const selectedCollection =
+  ref<ReturnType<typeof collectionsStore.getCollection>>();
```
gantoine: you could probably use `UpdatedCollection` or `Collection` over `ReturnType<typeof collectionsStore.getCollection>`

gantoine (changes requested):
excellent AI PR 👌🏼

Spinnich:
> excellent AI PR 👌🏼

Thank you sir! I’ll dig into this suggestion (and the attached CoPilot chatter) tomorrow.

gantoine (approved):

PR #3272, opened 2026-04-16
Add real-time activity tracking for active game sessions

gantoine:
## Description

This PR adds comprehensive real-time activity tracking to display which users are currently playing games across the platform. The feature includes:

**Backend Changes:**
- New `ActivityHandler` class that stores ephemeral play session state in Redis with automatic TTL expiration (90 seconds, refreshed by heartbeats)
- Socket.IO event handlers for `activity:start`, `activity:heartbeat`, `activity:stop`, and `disconnect` to manage browser-based activity
- REST endpoints for external devices (muOS, Android, etc.) to report play sessions via `/activity/heartbeat` and clear activity via `DELETE /activity/heartbeat`
- Endpoints to query all active sessions or filter by ROM
- Automatic cleanup on socket disconnect as a safety net

**Frontend Changes:**
- New `Activity.vue` view displaying all currently active play sessions in a card grid with user avatars, ROM covers, platform info, and elapsed time
- `ActivePlayers.vue` component showing live players on individual ROM detail pages
- `ActivityBtn.vue` navigation button with badge showing active session count
- Activity store (`activity.ts`) managing real-time state via Socket.IO events (`activity:update`, `activity:clear`)
- Integration with the emulator player to emit activity events when games start/stop and periodic heartbeats while playing
- Activity API service for REST endpoints and Socket.IO communication
- i18n translations for activity-related UI strings

**Key Features:**
- Real-time updates across all connected clients via Socket.IO
- Automatic session expiration if heartbeats stop (device disconnect, network loss)
- Support for both browser-based play (web) and external devices (grout, argosy-launcher, etc.)
- Preserves session start time across heartbeat refreshes
- Graceful fallback to stored session data if disconnect event lacks payload

The implementation uses Redis for ephemeral storage with automatic cleanup, ensuring no manual session management is required.

https://claude.ai/code/session_01WzWu5XEEYcAc3EJcfteiFd


backend/endpoints/sockets/activity.py
```diff
@@ -0,0 +1,196 @@
...
+        data_rom_id = data.get("rom_id")
+        rom_id = int(data_rom_id) if data_rom_id else None
+    except (TypeError, ValueError):
+        rom_id = None
+    return user_id, device_id, rom_id
+
+
+@socket_handler.socket_server.on("activity:start")  # type: ignore
+async def activity_start(sid: str, data: ActivityEventPayload) -> None:
+    user_id, device_id, rom_id = _extract_payload(data)
```
gantoine: **Identity is taken from the client payload, not the authenticated socket.** `activity:start`/`activity:heartbeat` build the `ActivityEntry` from the client-supplied `user_id` (via `_extract_payload`), and the connect handler in `endpoints/sockets/logs.py` accepts every connection unconditionally. That means any connected client can broadcast (or `activity:stop`) a "now playing" session on behalf of an arbitrary `user_id`/`rom_id` — the presence board can be spoofed.

The REST path does this correctly by using `request.user.id`. Consider resolving the authenticated user on `connect` (`get_session_from_environ`, like logs.py does) and storing `user_id` in the socket session, then deriving it here instead of trusting the payload. Only `rom_id`/`device_id` should come from the client.

---

backend/endpoints/activity.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class DeviceHeartbeatPayload(BaseModel):
+    rom_id: int = Field(ge=1)
+    device_id: str = Field(min_length=1, max_length=255)
+
+
+def _get_socket_manager() -> socketio.AsyncRedisManager:
+    """Create a write-only Redis manager for emitting from REST endpoints."""
+    return socketio.AsyncRedisManager(REDIS_URL, write_only=True)
```
gantoine: **A fresh `AsyncRedisManager` is created on every heartbeat.** `_get_socket_manager()` is called per request in both `device_heartbeat` and `clear_device_activity` (lines 129 & 152), opening a new Redis connection each time that is never closed. Heartbeats fire roughly every 30s per active device, so this is steady connection churn.

Since the REST app and the Socket.IO server run in the same process, you can emit directly through the already-initialised, Redis-backed `socket_handler.socket_server` (it fans out across workers via its `AsyncRedisManager`), or at minimum cache one module-level `write_only` manager rather than constructing one per call.

---

backend/handler/activity_handler.py
```diff
@@ -0,0 +1,139 @@
...
+            return json.loads(raw)
+        except ValueError:
+            return None
+
+    async def get_all_active(self) -> list[ActivityEntry]:
+        """Get all currently active play sessions across all users."""
+        entries: list[ActivityEntry] = []
+        pattern = f"{self.KEY_PREFIX}*"
+        async for key in async_cache.scan_iter(match=pattern):
+            raw = await async_cache.get(key)
```
gantoine: `get_all_active` does a `scan_iter` followed by a separate `GET` per key (N+1 round-trips), and this runs on every `/activity` load and is the source for the broadcast board. Consider collecting the keys from the scan and issuing a single `MGET`, which removes the per-entry round-trip. Minor at small scale, but it grows linearly with concurrent players.

---

backend/endpoints/activity.py
```diff
@@ -0,0 +1,165 @@
...
+        avatar_path=avatar_path,
+        rom_id=rom_id,
+        rom_name=rom_name,
+        rom_cover_path=rom_cover_path,
+        platform_slug=platform_slug,
+        platform_name=platform_name,
+        device_id=device_id,
+        device_type=device_type,
+        started_at=started_at,
+    )
```
gantoine: `_build_activity_entry` is a pure pass-through: every keyword argument is forwarded to `ActivityEntry(...)` unchanged with identical names. It adds no validation or defaulting, so the indirection can be dropped — call `ActivityEntry(...)` directly in `device_heartbeat`. (`endpoints/sockets/activity.py` already constructs `ActivityEntry` inline.)

---

frontend/src/v2/views/Player/EmulatorJS.vue
```diff
@@ -365,11 +424,17 @@ onMounted(async () => {
...
-// `gameRunning` false), put focus back on Play so a Start-Play loop
-// stays on the pad.
+// Drive the live-activity lifecycle off the deterministic running state:
+// announce on enter, clear + stop heartbeats on exit. Also restores focus
+// to Play on exit so a Start-Play loop stays on the pad.
 watch(gameRunning, (running, prev) => {
+  if (running && !prev) {
+    emitActivityStart();
+    startActivityHeartbeat();
+  }
```
gantoine, line 423: Worth confirming this is the intended trade-off: under v2 the reused v1 `<Player>` (`views/Player/EmulatorJS/Player.vue`) also emits `activity:start`/`heartbeat`/`stop` and runs its own 30s timer, so a single v2 session fires two start emits, two concurrent heartbeat intervals, and two stop emits. The comment above notes the backend ops are idempotent, which keeps it correct, but it's 2× socket traffic and a duplicate timer per player. If the v1 emitter can be gated when running under the v2 shell, only one path would drive the lifecycle.

---

gantoine (commented):
Reviewed the activity-tracking feature end to end (Redis handler, REST + Socket.IO endpoints, stores, and the v1/v2 player + view integration). Solid overall — the ephemeral-TTL-with-heartbeat design is clean, the rom index with stale-member cleanup is a nice touch, `decode_responses=True` means the str members round-trip safely, and the v2 view/card are constitution-compliant (tokens only, breakpoint attrs, `useWrapGridNav`, skeleton/empty states, i18n keys mirrored across all locales).

A few items worth addressing, left inline:

- **Security (highest value):** the Socket.IO handlers trust the client-supplied `user_id`, so presence can be spoofed. Derive identity from the authenticated socket session instead.
- **Resource churn:** a new `AsyncRedisManager` is built on every REST heartbeat — reuse the existing socket server or a single cached manager.
- **Minor:** `get_all_active` N+1 (scan + per-key GET → MGET); the `_build_activity_entry` pass-through helper can be dropped; and the v2 path double-emits via the reused v1 `<Player>` (currently safe via idempotency, but 2× traffic + duplicate timer).

Nothing blocking beyond the spoofing concern, which I'd recommend resolving before merge.

---

PR #3296, opened 2026-04-21
feat(playmatch): send a suggestion to playmatch on manual match when enabled

DevYukine:
**Description**
 
Sends a fire-and-forget suggestion to Playmatch whenever a user updates a ROM through the edit/match endpoint. Its added as as suggestion which gets enqueued on playmatch side and then assuming its valid, the rom exitss on playmatch side and is unmatched will create a suggestion which will be manually reviewed on playmatch side before approving.

This helps in the long run to fill up playmatchs match database for all other RomM users when RomM users do manually matching and have playmatch enabled, arcaneasada agreed with me on Discord about it. Generally improving the Database for everybody.

When `PLAYMATCH_API_ENABLED=true` and a user calls `PUT /roms/{id}`, RomM posts the ROM's resolved provider IDs, hashes, filename, and filesize to `POST /api/suggestion/external/game`.

Payload details:

- `md5`, `sha1`, `fileName`, and `fileSize` come from the first ROM file with size > 0, falling back to ROM-level fields. `sha256` is always `null` since RomM does not compute it.
- `mappings` includes every provider id currently set on the ROM: IGDB, MobyGames, ScreenScraper, RetroAchievements, LaunchBox, Hasheous, TheGamesDB, Flashpoint, HowLongToBeat, Libretro, SteamGridDB, Gamelist. Only `IGDB` is consumed by Playmatch today. Unknown tags are dropped server-side, so future Playmatch providers will start working without a RomM change.

Behavior:

- Scheduled via `asyncio.create_task` and strong-referenced in a module-level set, so the PUT response is not delayed and the task is not garbage-collected before it runs. (Claude recommended this, i wasn't sure if it was needed, feel free to give feedback on this!)
- All exceptions are caught and logged at debug. A Playmatch outage should not interrupt RomM at all.
- No-ops when Playmatch is disabled, when the ROM has no provider IDs, or when the caller passes `unmatch_metadata=true` (the existing early return).

**Disclosure**
I am not a Python developer and am not very familiar with the RomM codebase. I used Claude to analyze the existing Playmatch integration and to help write the new handler method and endpoint hook. I reviewed the result before submitting.

**Checklist**

- [ ] I've added unit tests that cover the changes (should not be needed for this change)

### DIFF
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -1,3 +1,4 @@
+import asyncio
 import binascii
 import json
 from base64 import b64encode
@@ -54,6 +55,7 @@ from handler.metadata import (
     meta_igdb_handler,
     meta_launchbox_handler,
     meta_moby_handler,
+    meta_playmatch_handler,
     meta_ra_handler,
     meta_ss_handler,
 )
@@ -73,6 +75,16 @@ from .files import router as files_router
 from .manual import router as manual_router
 from .notes import router as notes_router
 from .upload import router as upload_router
+
+# The event loop only holds weak refs to tasks; hold strong refs until they finish.
+_background_tasks: set[asyncio.Task[Any]] = set()
+
+
+def _fire_and_forget(coro: Any) -> None:
+    task = asyncio.create_task(coro)
+    _background_tasks.add(task)
+    task.add_done_callback(_background_tasks.discard)
+
 
 router = APIRouter(
     prefix="/roms",
@@ -1439,6 +1451,8 @@ async def update_rom(
     rom = db_rom_handler.get_rom(id)
     if not rom:
         raise RomNotFoundInDatabaseException(id)
+
+    _fire_and_forget(meta_playmatch_handler.submit_manual_match_suggestion(rom))
 
     return DetailedRomSchema.from_orm_with_request(rom, request)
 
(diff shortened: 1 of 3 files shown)


backend/endpoints/roms/__init__.py
```diff
@@ -74,6 +76,16 @@
...
 from .upload import router as upload_router
 
+# The event loop only holds weak refs to tasks; hold strong refs until they finish.
+_background_tasks: set[asyncio.Task[Any]] = set()
+
+
+def _fire_and_forget(coro: Any) -> None:
+    task = asyncio.create_task(coro)
+    _background_tasks.add(task)
+    task.add_done_callback(_background_tasks.discard)
```
gantoine: can you move this into a `util` file?

gantoine (changes requested):
makes sense 👍🏼

DevYukine:
moved `_fire_and_forget` + its strong-ref set into `backend/utils/background_tasks.py` as `fire_and_forget`. updated the import in `backend/endpoints/roms/__init__.py` to use it.

backend/endpoints/roms/__init__.py
DevYukine: done

gantoine (approved):

PR #3306, opened 2026-04-23
fix(screenscraper): improve API response handling and request correctness

### LINKED ISSUE #3305: [Bug] ScreenScraper: missing lookup parameters, unfiltered non-game entries, and HTML entities in metadata
**RomM version**
v4.8.1

**Describe the bug**
Several issues with the ScreenScraper metadata integration cause unreliable ROM matching, unfiltered non-game entries appearing in results, and malformed metadata being stored.

Hash-based lookups omit recommended parameters romnom (ROM filename) and romtype (file type) are not sent alongside hash values during lookup, despite being documented as recommended by the ScreenScraper API. This reduces match accuracy and prevents KO (unknown ROM) submissions from being attributed to RomM.

Non-game entries not filtered, the ScreenScraper API returns BIOS files and other non-game entries with names prefixed ZZZ(notgame) or with a notgame: "true" field. These are not detected or skipped, allowing them to populate ROM metadata.

HTML entities appear literally in metadata. Game names and descriptions returned by the API can contain HTML-encoded characters (&amp;, &#39;, &nbsp;, etc.) that are stored and displayed without decoding.

**To Reproduce**
Configure ScreenScraper credentials and run a library scan
Scan a platform with BIOS files listed in ScreenScraper (e.g. arcade), non-game entries may appear in results
Inspect stored metadata for games with apostrophes or ampersands in their names, HTML entities may be present

**Expected behavior**
Hash-based lookups send all recommended parameters to maximize match rate and enable KO attribution
Non-game entries are silently skipped during scanning
Game names and descriptions are stored with decoded characters

- OS: UnRAID
- Version: v7.2.4

- Device: Desktop PC
- OS: Win. 10/11
- Browser: Brave
- Version: v1.89.137

**Additional context**
These issues were identified by comparing RomM's ScreenScraper integration against the Screenscraper.fr API documentation and ES-DE's reference implementation. A PR with fixes and tests will be submitted soon.

Spinnich:
Closes #3305

### ScreenScraper API improvements

- **Pass `romnom` and `romtype` in hash-based lookups** — the ScreenScraper API recommends sending the ROM filename and file type alongside hash values to improve match accuracy and enable KO (unknown ROM) submissions to be attributed to RomM. Neither was previously sent.
- **Filter non-game entries** — the SS API returns BIOS files and extras with a `notgame: "true"` field or a name prefixed `ZZZ(notgame)`. These are now detected and skipped during both hash-based and name-based lookups.
- **Decode HTML entities in metadata** — game names and descriptions can contain HTML-encoded characters (`&amp;`, `&#39;`, `&nbsp;`, etc.) that are now decoded before being stored.
- **Guard against empty search terms** — after tag-stripping, a filename can produce an empty string; sending a blank query to SS triggers a French error response rather than an empty result set. An early return is now applied in this case.
- **Redact credentials from log output** — ScreenScraper API credentials (`ssid`, `sspassword`, `devid`, `devpassword`) were appearing in plaintext in httpx INFO-level request logs. They are now masked at the formatter level.
- **Add tests** — new unit tests cover `_is_notgame`, `_get_rom_type`, `_decode_html_entities`, media URL handling, region priority logic, and SS-specific HTTP error codes (423, 426, 430, 431).

### CHD file hashing

- **Raw file hashes for CHD files** — CHD files now follow the same hash logic as all other file types: CRC32, MD5, and SHA1 are computed from raw container bytes. This allows ScreenScraper to log KO entries for unrecognised CHD files, which previously never generated KO submissions.
- **Disc-data SHA1 stored separately** — the CHD v5 header SHA1 (the embedded SHA1 of the original disc data) is extracted and stored in a new `chd_sha1_hash` field on `RomFile`. This field is used by metadata providers that index by disc-data SHA1 (specifically Hasheous), rather than raw file SHA1.
- **Hasheous receives only disc-data SHA1** — for CHD files, Hasheous is sent only the CHD header SHA1 with no CRC or MD5, matching the hashes it actually indexes. For all other file types, the raw file hashes are sent as before.
- **Non-blocking hash computation** — large file hash reads are now wrapped in `asyncio.to_thread` to avoid blocking the event loop during CHD file scanning.
- **Disc SHA-1 displayed in UI** — the CHD header SHA1 is shown in the ROM detail view as "Disc SHA-1": per-file in the file selector dropdown for multi-file games, and in the main info section for single-file games.
- **New Alembic migration** — `0080_add_chd_sha1_hash` adds the `chd_sha1_hash` column to the `rom_files` table.

### Hash-based metadata matching fix

- **UNMATCHED and UPDATE scans now use stored hashes** — hash-lookup metadata providers (ScreenScraper, Hasheous, Playmatch) previously only received file hashes when files were being rehashed in the same scan. For UNMATCHED and UPDATE scans of existing ROMs, the handlers now fall back to the stored `rom.files` hashes from the database, enabling hash-based matching without requiring a full rehash.

## AI Disclosure

This PR was written primarily by Claude Code (Anthropic's AI CLI). Code generation, debugging, and research were all AI-assisted. Changes were reviewed and tested manually by the contributor.

## Test plan

- [ ] Run `cd backend && uv run pytest tests/handler/metadata/test_ss_handler.py tests/adapters/services/test_screenscraper.py tests/handler/filesystem/test_roms_handler.py -v`
- [ ] Trigger a library scan and confirm `romnom`/`romtype` appear in ScreenScraper request logs
- [ ] Confirm credentials are masked (`***`) in httpx log output
- [ ] Scan a platform with known non-game entries (e.g. arcade BIOS files) and confirm they are skipped with a warning log
- [ ] Scan a platform with CHD files and confirm CRC/MD5/SHA1 are all populated in the ROM detail view
- [ ] Confirm "Disc SHA-1" chip appears for CHD files in the ROM detail view
- [ ] Confirm Hasheous lookups for CHD files use only the disc-data SHA1 (check debug logs)
- [ ] Run an UNMATCHED scan and confirm hash-based matching is attempted for existing unmatched ROMs
- [ ] Inspect stored metadata for games with special characters in their names and confirm entities are decoded

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Spinnich:
Hey guys, hold off on this for now. I realize the way we handle .chd file hashes (i.e. pull the SHA1 header hash) conflicts with how ss.fr handles hashes on their end.

To make a long story short, I want to try and improve the logic here to do the following...

For ss.fr, when hash matching a .chd file we should compute a CRC, MD5, and SHA1 hash just like we would for any other file type (i.e. .zip, .iso, etc.) and submit that to their jeuInfos.php endpoint for both the best chance of hash matching AND correctly submitting KO scrapes. However, for our other metadata providers that support hash matching (ex. Hasheous), the extracted SHA1 header hash is what we would want to continue to use.

P.S. I'll update the PR's original post to reflect additional changes and follow up with a reply that I've done this.

Spinnich:
Alright, as promised I've updated the original PR description! Things definitely got more complex then what I expected when I started this!

gantoine:
this PR is doing way too much for us to review in 1 go, if you think there are some good changes here i'd suggest breaking it up into smaller chunks/single features!

Spinnich:
Makes sense! I think I can break this out into two PRs to simplify review.

PR #3114, opened 2026-03-10
Add client API tokens with QR pairing flow

tmgast:
## Description

Adds long-lived, revocable, scope-restricted API tokens for external clients (mobile apps, retro handhelds, third-party tools). Tokens support a device-friendly QR code / short-code pairing flow for provisioning without manual credential entry.

### Features

- **Token Management**: Create, list, regenerate, and revoke tokens from settings
- **Per-Token Scopes**: Select a subset of your role's permissions; enforced as intersection at auth time
- **Flexible Expiry**: 30 days, 90 days, 1 year, or never
- **QR Code Pairing**: 8-character short codes with 60s TTL for device-friendly provisioning
- **Copy or Pair**: Choose to copy the raw token directly or pair via QR/code
- **Admin Oversight**: Admins can view all tokens (with username) and revoke any token
- **Standalone Pair Page**: Auth-exempt `/pair` page for QR scan landing with optional custom-scheme callback

> [!NOTE]
> This is a foundational PR to enable persistent API access for client applications, in coordination with device registration, previously released.

### Known Gaps

- `device_id` FK not yet added (deferred until device-registration branch merges)
- Callback support limited to custom URL schemes (no http/https)
- No batch token operations
- Features not exposed to the front-end admin beyond list/revoke

---

### ClientToken

| Field | Type | Description |
|-------|------|-------------|
| `id` | `int` (PK) | Auto-incrementing ID |
| `user_id` | `int` (FK) | Owner of the token |
| `name` | `str` | User-friendly name (e.g., "Steam Deck", "RetroArch") |
| `hashed_token` | `str` | SHA-256 hash of the raw token (unique, indexed) |
| `scopes` | `str` | Space-separated scope list |
| `expires_at` | `datetime?` | Expiration timestamp (null = never) |
| `last_used_at` | `datetime?` | Last authentication timestamp (5-min debounce) |

### ClientTokenSchema Sample

```json
{
  "id": 1,
  "name": "Steam Deck",
  "scopes": ["roms.read", "assets.read", "devices.write"],
  "expires_at": "2026-06-10T00:00:00Z",
  "last_used_at": "2026-03-10T14:30:00Z",
  "created_at": "2026-03-10T12:00:00Z",
  "user_id": 1
}
```

### ClientTokenCreateSchema Sample

```json
{
  "id": 1,
  "name": "Steam Deck",
  "scopes": ["roms.read", "assets.read"],
  "expires_at": "2026-06-10T00:00:00Z",
  "last_used_at": null,
[...]

PR #3308, opened 2026-04-24
Add device authorization flow for TV-app-style pairing (RFC 8628)

tmgast:
## Description

Adds an RFC 8628-style device authorization flow so clients (argosy-launcher and grout) can pair themselves by display rather than requiring the user to manually create a client token and copy it into the device.

A device hits an unauthenticated endpoint with its identifier, client type, and the scopes it wants. The server returns a long `device_code` (secret, for polling) and a short `user_code` plus a QR-ready URL. The user scans the QR on their phone, lands on `/pair/device` in the web UI, logs in if needed, and approves — optionally editing the device name, trimming scopes, and choosing an expiry. The device's next poll retrieves a `ClientToken` that's bound 1:1 to a new (or deduped) `Device` record. Downstream endpoints (`/play-sessions`, `/sync/negotiate`) infer `device_id` from the bound token so the client doesn't have to ship it on every call.

### Features

- **Device-initiated pairing**: no manual copy/paste of tokens — the device drives the handshake
- **1:1 token↔device binding**: every token issued via this flow is linked to its Device
- **Automatic `device_id` inference**: bound tokens surface `device_id` on `request.state` so downstream endpoints pick it up for free
- **Dedupe on re-install**: clients supply a stable `client_device_identifier`; re-running the flow reuses the same Device row
- **Editable scope approval**: user sees exactly what the device requested, can uncheck items
- **Whoami recovery**: `GET /users/me` returns `current_device_id` for bound tokens so a device can identify itself from its token alone
- **TV-app UX**: QR + polling with RFC-compliant `slow_down` / `authorization_pending` / `expired_token` / `access_denied` errors
- **Never-expire default**: suited for long-lived handhelds; 30d / 90d / 1y still selectable
- **Auto-close**: script-opened tabs count down and close; QR-scanned tabs that can't self-close show a close-it-yourself message

> [!NOTE]
> The new flow is purely additive. The existing `/client-tokens` create + pair + exchange path from PR #3114 remains unchanged for users who want CLI/script tokens.

---

### `ClientToken` — added column

| Field | Type | Description |
|---|---|---|
| `device_id` | `str?` (FK → `devices.id`) | Non-null when the token was issued via the device-auth flow. Nullable for legacy/manual tokens. `ON DELETE SET NULL` — revoking a device unbinds rather than cascading. |

### `Device` — added column

| Field | Type | Description |
|---|---|---|
| `client_device_identifier` | `str?` | Client-supplied stable ID (install UUID, hardware ID). Used to dedupe re-registrations of the same device across token resets. Scoped per-user via compound index. |

### `DeviceAuthInitResponse`

| Field | Type | Description |
|---|---|---|
| `device_code` | `str` | 64-char secret; client keeps private, uses to poll |
| `user_code` | `str` | 8-char human-friendly code from ambiguous-glyph-stripped alphabet |
| `verification_path` | `str` | Relative `/pair/device` path; client joins it with its own configured origin |
| `verification_path_complete` | `str` | With `?user_code=…` appended |
| `expires_in` | `int` | 600 |
| `interval` | `int` | 5 |

### `DeviceAuthPendingSchema`

| Field | Type | Description |
|---|---|---|
| `client_device_identifier` | `str` | For user verification on the approval screen |
| `name` / `client` / `platform` / `client_version` | `str` | From the device's init call |
| `requested_scopes` | `list[str]` | Everything the device asked for |
| `allowed_scopes` | `list[str]` | `requested ∩ user.oauth_scopes` — what the UI pre-selects |
| `expires_at` | `datetime` | Derived from current Redis TTL |

### `DeviceAuthTokenResponse`

| Field | Type | Description |
|---|---|---|
| `access_token` | `str` | `rmm_`-prefixed bearer token |
| `device_id` | `str` | Bound device's UUID |
| `scopes` | `list[str]` | Approved subset |
| `expires_at` | `datetime?` | `null` for never-expiring |

---

## New API Endpoints

All under `/api/auth/device/*`.

### `POST /api/auth/device/init`

Open, per-IP rate-limited (10/60s). Initiates a pairing flow.

```json
REQUEST
{
  "client_device_identifier": "argosy-abc123",
  "name": "AYN Thor",
  "client": "argosy-launcher",
  "platform": "Android",
  "client_version": "0.16.0",
  "requested_scopes": [
    "me.read", "me.write",
    "platforms.read", "roms.read",
    "assets.read", "assets.write",
    "roms.user.read", "roms.user.write",
    "collections.read", "collections.write"
  ]
}

RESPONSE (201)
{
  "device_code": "<64 hex chars>",
  "user_code": "ABCD1234",
  "verification_path": "/pair/device",
  "verification_path_complete": "/pair/device?user_code=ABCD1234",
  "expires_in": 600,
  "interval": 5
}
```

### `GET /api/auth/device/pending/{user_code}`

Auth required (`Scope.ME_READ`). Web UI fetches pending metadata to render the approval screen.

Returns `404` for unknown/expired codes, `410` once already approved/denied.

### `POST /api/auth/device/approve`

Auth required (`Scope.ME_WRITE`). User confirms pairing, optionally editing name/scopes/expiry.

```json
REQUEST
{
  "user_code": "ABCD1234",
  "approved_scopes": ["roms.read", "roms.user.write"],
  "device_name": "Retroid Pocket 6",
  "expires_in": "never"
}

RESPONSE (200)
{ "device_id": "uuid-...", "device_name": "Retroid Pocket 6" }
```

Returns `403` if `approved_scopes` exceed `requested ∩ user.oauth_scopes`.

### `POST /api/auth/device/deny`

Auth required (`Scope.ME_WRITE`). Rejects the pairing. TTL shortened to 60s.

### `POST /api/auth/device/token`

Open, per-IP rate-limited (60/60s). Device polls to retrieve the credential.

```json
REQUEST { "device_code": "..." }

RESPONSE on approval (200, one-shot):
{
  "access_token": "rmm_...",
  "device_id": "uuid-...",
  "scopes": ["roms.read", "roms.user.write"],
  "expires_at": null
}

RESPONSE otherwise (400):
{ "detail": "authorization_pending" | "slow_down" | "access_denied" | "expired_token" }
```

### Updated: `GET /api/users/me`

Now returns `current_device_id` for callers authenticated via a device-bound client token (in addition to the existing web-session source). Lets a device identify itself from its token alone.

---

## Auth Integration

The existing `HybridAuthBackend` at `backend/handler/auth/hybrid_auth.py` gains two behaviors when it resolves an `rmm_`-prefixed bearer token:

1. Sets `conn.state.device_id` from `client_token.device_id` (may be `null` for legacy tokens)
2. Bumps `devices.last_seen` via `DBDevicesHandler.update_last_seen_debounced` when the device is bound — 5-minute debounce, same cost profile as the existing `update_last_used` for the token itself

Endpoints that accept a `device_id` in their payload (`play_sessions.py`, `sync.py`) fall back to `request.state.device_id` when the caller omits it. Explicit payload values always win for back-compat.

---

## Device Flow Details

```mermaid
sequenceDiagram
    participant D as Device (argosy-launcher / grout)
    participant S as RomM Server
    participant R as Redis
    participant U as User on phone

    D->>S: POST /api/auth/device/init
    Note over S: Rate limit 10/IP/60s
    S->>R: SETEX device_auth:dc:{code} 600s
    S->>R: SETEX device_auth:uc:{user_code} 600s
    S-->>D: {device_code, user_code, verification_path_complete, expires_in, interval}

    D->>D: render QR of (configured origin + verification_path_complete)

    loop poll while pending
        D->>S: POST /api/auth/device/token {device_code}
        S->>R: GET device_auth:dc:{code}
        R-->>S: {status: pending}
        S-->>D: 400 {detail: "authorization_pending"}
        D->>D: sleep(interval)
    end

    U->>U: scan QR, phone browser opens verification_path_complete
    U->>S: GET /pair/device?user_code=...
    Note over U,S: Vue view, login bounce if needed
    U->>S: GET /api/auth/device/pending/{user_code}
    S-->>U: {name, client, requested_scopes, allowed_scopes, ...}
    U->>U: edit name/scopes/expiry, click Approve
    U->>S: POST /api/auth/device/approve
    Note over S: Scope clamp: approved in (requested intersect user.oauth_scopes)
    S->>S: find-or-create Device by (user_id, client_device_identifier)
    S->>S: create ClientToken bound to device
    S->>R: SETEX device_auth:dc:{code} {status: approved, raw_token, device_id}
    S->>R: DEL device_auth:uc:{user_code}
    S-->>U: {device_id, device_name}
    U->>U: countdown 3s, window.close()

    D->>S: POST /api/auth/device/token {device_code} (next poll)
    S->>R: GETDEL device_auth:dc:{code}
    R-->>S: {status: approved, raw_token, ...}
    S-->>D: {access_token, device_id, scopes, expires_at}

    D->>S: POST /api/play-sessions (no device_id in payload)
    Note over S: HybridAuthBackend sets request.state.device_id
    S-->>D: 201 — session attached to bound device
```

### State transitions (Redis blob at `device_auth:dc:{device_code}`)

| State | Entered by | Response on poll | Retention |
|---|---|---|---|
| `pending` | `/init` | `authorization_pending` | 600s (10 min hard ceiling) |
| `approved` | `/approve` | 200 once, then deleted via `GETDEL` | Same TTL; wiped on first successful poll |
| `denied` | `/deny` | `access_denied` | Shortened to 60s |
| (gone) | TTL or one-shot read | `expired_token` | n/a |

---

## Recommended Client Flow

Pseudocode covering the full lifecycle including expiry + restart. Implement on the device side.

```
loop forever:
    # 1. Request a code
    init = POST /api/auth/device/init {
        client_device_identifier: stable_install_id(),
        name: device_display_name(),
        client: "argosy-launcher",
        platform: "Android",
        client_version: "0.16.0",
        requested_scopes: [
            "me.read", "me.write",
            "platforms.read", "roms.read",
            "assets.read", "assets.write",
            "roms.user.read", "roms.user.write",
            "collections.read", "collections.write",
        ],
    }

    display_qr(server_origin + init.verification_path_complete)
    interval = init.interval
    deadline = now() + init.expires_in

    # 2. Poll for the token
    while now() < deadline:
        sleep(interval)
        resp = POST /api/auth/device/token { device_code: init.device_code }
        match resp.status, resp.detail:
            200, _:                       break   # 200 -> token in hand
            400, "authorization_pending": continue
            400, "slow_down":             interval += 5; continue
            400, "access_denied":         show_user_denied(); return
            400, "expired_token":         break   # restart from step 1

    if got_token:
        store_token(resp.access_token)
        store_device_id(resp.device_id)
        return
    else:
        # Expired or denied -> user can retry
        show_user_retry_prompt()
```

For long-lived tokens (`expires_at: null`), the device never needs to restart until the token is revoked server-side. When calls begin returning `401`, run this flow again — dedupe-by-identifier means the same Device row is reused, preserving the device's session history.

---

## Security

- **No user-supplied callback URLs** — verification URLs are fixed relative paths (`/pair/device`) that the client joins with its own configured origin; the server never reads the request origin and never interpolates client input into it. Because the flow accepts no callback/redirect parameter, there is no `javascript:` / `data:` open-redirect vector to sanitize (unlike the existing `/pair` page).
- **Per-IP rate limits** on both open endpoints (`/init` 10/60s, `/token` 60/60s). 429 on overflow.
- **Per-`device_code` poll pacing** returns RFC 8628 `slow_down` when a single flow polls faster than `interval`.
- **Server-owned `device_id`** — client supplies only the opaque `client_device_identifier`; server generates the UUID. No cross-user squatting on the primary key.
- **Scope clamping** enforced server-side as `approved ⊆ requested ∩ user.oauth_scopes`. UI filters too, but server is the source of truth.
- **Unknown `device_code` → `expired_token`** (same response as TTL-expired). No timing oracle distinguishing "exists but pending" from "never existed."
- **One-shot approved state** — atomic `GETDEL` on first successful `/token` response; replays return `expired_token`.
- **10-minute hard ceiling** — no state persists past the original Redis TTL, regardless of transitions.
- **Log hygiene** — only `device_code_prefix` (first 8 chars) is logged; full secret never appears.
- **CSRF exempt** (anchored `^/api/auth/device/(init|token)/?$`) only on the two open endpoints. Authenticated endpoints keep CSRF protection.

---

### New approval screen at `/pair/device`

- Auth-gated (global guard bounces unauthenticated users through `/login?next=…`)
- Editable device name field, pre-filled from the init request
- Client/platform chips (read-only)
- `client_device_identifier` shown in small muted monospace for user verification
- Scopes as filter-chips pre-selected with server-computed `allowed_scopes`; scopes the device requested but the user lacks appear disabled in a distinct warning color
- Expiry dropdown: 30d / 90d / 1y / Never (defaults to Never)
- On approve: a short countdown auto-closes script-opened tabs; camera-scanned QR tabs that can't self-close clear the countdown and keep the success message ("You can close this tab").
- No callback/redirect URL parameters accepted; the device polls independently.

### Settings → Client API Tokens

Existing tokens list now shows a "linked device" chip next to each bound token. Manual tokens render exactly as before.

### i18n

20 new keys added to all 17 locale `settings.json` files; `check_i18n_locales.py` passes.

---

## Database Changes

- Migration **0080_devices_client_identifier** — `devices.client_device_identifier` + `(user_id, client_device_identifier)` index
- Migration **0081_client_tokens_device_id** — `client_tokens.device_id` FK to `devices.id` (`ON DELETE SET NULL`) + index. Downgrade drops the FK before its backing index (MariaDB requires this).

Both use `batch_alter_table` with `if_not_exists` / `if_exists` flags, MariaDB- and PostgreSQL-safe.

---

> AI Disclosure
> Planning and review assisted by Claude Code

## Checklist

- [x] I've tested the changes locally (flow walked end-to-end against a fresh dev stack)
- [x] I've added unit tests that cover the changes (52 new tests across 13 classes; full suite 1333 passed)


backend/endpoints/responses/identity.py
```diff
@@ -43,7 +43,9 @@ def from_orm_with_request(
             return None
 
         schema = cls.model_validate(db_user)
-        schema.current_device_id = request.session.get("device_id")
+        schema.current_device_id = getattr(
+            request.state, "device_id", None
+        ) or request.session.get("device_id")
```
gantoine, line 48: wouldn't `request.state.get("device_id", None)` work?

backend/models/device.py
```diff
@@ -55,6 +55,10 @@ class Device(BaseModel):
     mac_address: Mapped[str | None] = mapped_column(String(17))
     hostname: Mapped[str | None] = mapped_column(String(255))
 
+    # Stable identifier supplied by the client itself (install UUID, hardware ID),
+    # used to dedupe re-registrations of the same device across token resets.
+    client_device_identifier: Mapped[str | None] = mapped_column(String(255))
```
gantoine, line 60: nice addition

backend/utils/device_auth.py
```diff
@@ -0,0 +1,220 @@
...
+def build_verification_urls(request: Request, user_code: str) -> tuple[str, str]:
+    """Build server-origin URLs only — never from client input.
+
+    This is intentional: no user-supplied callback URLs, so no javascript:/data:
+    XSS surface on the approval page.
+    """
+    base = str(request.base_url).rstrip("/")
+    verification_url = f"{base}/pair/device"
+    verification_url_complete = f"{verification_url}?user_code={user_code}"
+    return verification_url, verification_url_complete
```
gantoine: you should use `yarl.URL` for this

gantoine (changes requested):
backend implementation looks great! can you ask your bot to implement it in the new UI as well? or move it since the old UI is meant to be static.

gantoine (approved):

backend/alembic/versions/0084_devices_client_device_identifier.py
```diff
@@ -0,0 +1,36 @@
...
+Create Date: 2026-04-24 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0084_devices_client_identifier"
+down_revision = "0083_rom_category_soundtrack"
```
gantoine: **Blocking: this branches the Alembic revision tree and produces multiple heads.** `down_revision` here is `0083_rom_category_soundtrack`, but on current `master` the chain past 0083 is `0083 → 0084_add_roms_search_index → 0085_rom_category_screenshot → 0086_screenshot_visibility`. So this migration (and `0085_client_tokens_device_id` chained after it) forms a second branch off 0083, while this PR's own `0087_save_state_visibility` correctly extends 0086. After merge there are **two heads** (`0085_client_tokens_device_id` and `0087_save_state_visibility`) plus duplicate revision *numbers* (two `0084_*`, two `0085_*` files). `alembic upgrade head` will fail with "Multiple head revisions are present" and the backend won't start.

Re-chain the two device-auth migrations onto the real head and renumber to avoid the collision, e.g. `..._devices_client_identifier` → `down_revision = "0087_save_state_visibility"` (renamed to 0088), then `..._client_tokens_device_id` → 0089 after it. Re-run `alembic heads` to confirm a single head.

---

backend/utils/device_auth.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def check_token_poll_rate_limit(request: Request) -> None:
+    client_ip = request.client.host if request.client else "unknown"
+    key = _KEY_TOKEN_RATE.format(client_ip)
+    pipe = sync_cache.pipeline()
+    pipe.incr(key)
+    pipe.expire(key, RATE_LIMIT_WINDOW_SECONDS)
+    count, _ = pipe.execute()
+    if count > TOKEN_POLL_RATE_LIMIT:
```
gantoine, line 98: **This rate limiter will lock out legitimate polling before the flow can complete.** `EXPIRE` is reset on every call, so the per-IP key never lapses while a device keeps polling. The `/token` endpoint is designed to be polled every `interval` (5s) for up to the 600s ceiling — ~120 polls — but the counter climbs monotonically and trips `TOKEN_POLL_RATE_LIMIT` (60) at ~5 minutes, then stays tripped (429) for as long as polling continues. A user who takes longer than ~5 min to scan + approve, or several devices behind one NAT, get a permanent 429 before the 10-minute ceiling they're told they have.

Set the TTL only when the counter is first created so the window actually resets:

```python
count = sync_cache.incr(key)
if count == 1:
    sync_cache.expire(key, RATE_LIMIT_WINDOW_SECONDS)
```

Same applies to `check_authorize_rate_limit` above. (This mirrors `client_tokens.check_rate_limit`, which is safe there only because that endpoint isn't polled continuously — here the pattern is misapplied to a polling endpoint.)

---

backend/endpoints/device_auth.py
```diff
@@ -0,0 +1,336 @@
...
+                platform=data.get("platform"),
+                client_version=data.get("client_version"),
+                client_device_identifier=data["client_device_identifier"],
+                sync_mode=SyncMode.API,
+                last_seen=now,
+            )
+        )
+
+    raw_token = auth_handler.generate_client_token()
+    token = db_client_token_handler.add_token(
```
gantoine: The `Device` row and its bound `ClientToken` are created in two separate handler calls with no shared transaction. If `add_token` fails (or anything between here and `mark_approved` raises), the newly-created `Device` is left orphaned — and on retry the find-or-create path will reuse it, so it lingers. Consider creating both in a single unit of work so the device insert rolls back with the token.

---

backend/endpoints/responses/device_auth.py
```diff
@@ -0,0 +1,66 @@
...
+from .base import BaseModel, UTCDatetime
+
+
+class DeviceAuthInitPayload(BaseModel):
+    client_device_identifier: str = Field(min_length=1, max_length=255)
+    name: str = Field(min_length=1, max_length=255)
+    client: str = Field(min_length=1, max_length=50)
+    platform: str | None = Field(default=None, max_length=50)
+    client_version: str | None = Field(default=None, max_length=50)
+    requested_scopes: list[str] = Field(min_length=1)
```
gantoine: `requested_scopes` is accepted with no cap on list length and no validation that each entry is a real `Scope`. It's clamped at approval (`requested ∩ user.oauth_scopes`) so nothing invalid can actually be granted, but a device can still push an arbitrarily large / junk scope list that gets stored in Redis and rendered verbatim on the approval screen (as "disallowed" chips). Worth adding a `max_length` and validating entries against the Scope enum at the boundary.

---

frontend/src/v2/views/DevicePair.vue
```diff
@@ -0,0 +1,447 @@
...
+  flex-wrap: wrap;
+  gap: 8px;
+  max-height: 168px;
+  overflow-y: auto;
+}
+
+.r-v2-devpair__scope {
+  cursor: pointer;
+}
+.r-v2-devpair__scope:focus-visible {
```
gantoine: v2 constitution (§V "Universal input" / anti-pattern #3) forbids bare `:focus` / `:focus-visible` in styles — focus rings must be modality-gated so they only appear for keyboard/gamepad (`key`/`pad`), not on mouse click. This raw `:focus-visible` will paint an outline on pointer interaction in browsers that treat it that way. Use the modality-gated selectors from `global.css` (the `data-input` pattern the rest of v2 uses) instead.

---

gantoine (commented):
Reviewed the device authorization flow end to end — the Redis state machine (`utils/device_auth.py`), the five endpoints, the `HybridAuthBackend` device binding, the `play_sessions`/`sync` `device_id` fallback, the migrations and models, and the `/pair/device` approval UI plus its auth-gating.

The design is genuinely solid: server-owned `device_id`, scope clamping to `requested ∩ user.oauth_scopes` enforced server-side, one-shot `GETDEL` on the approved blob, `expired_token` for both unknown and TTL-lapsed codes (no oracle), tightly-anchored CSRF exemptions on only the two open endpoints, the `/pair/device` route correctly left out of `authExemptRoutes` so unauthenticated users bounce through login, and logs that only carry a `device_code` prefix. Good test coverage too.

Two things I'd treat as must-fix before merge, left inline:

- **Blocking — migration tree branches into multiple heads.** The two device-auth migrations chain off `0083` instead of the current master head (`0086`), so after merge `alembic upgrade head` fails with "Multiple head revisions are present" (plus duplicate `0084`/`0085` revision numbers). Re-chain onto the real head and renumber.
- **High — the `/token` poll rate limiter refreshes its TTL on every call**, so a continuously-polling device trips the 60-cap at ~5 min and is then locked out (429) before the 10-min approval ceiling it's promised. Set `EXPIRE` only when the counter is first created.

Plus three smaller items inline: non-transactional device+token creation (orphan risk), unvalidated/uncapped `requested_scopes` at `/init`, and a bare `:focus-visible` in `DevicePair.vue` that the v2 constitution wants modality-gated.

One process note: this PR also carries the activity-tracking, save/state-visibility, and asset-list changes, which makes it large to review as one unit — but I scoped this review to the device-auth feature.

---

PR #3318, opened 2026-04-29
ES-DE: Use original media location instead of copying

Namaneo:
**Description**

I kinda bigger feature, that I'll continue to test on my own instance.

Add an opt-in configuration entry to disable media copy to RomM resource directory. With huge libraries, this saves a lot a space on files that would otherwise be duplicated. The downside of this mode is that no media is processed anymore, e.g. no small and big images variants.  

AI Disclaimer: I used Claude to figure out quicker where to look as I'm quite new to this project, and asked it an initial implementation. All the code that is proposed here has been reviewed, understood, cleaned and tested.

**Checklist**

N/A.


gantoine:
i see what you're trying to do here but i think this is the wrong approach, so i'm going to close this PR and tackle this in a more generic way. also the point of gamelist import/export is that romm and es-de don't usually live on the same system, so this duplicate storage use is kinda moot.

PR #3367, opened 2026-05-13
Prefer ROM's own region tag for ScreenScraper and IGDB artwork

gantoine:
## Summary
- ScreenScraper and IGDB now check the ROM's filename region tag (e.g. `(Europe)`) before falling back to `scan.priority.region` when picking artwork and localized titles. Previously, a US-first config would force US covers onto European ROMs even when an EU asset existed.
- Adds a shared `REGION_NAME_TO_PROVIDER_SHORTCODE` map (`"Europe"`→`"eu"`, etc.) so the ROM's parsed region names line up with what providers expect.
- Threads the `Rom` through IGDB's `get_rom`, `get_rom_by_id`, `get_matched_rom_by_id`, and `get_matched_roms_by_name` so they can consult its regions; signatures now match the ScreenScraper handler (rom first, required).

## Test plan
- [x] Scan a library with mixed region tags (e.g. `Game (Europe).rom`, `Game (Japan).rom`) and confirm SS pulls EU/JP covers per the tag while config priority still governs untagged ROMs

## Screenshots

Screenshot 2026-05-13 at 9 30 45 AM
Screenshot 2026-05-13 at 9 30 48 AM


zurdi15 (approved):

Spinnich:
FYI, this is probably causing weird/undesired behavior (imo) in specific instances. As an example...

I have a rom with a file name of `Zaxxon's Motherbase 2000 (Japan, USA) (En).zip`. When I scrape against ss.fr the rom's `Name` field is filled with ss.fr Japanese region game name `Parasquad`. It also pulled Japan region box art. While not technically incorrect, I would prefer the USA region name, art, etc.

Not sure if it would be the best method for approaching this, but maybe have this check all regions in the file name and reference them against the config.yml region priorities to try and decide on user preference when two or more correct answers exist?

https://www.screenscraper.fr/gameinfos.php?gameid=12685

image

PR #3381, opened 2026-05-16
feat(playmatch): add SteamGridDB, ScreenScraper, MobyGames & Launchbox hash support

DevYukine:
**Description**

Playmatch recently expanded its supported `MetadataProvider` to `IGDB, SteamGridDB, ScreenScraper, MobyGames, LaunchBox, EmuReady, OpenVGDB`. This PR removes the`IGDB` gate and forwards every provider id Playmatch returns into the matching fetch function, the same way `igdb_id` was already used.

When `PLAYMATCH_API_ENABLED=true` and at least one Playmatch-supported source (IGDB, MobyGames, ScreenScraper, LaunchBox, SteamGridDB) is enabled, the scan handler now runs Playmatch on hash and uses each returned id as a direct lookup hint for the corresponding metadata source. No platform-level external id is required, since Playmatch matches by file hash.

**Disclosure**
I am not a Python developer and am not very familiar with the RomM codebase. I used Claude to analyze the existing Playmatch integration and to help write the extension of playmatch to other metadata providers. I reviewed the result before submitting.

**Checklist**

### DIFF
--- a/backend/handler/metadata/playmatch_handler.py
+++ b/backend/handler/metadata/playmatch_handler.py
@@ -16,12 +16,23 @@ from utils.context import ctx_httpx_client
 
 class PlaymatchProvider(str, Enum):
     IGDB = "IGDB"
+    SteamGridDB = "SteamGridDB"
+    ScreenScraper = "ScreenScraper"
+    MobyGames = "MobyGames"
+    LaunchBox = "LaunchBox"
+    EmuReady = "EmuReady"
+    OpenVGDB = "OpenVGDB"
 
 
-# (rom attribute, provider tag). Playmatch drops unknown tags server-side.
+# (rom attribute, provider tag). Tag casing matches Playmatch's MetadataProvider
+# enum names (uppercased; Playmatch parses case-insensitively but spacing and
+# underscores must match). Entries for providers Playmatch doesn't currently
+# recognize are kept on purpose: Playmatch drops unknown tags server-side, and
+# preserving them means older RomM clients keep submitting correct tags when
+# Playmatch adds support for more providers.
 _PLAYMATCH_PROVIDER_TAGS: tuple[tuple[str, str], ...] = (
     ("igdb_id", "IGDB"),
-    ("moby_id", "MOBY_GAMES"),
+    ("moby_id", "MOBYGAMES"),
     ("ss_id", "SCREENSCRAPER"),
     ("ra_id", "RETRO_ACHIEVEMENTS"),
     ("launchbox_id", "LAUNCHBOX"),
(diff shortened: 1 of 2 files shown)


gantoine (approved):

Spinnich:
This is just my opinion but I wanted to push back on this PR in it's current state.

Now that Playmatch is expanding to cover so many sources can we make Playmatch an independent metadata source selection like Hasheous? With this PR Playmatch is now inserting itself as a proxy for the majority of RomM's sources in an obfuscated way just from having Playmatch enabled via the env. variables. Currently no other metadata source works this way on RomM and I believe the current logic will be confusing to RomM admins who run scans against their libraries. I want to stress I don't think this was the intent, but the PR subtle puts Playmatch above any other metadata sources via the backend logic when RomM should be treating all metadata sources on equal footing. Lastly, the requested change protects RomM in the event the Playmatch service is ever discontinued as it is less intertwined with other sources directly.

P.S. I realize this is how Playmatch was already working for IGDB but in hindsight I think this is going to be a good a time as any to break Playmatch out as it's own source.

P.P.S. I have no idea how we're going to make the scanner not freak out if, for example, a scan was run with basically every metadata source selected but I may also be overthinking this as it doesn't seem to be an issue with Hasheous currently which is the closest analog.

Edit: I changed how I phrased this as in retrospect it felt hostile to the PR/dev. and that's not my intention!

DevYukine:
> Some pushback on this. Can we make Playmatch an independent metadata source selection like Hasheous for scans instead of it secretly inserting itself as a proxy when selecting other metadata providers? I love both of these projects but imo we should not be forced to go through Playmatch for other metadata sources just from enabling Playmatch via the env. variables.
> 
> I realize this is probably how Playmatch was already working for IGDB but it's becoming obvious to me now that it's putting itself in front of so many metadata providers.

We can discuss this but the whole idea of Playmatch/Hasheous is that you hit it up before the service to do a match and if the provide an id you then get the data of that id, would be kinda weird to enable it and then not use it or make not needed requests to the wanted metadata provider? 

Probably a good idea to discuss this via Discord :)

Spinnich:
> > Some pushback on this. Can we make Playmatch an independent metadata source selection like Hasheous for scans instead of it secretly inserting itself as a proxy when selecting other metadata providers? I love both of these projects but imo we should not be forced to go through Playmatch for other metadata sources just from enabling Playmatch via the env. variables.
> > I realize this is probably how Playmatch was already working for IGDB but it's becoming obvious to me now that it's putting itself in front of so many metadata providers.
> 
> We can discuss this but the whole idea of Playmatch/Hasheous is that you hit it up before the service to do a match and if the provide an id you then get the data of that id, would be kinda weird to enable it and then not use it or make not needed requests to the wanted metadata provider?

Really quick, I just updated my original comment, sorry about that!

PR #3384, opened 2026-05-17
fix(screenscraper): improve API handling, KO scrape data, and metadata sanitization

Spinnich:
**Description**

Several related fixes to RomM's ScreenScraper integration. Grouped here because they all touch the same request/response path and the same test file, but kept as one focused PR (no migration / no schema changes).

### 1. KO scrapes now submit useful data

Previously, hash-based lookups against `jeuInfos.php` only sent `crc`/`md5`/`sha1` (and `systemeid` / `romtaille`). When ScreenScraper couldn't match the hash, the resulting **KO scrape** entry on their side had no filename or ROM type recorded — so it was effectively useless for SS to build their database from.

`ScreenScraperService.get_game_info` already accepted `rom_name` and `rom_type` parameters, they were just never being passed. This PR wires them up at the call site:

- `romnom` is now sent as the actual filename of the largest ROM file in the rom_files set.
- `romtype` is computed by a new `_get_rom_type()` helper that maps file extension to ScreenScraper's expected values:
  - `dossier` — non-top-level files (inside a folder)
  - `iso` — `.iso`, `.cue`, `.chd`, `.gdi`, `.cdi`, `.bin`
  - `rom` — everything else

Net effect: when a ROM doesn't match SS's database, the KO entry SS records now contains enough information to be actionable for them.

### 2. SS-specific HTTP error code handling

ScreenScraper uses non-standard HTTP status codes and returns 401 in situations that aren't actually auth failures. The error handler is updated to:

- `426` → raise 403 "ScreenScraper has blacklisted this application version. Please update RomM."
- `430` → raise 429 "ScreenScraper daily scrape quota exhausted. Try again tomorrow."
- `431` → raise 429 "ScreenScraper daily unrecognized-ROM quota exhausted. Try again tomorrow."
- `423` → raise 503 "ScreenScraper API is currently offline."
- `401` → SS quirk: returned when their server CPU is >60%, **not** when credentials are bad. Now logs a warning and returns an empty dict instead of treating it as an auth error.
- `429` → log warning and retry after 2s (was previously silent).

The same error mapping is applied in both the generic `_request` path and the `get_game_info`-specific exception handler.

### 3. ZZZ(NOTGAME) filtering

ScreenScraper marks non-game entries (BIOS files, demos, etc.) either via a `notgame: "true"` field or by prefixing the name with `ZZZ(NOTGAME)`. A new `_is_notgame()` helper checks both, and these entries are now filtered out from:

- Search results in `_search_rom()`
- Hash-lookup results in `lookup_rom()`
- Name-based matching in `get_matched_roms_by_name()`

Without this, a hash collision with a BIOS file or similar would surface a non-game match as if it were the user's ROM.

### 4. HTML entity sanitization in metadata text

SS returns `&amp;`, `&#x26;`, `&#39;`, `&nbsp;`, `&quot;`, `&copy;` literally in names and summaries. A `_decode_html_entities()` helper now decodes the common ones before metadata is stored, so titles like `"Donkey Kong &amp; Diddy's Kong Quest"` render as `"Donkey Kong & Diddy's Kong Quest"`.

### 5. Strip ScreenScraper credentials from log output

Adds a regex pass in `backend/logger/formatter.py` that strips `ssid` / `sspassword` / `devid` / `devpassword` query parameters from any URL that appears in log output.

The existing storage-layer stripping (`strip_sensitive_query_params` in `extract_media_from_ss_game()`) handles the database side — this complements it at the logging side so URLs surfaced by httpx, exception traces, or future debug logging are also scrubbed.

### 6. Misc. small fix

- `get_rom()` now early-returns if the cleaned-up search term is empty (avoids spurious SS calls for filenames that are entirely tags).

### Test coverage

- `tests/adapters/services/test_screenscraper.py` — new tests for each of the SS-specific HTTP error codes (KO/426/430/431/423/401-CPU-throttle), plus the 429 retry path.
- `tests/handler/metadata/test_ss_handler.py` — new file. Covers `_is_notgame`, `_get_rom_type`, `_decode_html_entities`, the notgame filtering in search/hash/name paths, the empty-search early-return, and the existing credential-stripping behaviour for stored media URLs.

> [!NOTE]
> This PR is one half of an earlier combined PR that was split per maintainer feedback. The companion PR (CHD raw hashing) touches a disjoint set of files and can be reviewed in parallel.

> [!NOTE]
> **AI assistance disclosure**: This PR was developed with assistance from Claude (Anthropic). Claude contributed to authoring portions of the original code, and was used to split the original combined branch into smaller per-feature PRs, draft this description, and verify tests / lint / migrations locally. All code was reviewed and is endorsed by me before submission.

**Checklist**

### DIFF
--- a/backend/adapters/services/screenscraper.py
+++ b/backend/adapters/services/screenscraper.py
@@ -85,10 +85,34 @@ class ScreenScraperService:
             ) from exc
         except aiohttp.ClientResponseError as err:
             if err.status == http.HTTPStatus.TOO_MANY_REQUESTS:
-                # Retry after 2 seconds if rate limit hit
+                log.warning("ScreenScraper: rate limit hit, retrying after 2s")
                 await asyncio.sleep(2)
+            elif err.status == 426:
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="ScreenScraper has blacklisted this application version. Please update RomM.",
+                ) from err
+            elif err.status == 430:
+                raise HTTPException(
+                    status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+                    detail="ScreenScraper daily scrape quota exhausted. Try again tomorrow.",
+                ) from err
+            elif err.status == 431:
+                raise HTTPException(
+                    status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+                    detail="ScreenScraper daily unrecognized-ROM quota exhausted. Try again tomorrow.",
+                ) from err
+            elif err.status == 423:
+                raise HTTPException(
+                    status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                    detail="ScreenScraper API is currently offline.",
+                ) from err
+            elif err.status == http.HTTPStatus.UNAUTHORIZED:
+                log.warning(
+                    "ScreenScraper API is temporarily unavailable (server CPU >60%)"
+                )
+                return {}
             else:
-                # Log the error and return an empty dict if the request fails with a different code
                 log.error(err)
                 return {}
         except json.JSONDecodeError as exc:
@@ -117,11 +141,32 @@ class ScreenScraperService:
                 )
             return await res.json()
         except (aiohttp.ClientResponseError, aiohttp.ServerTimeoutError) as err:
-            if (
-                isinstance(err, aiohttp.ClientResponseError)
-                and err.status == http.HTTPStatus.UNAUTHORIZED
-            ):
-                return {}
+            if isinstance(err, aiohttp.ClientResponseError):
+                if err.status == http.HTTPStatus.UNAUTHORIZED:
+                    log.warning(
+                        "ScreenScraper API is temporarily unavailable (server CPU >60%)"
+                    )
+                    return {}
+                elif err.status == 426:
+                    raise HTTPException(
+                        status_code=status.HTTP_403_FORBIDDEN,
+                        detail="ScreenScraper has blacklisted this application version. Please update RomM.",
+                    ) from err
+                elif err.status == 430:
+                    raise HTTPException(
+                        status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+                        detail="ScreenScraper daily scrape quota exhausted. Try again tomorrow.",
+                    ) from err
+                elif err.status == 431:
+                    raise HTTPException(
+                        status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+                        detail="ScreenScraper daily unrecognized-ROM quota exhausted. Try again tomorrow.",
+                    ) from err
+                elif err.status == 423:
+                    raise HTTPException(
+                        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                        detail="ScreenScraper API is currently offline.",
+                    ) from err
 
             log.error(err)
             return {}
--- a/backend/handler/metadata/ss_handler.py
+++ b/backend/handler/metadata/ss_handler.py
@@ -1,4 +1,4 @@
-import base64
+import html
 import re
 from datetime import datetime
 from typing import Final, NotRequired, TypedDict
@@ -30,8 +30,6 @@ from .base_handler import (
     strip_sensitive_query_params,
 )
 
-SS_DEV_ID: Final = base64.b64decode("enVyZGkxNQ==").decode()
-SS_DEV_PASSWORD: Final = base64.b64decode("eFRKd29PRmpPUUc=").decode()
 SENSITIVE_KEYS = {"ssid", "sspassword"}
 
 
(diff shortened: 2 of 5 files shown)


Spinnich:
image

backend/adapters/services/screenscraper.py
```diff
@@ -85,10 +85,34 @@ async def _request(self, url: str, request_timeout: int = 120) -> dict:
...
+            elif err.status == 431:
+                raise HTTPException(
+                    status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+                    detail="ScreenScraper daily unrecognized-ROM quota exhausted. Try again tomorrow.",
+                ) from err
+            elif err.status == 423:
+                raise HTTPException(
+                    status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                    detail="ScreenScraper API is currently offline.",
+                ) from err
```
gantoine, line 109: you sure these codes are actually returned by ssfr? if so you should use the `HTTStatus.*` equivalents if they exist: https://docs.python.org/3/library/http.html

backend/handler/metadata/ss_handler.py
```diff
@@ -82,6 +79,19 @@ def get_preferred_media_types() -> list[MetadataMediaType]:
...
+_ISO_EXTENSIONS: Final = frozenset({"iso", "cue", "chd", "gdi", "cdi", "bin"})
+
+_HTML_ENTITIES: Final[dict[str, str]] = {
+    "&amp;": "&",
+    "&#x26;": "&",
+    "&#39;": "'",
+    "&nbsp;": " ",
+    "&quot;": '"',
+    "&copy;": "©",
+}
```
gantoine: 99% sure you can unquote to get these lol

backend/handler/metadata/ss_handler.py
```diff
@@ -482,8 +515,8 @@ def build_ss_game(rom: Rom, game: SSGame) -> SSRom:
     ss_id = int(game["id"]) if game.get("id") is not None else None
     game_rom: SSRom = {
         "ss_id": ss_id,
-        "name": res_name.replace(" : ", ": "),  # Normalize colons
-        "summary": res_summary,
+        "name": _decode_html_entities(res_name.replace(" : ", ": ")),
```
gantoine: keep "Normalize colons" comment

backend/logger/formatter.py
```diff
@@ -55,6 +56,11 @@
 }
 
 
+_CREDENTIAL_PATTERN = re.compile(
+    r"(ssid|sspassword|devid|devpassword)=[^&\s\"]*", re.IGNORECASE
+)
```
gantoine: we have an existing `SENSITIVE_KEYS` object, use it instead to build the pattern

gantoine (changes requested):

backend/adapters/services/screenscraper.py
Spinnich: Yes, it's a bit strange but you can find them in ss.fr's API documentation in the "Retour d'erreurs" (Error Feedback) section at https://api.screenscraper.fr/webapi2.php The PR also includes a a test for each that I just did a sanity check on again to be safe.

Ss.fr reuses these codes for their own meanings that don't match the meanings in the documentation you linked (e.g. ss.fr uses 431 for "daily unrecognized-ROM quota exceeded," where as what you linked shows 431 meaning "request header fields too large.").
 
I'm getting out of my depth but as I understand it is if we used the meanings documented in your link it would mislead someone reading the logs since ss.fr assigns custom meanings to each reponse.

backend/handler/metadata/ss_handler.py
Spinnich: Hey thank you! Should be an easy fix.

backend/handler/metadata/ss_handler.py
Spinnich: Also easy enough!

backend/logger/formatter.py
Spinnich: Makes sense!

backend/handler/metadata/ss_handler.py
Spinnich: Replaced the _HTML_ENTITIES dict and _decode_html_entities helper with calls to Python's stdlib html.unescape() in build_ss_game(). Same behavior for the entities we previously handled (&amp;, &quot;, etc.), plus broader coverage for any other entity SS might return.

backend/handler/metadata/ss_handler.py
Spinnich: Added the # Normalize colons comment back next to res_name.replace(" : ", ": ") in build_ss_game()

backend/logger/formatter.py
Spinnich: Replaced the hardcoded 4-key regex in formatter.py with a pattern dynamically built from base_handler.SENSITIVE_KEYS, lazy-imported with a try/except to survive the boot-time circular import (formatter is loaded transitively by base_handler).

PR #3385, opened 2026-05-17
feat(hashing): compute raw CHD hashes and route disc-data SHA1 to Hasheous

Spinnich:
**Description**

### The problem

CHD files were being treated as a special case in RomM's hashing pipeline. Instead of computing CRC32 / MD5 / SHA1 from the raw container bytes the way every other file type does, only the disc-data SHA1 from the CHD v5 header was extracted. That value was then injected into the regular `sha1_hash` field via a `CHDHashWrapper` shim that mimicked the hashlib interface.

This caused two concrete downstream problems:

1. **ScreenScraper KO scrapes for CHD files were useless to SS.** When SS doesn't recognise a hash, it logs a "KO" entry to help build its database. RomM was sending it the disc-data SHA1 in the `sha1` field — a value SS doesn't index by — so SS would log a KO entry against a hash that could never match anything in their pipeline. The companion PR ([#XXXX](#)) ensures KO scrapes carry `romnom`/`romtype`, but that fix is moot for CHDs unless we send the actual file hashes.
2. **The raw container hashes were never available**, so anything that legitimately wanted the file's CRC32 / MD5 / SHA1 (UI display, comparison against external DAT files, future integrations) had no value to use.

### The fix

CHD files now follow the same hash path as every other file type — `_calculate_rom_hashes()` runs CRC32 / MD5 / SHA1 over the raw container bytes. The disc-data SHA1 (extracted from the CHD v5 header) is kept, but stored in a **new dedicated field** so it doesn't collide with the raw SHA1.

Specifically:

- **New `rom_files.chd_sha1_hash` column** (alembic migration `0080`). Nullable, defaults to NULL. Only populated for `.chd` files.
- **`CHDHashWrapper` removed.** No longer needed — disc-data SHA1 is extracted separately via `extract_chd_hash()` instead of being smuggled through the hashlib pipeline.
- **CHD-specific branch in `extract_chd_hash` removed from `_calculate_rom_hashes()`** — CHDs now fall through to the standard raw-byte read.
- **`FileHash` TypedDict gains `chd_sha1_hash`** so it flows through `_build_rom_file()` into `RomFile`.

### Hasheous routing

Hasheous indexes disc-based games by disc-data SHA1 only — raw file hashes won't match. The `HasheousHandler.lookup_rom()` is updated so that for CHD files (`first_file.chd_sha1_hash` populated), only the disc-data SHA1 is sent and MD5/CRC are left empty. For all other files, behaviour is unchanged.

### RAHasher multi-file fix

For multi-file ROM directories, RomM was calling RAHasher with a `/*` wildcard path. RAHasher can't expand wildcards, so it would silently produce no hash for any folder-based ROM. For dirs that contain CHDs, the code now finds the largest CHD and passes it directly — matching the single-file CHD behaviour and aligning with how RAHasher expects multi-disc games to be addressed.

### Async-safe hash computation

Hash calculations on large files (multi-GB ISOs/CHDs) were blocking the asyncio event loop. `_calculate_rom_hashes()` calls are now wrapped in `asyncio.to_thread(...)` so the scan doesn't stall other operations while a big file is being read.

### `rom.files` fallback during UPDATE / UNMATCHED scans

When a scan doesn't rehash files (UPDATE and UNMATCHED scan types), `fs_rom["files"]` comes back empty, which meant the hash-lookup metadata handlers (Playmatch, Hasheous, ScreenScraper) had no hashes to send. Three call sites in `scan_handler.py` now fall back to `rom.files` (the DB-stored hashes from the last successful scan) when `fs_rom["files"]` is empty:

```python
fs_rom["files"] or rom.files
```

This restores hash-based matching for those scan types.

### Frontend: Disc SHA-1 display

The new disc-data SHA1 is surfaced in the ROM detail view:

- **[FileInfo.vue](frontend/src/components/Details/Info/FileInfo.vue)** — single-file CHDs show the full hash in the ROM info panel.
- **[FileSelectItem.vue](frontend/src/components/Details/Info/FileSelectItem.vue)** — multi-file CHD sets (e.g. multi-disc games) show a truncated `first6…last6` hash per file in the file selector.

### Files

- [backend/alembic/versions/0080_add_chd_sha1_hash.py](backend/alembic/versions/0080_add_chd_sha1_hash.py) — new column
- [backend/models/rom.py](backend/models/rom.py) — `RomFile.chd_sha1_hash` field
- [backend/endpoints/responses/rom.py](backend/endpoints/responses/rom.py) — schema exposes field
- [backend/endpoints/sockets/scan.py](backend/endpoints/sockets/scan.py) — socket payload includes field
- [backend/handler/filesystem/roms_handler.py](backend/handler/filesystem/roms_handler.py) — raw-hash path, RAHasher fix, asyncio.to_thread, CHDHashWrapper removal
- [backend/handler/metadata/hasheous_handler.py](backend/handler/metadata/hasheous_handler.py) — disc-data SHA1 routing for CHDs
- [backend/handler/scan_handler.py](backend/handler/scan_handler.py) — `rom.files` fallback for UPDATE / UNMATCHED scans
- Two Vue components + the generated TS schema for the new field

### Test coverage

`tests/handler/filesystem/test_roms_handler.py` was updated to reflect the new behaviour:

- `test_get_rom_files_with_chd_v5_uses_internal_hash` — confirms disc-data SHA1 ends up in `chd_sha1_hash` (not in `sha1_hash`).
- `test_get_rom_files_with_non_v5_chd_fallback_to_std_hashing` — confirms non-v5 CHDs fall through to standard raw hashing.
- The full `TestExtractCHDHash` class (21 tests) covers header parsing edge cases — version boundaries, truncated headers, permission errors, zero/max SHA1, etc.
- Stale tests for the removed `CHDHashWrapper` were deleted.

> [!NOTE]
> This PR is one half of an earlier combined PR that was split per maintainer feedback. The companion PR ([#3384](#)) covers ScreenScraper API handling improvements and touches a disjoint set of files, so the two can be reviewed in parallel.

> [!NOTE]
> **AI assistance disclosure**: This PR was developed with assistance from Claude (Anthropic). Claude contributed to authoring portions of the original code (the squashed commit retains `Co-Authored-By: Claude Sonnet 4.6`), and was used to split the original combined branch into smaller per-feature PRs, draft this description, and verify tests / lint / migration round-trip / frontend rendering locally. All code was reviewed and is endorsed by me before submission.

**Checklist**

### DIFF
--- a/backend/alembic/versions/0080_add_chd_sha1_hash.py
+++ b/backend/alembic/versions/0080_add_chd_sha1_hash.py
@@ -1,0 +1,28 @@
+"""Add chd_sha1_hash to rom_files
+
+Revision ID: 0080_add_chd_sha1_hash
+Revises: 0079_add_rom_files_rom_id_index
+Create Date: 2026-04-24 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "0080_add_chd_sha1_hash"
+down_revision = "0079_add_rom_files_rom_id_index"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("rom_files", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("chd_sha1_hash", sa.String(length=100), nullable=True),
+            if_not_exists=True,
+        )
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("rom_files", schema=None) as batch_op:
+        batch_op.drop_column("chd_sha1_hash", if_exists=True)
--- a/backend/endpoints/responses/rom.py
+++ b/backend/endpoints/responses/rom.py
@@ -163,6 +163,7 @@ class RomFileSchema(BaseModel):
     md5_hash: str | None
     sha1_hash: str | None
     ra_hash: str | None
+    chd_sha1_hash: str | None
     category: RomFileCategory | None
 
 
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -347,6 +347,7 @@ async def _identify_rom(
                 md5_hash=file.md5_hash,
                 sha1_hash=file.sha1_hash,
                 ra_hash=file.ra_hash,
+                chd_sha1_hash=file.chd_sha1_hash,
             )
             for file in fs_rom["files"]
         ]
--- a/backend/handler/filesystem/roms_handler.py
+++ b/backend/handler/filesystem/roms_handler.py
@@ -1,3 +1,4 @@
+import asyncio
 import binascii
 import bz2
 import fnmatch
@@ -10,7 +11,7 @@ import zlib
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from pathlib import Path
-from typing import IO, Any, Final, Literal, TypedDict, cast
+from typing import IO, Any, Final, Literal, TypedDict
 
 import magic
 import zipfile_inflate64  # trunk-ignore(ruff/F401): Patches zipfile to support Enhanced Deflate
@@ -114,6 +115,7 @@ class FileHash(TypedDict):
     crc_hash: str
     md5_hash: str
     sha1_hash: str
+    chd_sha1_hash: str
 
 
 def is_compressed_file(file_path: str) -> bool:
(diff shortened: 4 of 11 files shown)


Spinnich:
image

image

gantoine (approved):

PR #3412, opened 2026-05-23
Composite hashing for compressed archives

### LINKED ISSUE #3411: [Feature] Compute composite hash & individual file hashes for multi-file archives)
**Is your feature request related to a problem?**
Multi-file archives (.zip, .tar, .7z) currently hash only the largest internal file, ignoring all others. This means Rom.crc_hash / Rom.md5_hash / Rom.sha1_hash are never populated for archive-based ROMs, and only one RomFile record is created regardless of how many files the archive contains. Folder-based multi-part ROMs already contain this logic where each file gets its own individual hash, and a composite hash is accumulated across all files. Adding this gives more hash data to work with to improve hash match success against a given metadata source.

Apply the same hashing strategy used for folder-based multi-part ROMs to archives: iterate all eligible internal files in deterministic ASCII order, compute per-file individual CRC/MD5/SHA1 stored on each RomFile, and feed all files through shared accumulators to produce a composite hash stored on Rom. Archives are assumed to be curated ROM sets where every file is intentional, so we only strip universally-junk files (.DS_Store, .log, etc.) but leave user-configured exclusions out.

Hashing the raw archive file itself (e.g. the .zip bytes) produces a compression-dependent hash (other variables would effect the hash as well), the same ROM data repackaged with a different tool yields a different value and won't match any database stably (probably, metadata source provider stored hashes are a bit unpredictable...).

**Additional context**
This is Phase 1 of a two-part change. Phase 2 (deciding which hash each metadata provider's platform should receive for lookup) is a separate issue/PR once the data is available.

Spinnich:
**Description**
For multi-file archives (`.zip`, `.tar`, `.7z`), RomM previously reduced to the single largest internal file before hashing, all other files were ignored and `Rom.crc_hash` / `Rom.md5_hash` / `Rom.sha1_hash` were never populated.

This PR brings archives in line with how folder-based multi-part ROMs already work:
- Each eligible internal file produces its own `RomFile` record with individual CRC/MD5/SHA1 hashes
- All files are fed through shared accumulators in strict ASCII order of their internal paths, producing a composite hash stored on `Rom`
- Only hardcoded default exclusions apply (`.DS_Store`, `gamelist.xml`, `.log`, etc.) — user-configured `EXCLUDED_MULTI_PARTS` filters are intentionally ignored for archives, as archives are assumed to be curated ROM sets where every file is relevant
- Empty, malformed, or all-excluded archives fall back to the previous single-file behavior

No changes to what hashes are submitted to metadata providers (SS, Playmatch, Hasheous), this needs separate consideration.

Fixes #3411

> **AI Disclosure:** This PR was written with Claude Code (Anthropic). Code generation, research, and debugging were AI-assisted; logic and design decisions were reviewed and directed by the author.

**Checklist**


gantoine (approved):

Spinnich:
Awesome additions here! I didn't even notice .rar files were in such a weird state as I don't have any. :sweat_smile:

PR #3438, opened 2026-05-26
fix(screenscraper): inject user credentials for all standard media downloads

Spinnich:
**Description**
PR #3414 added `add_ss_auth_to_url()` and wired it up for *extra* media
downloads (`store_media_file` — fanart, bezels, etc.), but the three standard
media fields — `url_cover`, `url_manual`, and `url_screenshots` — were left
unguarded in both the scan socket and the ROM update endpoint. Those downloads
were hitting screenscraper.fr without `ssid`/`sspassword`, consuming the shared
anonymous/dev quota instead of the user's quota.

- **`ss_handler.py`** — adds a `screenscraper.fr` domain guard to
  `add_ss_auth_to_url` so it is safe to call unconditionally on any URL;
  non-SS URLs (IGDB, LaunchBox, etc.) are returned unchanged, preventing
  credential leakage to third-party servers
- **`endpoints/sockets/scan.py`** — applies `add_ss_auth_to_url` to
  `url_cover`, `url_manual`, and `url_screenshots` in `_identify_rom`
- **`endpoints/roms/__init__.py`** — same at all three call sites in `update_rom`
- **`handler/scan_handler.py`** — fixes a `DetachedInstanceError` caused by
  accessing the deferred `multi_file` column_property on a detached Rom
  instance; replaces `rom.has_nested_single_file or rom.has_multiple_files`
  with `fs_rom["nested"]`, which carries the same meaning and is always
  populated from the filesystem scan

**Checklist**

### DIFF
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -1282,7 +1282,7 @@ async def update_rom(
             path_screenshots = await fs_resource_handler.get_rom_screenshots(
                 rom=rom,
                 overwrite=bool(screenshots_changed),
-                url_screenshots=cleaned_data.get("url_screenshots", []),
+                url_screenshots=[add_ss_auth_to_url(u) for u in url_screenshots],
             )
             cleaned_data.update(
                 {"path_screenshots": path_screenshots, "url_screenshots": []}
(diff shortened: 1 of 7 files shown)


Spinnich:
Sorry for needing to iterate on this one last time. I think we're truly covering all media types now however.

Also, I don't quite understand why this is failing but it seems to be related to PR #3437

gantoine (approved):

PR #3448, opened 2026-05-29
Fix save-sync hash drift, archival save leak, and dedupe scoping

tmgast:
## Description

Cleanup pass on save-sync addressing three independent failure modes
discovered during 4.9 alpha testing that interact in production data:
`content_hash` drift between client and server, null-slot archival
saves leaking into sync flows, and content-hash dedupe collapsing
legitimately-distinct slots.

### Bug fixes

- **Hash drift in `compute_content_hash`** — `zipfile.is_zipfile` was
  called with a relative path, so it silently returned `False` whenever
  the process's CWD wasn't `ASSETS_BASE_PATH`. Every zip save fell
  through to the raw-MD5 branch, persisting hashes that disagreed with
  clients computing the intended per-entry zip-hash. Resolve to a full
  path before the dispatch.

- **Null-slot leak into sync flows** — `_build_negotiate_plan`,
  `sync_push_pull_task`, and `sync_watcher` all treated null-slot
  saves as sync-eligible. Null-slot saves represent web-UI / archival
  uploads; including them in negotiate plans matched them against
  device pushes by filename and overwrote archival data. Filter at
  all three call sites.

- **Dedupe lookups ignored slot** — two upload-path lookups matched
  without considering slot, collapsing legitimately-distinct uploads
  into one record:
  - `get_save_by_content_hash` matched on `(rom_id, user_id,
    content_hash)`, so identical bytes uploaded to different slots
    collapsed.
  - `get_save_by_filename` matched on `(rom_id, user_id, file_name)`.
    When two uploads to different slots happened in the same
    wall-clock second (the datetime tag is per-second), the second
    upload UPDATED the first record's slot instead of creating a new
    row.

  Both now accept an optional slot kwarg, passed from the upload
  endpoint.

### One-shot recovery

- **\`recompute_save_content_hashes\` manual task** — walks every Save
  row, recomputes via the fixed dispatch, and updates rows whose
  values differ. Idempotent and safe to re-run. Counters tracked via
  RQ job meta.

- **Auto-enqueue on first startup** — backend startup runs a single
  \`COUNT(*) WHERE content_hash IS NULL\` query. If any rows exist, the
  recompute task is enqueued on the low-priority RQ queue and the API
  process moves on. Worker handles the recompute out-of-band.
  Subsequent restarts find zero NULL hashes and skip. Admins can
  still trigger the task manually.

> [!NOTE]
> On first restart after upgrade, the API enqueues a one-shot
> recompute of \`content_hash\` for any Save rows missing one. Runs in
> the background on the low-priority RQ worker; API startup is not
> blocked. Subsequent restarts are no-ops. Operators can also
> trigger \`recompute_save_content_hashes\` manually.

---

> AI Disclosure

**Checklist**

### DIFF
--- a/backend/endpoints/saves.py
+++ b/backend/endpoints/saves.py
@@ -162,7 +162,7 @@ async def add_save(
         actual_filename = _apply_datetime_tag(sanitized_save_filename)
 
     db_save = db_save_handler.get_save_by_filename(
-        user_id=request.user.id, rom_id=rom.id, file_name=actual_filename
+        user_id=request.user.id, rom_id=rom.id, file_name=actual_filename, slot=slot
     )
 
     if device and slot and not overwrite:
@@ -222,6 +222,7 @@ async def add_save(
             user_id=request.user.id,
             rom_id=rom.id,
             content_hash=scanned_save.content_hash,
+            slot=slot,
         )
         if existing_by_hash:
             try:
--- a/backend/endpoints/sync.py
+++ b/backend/endpoints/sync.py
@@ -115,9 +115,15 @@ def negotiate_sync(
 
     operations: list[SyncOperationSchema] = []
 
-    # Build a set of server saves for this user, keyed by (rom_id, file_name)
-    # We'll also track which server saves were mentioned by the client
-    server_saves = db_save_handler.get_saves(user_id=request.user.id)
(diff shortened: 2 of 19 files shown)


gantoine (approved):
🔥

PR #3453, opened 2026-05-29
Fix save-sync negotiate thrash and unblock content_hash recompute

tmgast:
## Description

Follow-up to #3448, continuing to refine the save-sync negotiation logic. That
PR fixed `content_hash` drift, the null-slot archival leak, and dedupe scoping;
testing against production data since then surfaced two further issues that
interact: `negotiate` pairs saves on the tagged filename and never converges,
and the recompute backfill #3448 added has been silently failing to enqueue.

### Bug fixes

- **Negotiate pairs on the tagged filename, so it never converges** —
  `add_save` datetime-tags every slot-bound upload's filename
  (`name [YYYY-MM-DD_HH-MM-SS].ext`), but `negotiate_sync` keyed its
  server-save map on the exact `file_name`. Clients report the *untagged*
  canonical name (`computeUploadFileName`), so the lookup always missed:
  every client save came back as `upload` ("exists on client but not on
  server"), and every tagged server row came back as `download` ("exists on
  server but not on client"). The result is perpetual bidirectional sync and
  unbounded row growth per slot (observed 100+ timestamped rows for a single
  game). The content-hash comparison never even ran. Pair on `(rom_id, slot)`
  and collapse each slot to its newest row so `compare_save_state` decides the
  action. The datetime tag is unchanged.

- **`content_hash` recompute never enqueues** — the bootstrap enqueue added in
  #3448 used a job_id containing `:` (`recompute_save_content_hashes:bootstrap`).
  RQ 2.x `Job.set_id` raises `ValueError('id must not contain ":"')`, which the
  broad `except` swallowed, so the one-shot recompute task never ran and legacy
  raw/NULL `content_hash` rows were never reconciled. Drop the colon, remove the
  unsupported `unique=True` kwarg (not an RQ enqueue option — it would be
  forwarded to the task as a `TypeError`), and replace the dead
  `DuplicateJobError` branch with a real `Job.exists` guard.

### Tests

- Real upload -> negotiate round-trip that runs `_apply_datetime_tag` for real
  (the `scan_save` mock only echoes the server-computed tagged filename, never a
  hand-authored one) and asserts the follow-up negotiate is `no_op`.
- 3-device convergence test: the uploader settles to `no_op`; each puller
  downloads exactly once, then converges. Both new tests fail against the old
  `file_name` keying.
- Startup enqueue tests updated; added a regression asserting the job_id
  contains no `:`.

> [!NOTE]
> This also repairs the #3448 recompute backfill so it actually runs: on the
> first restart after deploy, the API enqueues the one-shot `content_hash`
> recompute for any rows still missing/stale (low-priority worker,
> non-blocking). Until this lands, that backfill has been failing silently on
> every boot.

---

> AI Disclosure

**Checklist**

### DIFF
--- a/backend/tests/test_startup.py
+++ b/backend/tests/test_startup.py
@@ -1,7 +1,6 @@
 """Tests for startup-time auto-enqueue of the recompute task."""
 
 import startup
-from rq.exceptions import DuplicateJobError
 
 
 def test_enqueue_recompute_skips_when_no_missing_hashes(mocker):
@@ -21,6 +20,7 @@ def test_enqueue_recompute_fires_when_missing_hashes_present(mocker):
     mocker.patch.object(
         startup.db_save_handler, "count_saves_missing_content_hash", return_value=42
     )
+    mocker.patch.object(startup.Job, "exists", return_value=False)
     enqueue = mocker.patch.object(startup.low_prio_queue, "enqueue")
 
     startup._enqueue_recompute_save_hashes_if_needed()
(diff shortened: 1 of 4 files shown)


backend/tests/test_startup.py
```diff
@@ -39,26 +39,27 @@ def test_enqueue_recompute_fires_when_missing_hashes_present(mocker):
...
-    assert kwargs["unique"] is True
 
 
-def test_enqueue_recompute_silently_skips_duplicate(mocker):
-    """A DuplicateJobError from RQ (job already queued/running) is logged and
-    swallowed, not propagated."""
+def test_recompute_job_id_has_no_colon():
+    """RQ's Job.set_id rejects ':' in a job_id; a colon raised ValueError that
+    the broad except swallowed, so the backfill never enqueued."""
+    assert ":" not in startup.RECOMPUTE_SAVE_HASHES_JOB_ID
```
gantoine: good test but should probably be broadened to confirm the job_is is only valid chars (check discord)

gantoine (approved):

Issue #3471, opened 2026-06-03
[Bug] Region detection fails when renaming ROM files with region names in parentheses


danielhrdez:
**RomM version**
4.8.1

**Describe the bug**
When uploading a ROM file with `(Europe)` in the filename, the EU flag pill displays correctly. However, when renaming an existing ROM and adding `(Europe)` to the filename, the region pill does not appear.

**To Reproduce**

1. Rename a ROM file to include (Europe) in the filename
2. The region flag pill should appear, but it doesn't

**Expected behavior**

**Screenshots**

**Server (please complete the following information):**

- OS: Mint
- Version 22.1

**Client (please complete the following information):**

- Device: macbook air
- OS: MacOS
- Browser: Zen (firefox)
- Version 1.20.1b (Firefox 151.0.2) (aarch64)

**Additional context**

Spinnich:
Additional AI generated context.

**Describe the bug**
RomM reads region and language info from the tags in a ROM's filename — things
like `(Europe)`, `(USA)`, or `(En,Fr,De)` — and shows them as little flag pills
on each game.

The problem is that RomM only reads those tags when a ROM is **first added** to
the library (upload or scan). If you later **rename** an existing ROM and add a
region tag such as `(Europe)`, the flag pill never appears — even though the
exact same filename shows the flag correctly when uploaded fresh.

In short: renaming a ROM updates its filename, but not the region/language data
that RomM derives from that filename. The same bug means a flag from the *old*
name will incorrectly stick around if you rename the tag away.

**To Reproduce**
Steps to reproduce the behavior:

1. Add a ROM whose filename has no region tag (so it shows no region flag).
2. Edit/rename the ROM in the UI and add `(Europe)` to the filename.
3. Save.
4. The EU flag pill is expected to appear, but it doesn't.

**Expected behavior**
After renaming, the flags should reflect the new filename — adding `(Europe)`
should show the EU flag, and removing a region tag should remove the flag — just
like they do for a freshly uploaded `... (Europe)` file.

**Additional context**
Why it happens (for whoever picks this up):

- Filename tag parsing lives in `FSRomsHandler.parse_tags()`.
- The upload/scan path calls `parse_tags()` and saves the results
  (`regions`, `languages`, `version`, `revision`, `tags`) when the ROM is
  created — see `backend/endpoints/sockets/scan.py`.
- The rename path in the `update_rom` endpoint
  (`backend/endpoints/roms/__init__.py`) only updates `fs_name`,
[...]

PR #3472, opened 2026-06-03
test(roms): cover region-tag re-parse on rename (#3471)

Spinnich:
**Description**

Adds regression coverage for region-tag re-parsing when a ROM is renamed via `PUT /api/roms/{id}` (issue #3471) and complements the existing PR by Zurdi (#3295).

When a ROM's `fs_name` changes, the update endpoint re-parses the filename so region / language / revision / version / tags stay in sync with the new name. There was no test asserting the specific case where an **untagged** ROM is renamed to **add** a region — e.g. `test_rom.zip` → `test_rom (Europe).zip` should surface `regions: ["Europe"]`. This PR adds that test.

This is a **test-only** change — no production code is modified. It locks in the existing behavior of the re-parse path (`endpoints/roms/__init__.py` → `FSRomsHandler.parse_tags` → the `REGIONS` table in `handler/filesystem/base_handler.py`) so a future regression in region detection on rename is caught.

It complements the existing `test_update_rom_reparses_tags_on_fs_name_change` test (which covers a rename carrying languages + revision but no region); this new test covers the untagged → region-added transition and explicitly asserts the `rom.regions == []` precondition.

**Files modified**

- `backend/tests/endpoints/roms/test_rom.py` — adds `test_update_rom_adds_region_tag_on_rename`. Mocks `FSRomsHandler.rename_fs_rom` and `IGDBHandler.get_rom_by_id` (no FS writes, no live metadata calls), PUTs `fs_name="test_rom (Europe).zip"`, and asserts a `200` with `fs_name` updated and `regions == ["Europe"]`.

**Testing notes**

```sh
cd backend
uv run pytest tests/endpoints/roms/test_rom.py -k region -vv
```

- New test passes; full `test_rom.py` file is green (43 passed locally).
- No cassettes added or changed — IGDB is stubbed with `IGDBRom(igdb_id=None)`, so the test makes no network calls.

**Reviewer notes**

- Test-only, single-concern, additive; safe to merge once green in CI.
- The stacked `@patch.object` decorators bind bottom-up, so `rename_fs_rom_mock` / `get_rom_by_id_mock` parameter names are swapped relative to the mocks they receive. This is harmless here (neither is referenced by name in the body) and is consistent with the existing sibling tests, but is called out for transparency in case the maintainers want to tidy the convention separately.

---

> AI assistance disclosure: this PR (test code and description) was written primarily by Claude Code, reviewed by me before submission.

**Checklist**

N/A — backend test-only change.

### DIFF
--- a/backend/tests/endpoints/roms/test_rom.py
+++ b/backend/tests/endpoints/roms/test_rom.py
@@ -128,6 +128,30 @@ def test_update_rom_reparses_tags_on_fs_name_change(
     assert body["tags"] == []
 
 
+@patch.object(FSRomsHandler, "rename_fs_rom")
+@patch.object(IGDBHandler, "get_rom_by_id", return_value=IGDBRom(igdb_id=None))
+def test_update_rom_adds_region_tag_on_rename(
+    rename_fs_rom_mock: AsyncMock,
+    get_rom_by_id_mock: AsyncMock,
+    client: TestClient,
+    access_token: str,
+    rom: Rom,
+):
+    """Renaming an untagged ROM to add ``(Europe)`` surfaces the region (issue #3471)."""
+    assert rom.regions == []
+
+    response = client.put(
+        f"/api/roms/{rom.id}",
+        headers={"Authorization": f"Bearer {access_token}"},
+        data={"fs_name": "test_rom (Europe).zip"},
+    )
+    assert response.status_code == status.HTTP_200_OK
+
+    body = response.json()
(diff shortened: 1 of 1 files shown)


Spinnich:
P.S. This test may be superfluous, I'll leave that to the experts! It simply came up during my typical AI-based TDD approach when I started looking into issue #3471 and hadn't realized Zurdi submitted a PR fix yet.

gantoine (approved):

PR #3482, opened 2026-06-06
feat(playmatch): add ratelimiting

DevYukine:
**Description**
Add client-side rate limiting for Playmatch so RomM stays at or below 4 req/s, the limit the Playmatch API enforces per IP. During a scan RomM fans out one lookup per ROM up to `SCAN_WORKERS`, so it can burst past that today (the existing `429` sleep-and-retry in other handlers only reacts after you're already throttled).

Implementation done with a new async Ratelimiter class which prevents going over X (in playmatch case 4) requests per seocnd.

We should consider adding this to other metadata handlers as well because right now there is no proper ratelimiting in RomM so when users have a high `SCAN_WORKERS` env value set we would hit ratelimits on other metadata handlers such as IGDB and or ScreenScraper as well. This is out of scope of this PR but if wanted i can make a follow up PR for metadata handlers such as IGDB where it makes sense (ScreenScraper for example would need a different Ratelimiter as that one is concurrent request ratelimited and not req/s ratelimited)

**Disclosure**
I am not a Python developer and am not very familiar with the RomM codebase. I used Claude to analyze the existing Playmatch integration and to help implement the rate limiter and the lookup hardening. I reviewed the result before submitting.

**Checklist**

Backend change only, no screenshots needed

### DIFF
--- a/backend/handler/metadata/playmatch_handler.py
+++ b/backend/handler/metadata/playmatch_handler.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 from enum import Enum
 from typing import Final, NotRequired, TypedDict
@@ -12,6 +13,11 @@ from logger.logger import log
 from models.rom import Rom, RomFile
 from utils import get_version
 from utils.context import ctx_httpx_client
+from utils.rate_limiter import RateLimiter
+
+# Playmatch caps clients at 4 req/s per IP
+PLAYMATCH_MAX_REQUESTS_PER_SECOND: Final[float] = 4
+_rate_limiter = RateLimiter(PLAYMATCH_MAX_REQUESTS_PER_SECOND)
 
 
 class PlaymatchProvider(str, Enum):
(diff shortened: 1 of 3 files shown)


backend/handler/metadata/playmatch_handler.py
```diff
@@ -144,29 +150,47 @@ async def _request(self, url: str, query: dict) -> dict:
...
-        except (httpx.HTTPStatusError, httpx.ConnectError, httpx.ReadTimeout) as exc:
-            log.warning("Connection error: can't connect to Playmatch", exc_info=True)
-            raise HTTPException(
-                status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-                detail="Can't connect to Playmatch, check your internet connection",
-            ) from exc
-        except json.JSONDecodeError as exc:
-            log.error("Error decoding JSON response from Playmatch: %s", exc)
-            return {}
+        for attempt in range(2):
```
gantoine: move `2` out into a cosnt like `PLAYMATCH_MAX_REQUEST_ATTEMPTS`

gantoine (changes requested):

gantoine (approved):

PR #3490, opened 2026-06-07
Add opt-in files/siblings expansion to GET /api/roms

tmgast:
## Description

#3425 removed `files` from the gallery-list response (`SimpleRomSchema`), leaving `files`/`siblings` detail-only. Clients that build a local cache from the paginated `GET /api/roms` lost the data they use for downloading, multi-file/variant enumeration, and cache building; refetching each rom via the detail endpoint adds N round-trips per sync.

This adds two opt-in boolean params to `GET /api/roms`, matching the endpoint's existing `with_*` convention:

- `with_files=true` — include each rom's file entries
- `with_siblings=true` — include each rom's sibling roms

Default (both omitted) is unchanged — the lean gallery payload is preserved.

### Changes

- **`with_files` / `with_siblings`** query params, default `false`.
- **Serialization-only** — the list query already eager-loads `Rom.files` and `Rom.sibling_roms`, so no query/handler changes. `SimpleRomSchema.files` binds via a validation alias so the relationship is serialized only when requested (and the `lazy="raise"` relationship is never touched in factory contexts). Omitted → empty arrays.
- **`SiblingRomSchema.from_rom`** helper shared by `Simple`/`DetailedRomSchema`.

### Tests

- Default omits both; `with_files` populates files only; `with_siblings` populates siblings (via the `sibling_roms` view) only.

---

> AI Disclosure

**Checklist**

### DIFF
--- a/backend/endpoints/responses/rom.py
+++ b/backend/endpoints/responses/rom.py
@@ -301,6 +301,17 @@ class RomSchema(BaseModel):
     merged_screenshots: list[str]
     merged_ra_metadata: RomRAMetadata | None
 
+    files: list[RomFileSchema]
+    siblings: list[SiblingRomSchema]
+
+    @field_validator("files")
+    def sort_files(cls, v: list[RomFileSchema]) -> list[RomFileSchema]:
+        return sorted(v, key=lambda x: x.file_name)
+
+    @field_validator("siblings")
+    def sort_siblings(cls, v: list[SiblingRomSchema]) -> list[SiblingRomSchema]:
+        return sorted(v, key=lambda x: x.sort_comparator)
+
     @classmethod
     def populate_properties(cls, db_rom: Rom, request: Request) -> Rom:
         db_rom.rom_user = RomUserSchema.for_user(request.user.id, db_rom)  # type: ignore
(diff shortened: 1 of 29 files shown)


backend/endpoints/responses/rom.py
```diff
@@ -335,16 +336,50 @@ def sort_comparator(self) -> str:
...
+            fs_name_no_ext=rom.fs_name_no_ext,
+        )
+
 
 class SimpleRomSchema(RomSchema):
     sibling_ids: list[int]
+    # `files` binds to a dedicated attribute via validation alias rather than the
+    # `Rom.files` relationship: that keeps the full file list out of the default
+    # gallery payload, and avoids tripping the relationship's lazy="raise" in
+    # factory contexts where files aren't loaded. Populated only when requested.
```
gantoine: remove this comment

backend/endpoints/roms/__init__.py
```diff
@@ -571,15 +579,28 @@ def get_roms(
         rom_id_index = session.scalars(query.with_only_columns(Rom.id)).all()  # type: ignore
 
         def _transform(items: Sequence[Rom]) -> list[SimpleRomSchema]:
+            rom_ids = [i.id for i in items]
             sibling_ids_by_rom = db_rom_handler.get_sibling_ids_for_roms(
-                [i.id for i in items], session=session
+                rom_ids, session=session
+            )
```
gantoine: if we do add siblings do we still need sibling_ids?

gantoine (changes requested):

backend/endpoints/roms/__init__.py
tmgast: I think it's still necessary since the default response includes the id list and we are only expanding the full sibling data when requested (where your concern on redundancy comes up). The alternative would be to either:
- exclude either field based on the query state (more complex response model)
- drop sibling_ids and always return a shallow sibling object (just the id) instead in the siblings list

ie:
```json
{
   "siblings": [
      { "id": 1 },
      { "id": 5 }
   ]
}
```

We can go whichever way you'd prefer, but my instinct is to keep the more verbose model.

tmgast: Maybe I overlooked the underlying point you were getting at, though...

I'll add a commit to handle fetching the ID or full sibling models based on the query and prevent the double-hit since that could be a pretty severe performance hit. If we wanna restructure the response on top of this, I can.

gantoine (approved):

PR #3498, opened 2026-06-09
Send all top-level file hashes to Hasheous lookup

gantoine:
## Summary

The Hasheous `Lookup/ByHash` endpoint now accepts an array of hash objects instead of a single one. This updates `HasheousHandler.lookup_rom` to send the hashes of **every** top-level file, rather than picking only the largest file and sending its single hash set.

### Changes

- `lookup_rom` now builds a list payload, one entry per qualifying top-level file:
  - Files with a `chd_sha1_hash` send **only** `{"sha1": <chd_sha1_hash>}` — the raw container MD5/CRC won't match what Hasheous indexes for CHDs.
  - Other files send `{"md5", "sha1", "crc"}`.
  - Empty hashes are dropped per file; files with no usable hash are skipped.
  - If no file yields any hash, the request is skipped and the fallback is returned.
- New lowercase key names (`md5`/`sha1`/`crc`) matching the array payload format, replacing the old single-object camelCase keys (`mD5`/`shA1`).
- `_request` signature widened to accept `dict | list | None` for the list payload.

### Tests

- `test_lookup_rom_sends_all_top_level_file_hashes` — verifies the full list payload, chd_sha1_hash-only handling, and that zero-size / hashless files are filtered.
- `test_lookup_rom_skips_request_when_no_hashes` — verifies the API isn't hit when no hashes are available.

### DIFF
--- a/backend/handler/metadata/hasheous_handler.py
+++ b/backend/handler/metadata/hasheous_handler.py
@@ -172,7 +172,7 @@ class HasheousHandler(MetadataHandler):
             )
 
             # Prepare request kwargs
-            request_kwargs = {
+            request_kwargs: dict[str, Any] = {
                 "url": url,
                 "params": params,
                 "headers": {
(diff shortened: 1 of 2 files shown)


michael-j-green:
Regarding:

> New lowercase key names (md5/sha1/crc) matching the array payload format, replacing the old single-object camelCase keys (mD5/shA1).

Although it's (supposed to be) case insensitive, I would stick with the original casing.

PR #3549, opened 2026-06-18
feat: User screenshots support + dropzone refactor

zurdi15:
**Description**
Adding the last piece of the CRUD endpoints of the media section: screenshots.
Users can now upload their own screenshots under the `media` -> `screenshots` subtab, and there are actually **two different kinds** of screenshots that live in two different places:

- **ROM screenshots**: stored inside the ROM's own folder (`<rom_folder>/screenshots/`) as regular rom files. They are shared library content, so everyone who can see the ROM sees them. Only folder-based (multi-file) ROMs can have them, since a single-file ROM has nowhere to put a `screenshots/` folder.
- **User screenshots**: stored per-user under the user's asset folder (`assets/users/<user_hash>/screenshots/<platform>/<rom_id>/`), the same place where the save/state screenshots already live. These are **private by default**, but each one can be toggled **public** so other users can browse them, community style.

Because of that, the screenshots subtab is now split into three sections: **ROM**, **Mine** and **Community** (other users' public shots, read-only). Both upload paths support drag & drop.

On the backend, the per-user `Screenshot` asset gained `is_public` (visibility) and `is_gallery` (so the auto-captured save/state thumbnails never leak into the gallery), plus upload / delete / visibility-toggle endpoints. The ROM response now exposes `all_user_screenshots` (your own + everyone else's public ones), mirroring how notes already work.

**Dropzone component**
Since we already had the drag & drop box copy-pasted in a bunch of places (upload roms, patcher, firmware, media tabs…), I extracted it into a proper UI-library primitive: `RDropzone` (with its stories), and migrated every v2 call site to it. While at it, I also removed the intermediate "fake dropzone that opens a dialog with the real dropzone" step in the firmware tab and put the real dropzone directly inline.

**Checklist**

image

image


backend/alembic/versions/0084_roms_search_sort_indexes.py
```diff
@@ -30,7 +30,7 @@
 from utils.database import is_mariadb, is_mysql, is_postgresql
 
 # revision identifiers, used by Alembic.
-revision = "0084_add_roms_search_and_sort_indexes"
+revision = "0084_roms_search_sort_indexes"
```
gantoine: already renamed these on master

backend/endpoints/screenshots.py
```diff
@@ -20,19 +26,35 @@
...
 
+# Image formats every modern browser can decode.
+ALLOWED_SCREENSHOT_EXTENSIONS = frozenset(
+    {".png", ".jpg", ".jpeg", ".webp", ".gif", ".bmp", ".avif"}
+)
+
+
+def _is_allowed_screenshot_file(file_name: str) -> bool:
+    _, ext = os.path.splitext(file_name)
+    return ext.lower() in ALLOWED_SCREENSHOT_EXTENSIONS
```
gantoine: these are defined in multiple places

gantoine (changes requested):

backend/endpoints/screenshots.py
zurdi15: true, I extracted them to a utils

PR #3557, opened 2026-06-19
feat: real-time backend logs view

zurdi15:
## What

A new **Logs** view in the v2 frontend that streams the backend logs in real time, reachable from the **UserMenu** and the **Settings sidebar**. Admin-only.

Previously the only way to read backend logs was the dev terminal or `docker logs` on the deployed container — neither accessible from the UI.

### Producer — `LogStreamHandler`
A `logging.Handler` attached to the `romm` logger ([`backend/logger/logger.py`](backend/logger/logger.py)). Because every process (web app, RQ workers, scheduler, watchers) imports the logger, the stream covers the whole backend. Each record is serialized to `{ts, level, module, message}`, ANSI-stripped + sensitive-key-redacted, then in one Redis pipeline:
- `LPUSH`/`LTRIM` into a capped ring buffer (`romm:logs:buffer`, 1000 lines) for backfill, and
- `PUBLISH` to `romm:logs`.

Failures are swallowed silently — stdout stays the source-of-truth log and a Redis hiccup never raises into the app. nginx logs are out of scope (they don't go through Python logging).

### Forwarder — sync→async bridge
A logging handler is synchronous; Socket.IO emit is async. A single background task started in the app `lifespan` subscribes to `romm:logs` and relays each line to admin clients, using a `write_only` `AsyncRedisManager` — the same emit path scan/sync already use. It's guarded by a Redis lock so that with `WEB_SERVER_CONCURRENCY > 1` exactly one worker forwards (no duplicate lines).

### Socket auth — admin room
The main Socket.IO server had no `connect` handler. This adds one that resolves the session for the handshake (from the ASGI scope, falling back to parsing the session cookie → Redis session, same lookup as `RedisSessionMiddleware`) and, **only if the user is an admin**, joins them to the `admin` room. It never rejects a connection, so existing scan/sync sockets are unaffected for everyone. Log lines are emitted to that room only — non-admins never receive them.

### Backfill — REST
`GET /api/logs?limit=N` ([`backend/endpoints/logs.py`](backend/endpoints/logs.py)) returns the recent buffer, gated by the `users.write` scope. The view backfills on open, then streams live.

### Frontend
- View at [`Settings/Logs.vue`](frontend/src/v2/views/Settings/Logs.vue): backfill via REST, live via `useSocketEvent("logs:entry")`, windowed with `RVirtualScroller`. State is view-scoped/ephemeral (capped in-memory ring buffer, no store).
- Auto-tail that pauses on scroll-up with a smooth "jump to latest"; level filter, search, pause/resume, copy, download, and click-a-row-to-copy.
- Full-height panel that scrolls internally (additive `fill` route-meta mode on `SettingsLayout`).
- Admin-gated route + nav entries (`useCan("app.admin")` / `users.write`).

### `RTooltip` (additive, backward-compatible)
Three reusable improvements used by the log rows: `overflow-wrap: anywhere` (long content never overflows), a `maxWidth` prop, and `hint` + `hintIcon` props (the "Click to copy" affordance line).

## Verification
- `typecheck`, `eslint`, `i18n` check, and all 411 frontend tests pass; backend `mypy` clean on touched files.
- New i18n keys added to all locale dirs.

image


backend/endpoints/sockets/logs.py
```diff
@@ -0,0 +1,185 @@
...
+from logger.log_stream_handler import LOG_BUFFER_KEY, LOG_CHANNEL
+from logger.logger import log
+from models.user import Role
+from utils import json_module
+
+ADMIN_ROOM: Final = "admin"
+FORWARDER_LOCK_KEY: Final = "romm:logs:forwarder"
+FORWARDER_LOCK_TTL: Final = 30  # seconds
+# Session cookie name configured on RedisSessionMiddleware in main.py.
+SESSION_COOKIE_NAME: Final = "romm_session"
```
gantoine: instead of redefining it, import it

backend/endpoints/sockets/logs.py
```diff
@@ -0,0 +1,185 @@
...
+    if morsel is None:
+        return {}
+
+    session_data = await async_cache.get(f"session:{morsel.value}")
+    if not session_data:
+        return {}
+    try:
+        return json_module.loads(session_data)
+    except Exception:  # noqa: BLE001 - malformed session is "no session"
+        return {}
```
gantoine: this feels like it should be in an auth util

backend/endpoints/logs.py
```diff
@@ -0,0 +1,35 @@
...
+    prefix="/logs",
+    tags=["logs"],
+)
+
+# Largest backfill the buffer holds — derived from the producer's ring buffer
+# so the two never drift.
+MAX_LOG_LIMIT = LOG_BUFFER_SIZE
+
+
+@protected_route(router.get, "", [Scope.USERS_WRITE])
```
gantoine: we probably need a new `Scope` for this permission

backend/logger/log_stream_handler.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        # The redaction regex lives in handler.metadata.base_handler, which
+        # pulls a heavy import chain (and, during early boot, a partially
+        # initialized handler.redis_handler). Resolve it lazily and cache it
+        # once available; until then, skip redaction — exactly like the
+        # Formatter does for the first few boot lines.
+        self._redactor: re.Pattern[str] | None = None
```
gantoine: you could/should move it to a util instead

gantoine (changes requested):
🔥 
is any of this gated behind and env var? some servers _might_ have admins who aren't server owners and shouldn't have perms to view logs.

zurdi15:
> 🔥 is any of this gated behind and env var? some servers _might_ have admins who aren't server owners and shouldn't have perms to view logs.

Ohh okay didn't thought about it. I just made them visible for all admins but I guess we should put a whitelist env var of users that are allowed to see the logs?

Or just a new toggle/user prop that is `can_see_logs`?

gantoine (approved):

PR #3559, opened 2026-06-20
Run backend tests in parallel with pytest-xdist

gantoine:
## What

Adds `pytest-xdist` and runs the backend test suite across multiple workers (`-n 4` in CI), cutting wall-clock time substantially.

## How it stays isolated

The autouse `clear_database` fixture truncates all tables between tests, so workers can't share one database — one would wipe rows another is mid-test with. Each worker gets its own:

- **Rootdir `backend/conftest.py` (new)** sets a per-worker `DB_NAME` (`romm_test_gw0`, …) before any app module is imported, so each worker's engine binds to its own database. As the rootdir conftest it loads before `tests/conftest.py`.
- **`tests/conftest.py`** creates the per-worker database on demand (mariadb/mysql and postgresql paths) just before migrations run.
- The test user's grant is widened to `*.*` (`setup.sql` + CI workflow) so it can `CREATE DATABASE` for the workers.

Redis needs no equivalent handling — under pytest it's an in-process FakeRedis, so each worker is already isolated.

> Load-bearing detail: pytest-env's `DB_NAME=romm_test` in `pytest.ini` has no `D:` prefix, so it only sets the var when unset — the rootdir conftest sets it first, so the per-worker value wins.

## Verification

- Full suite under `-n 4`: **1709 passed** in ~70s locally.
- Per-worker databases confirmed created on demand (`romm_test_gw0`–`gw3`).
- mypy clean on the modified files.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Initiate MariaDB database
         if: matrix.db == 'mariadb'
         run: |
-          mysql --host 127.0.0.1 --port ${{ job.services.mariadb.ports['3306'] }} -uroot -ppasswd -e "GRANT ALL PRIVILEGES ON romm_test.* TO 'romm_test'@'%' WITH GRANT OPTION; FLUSH PRIVILEGES;"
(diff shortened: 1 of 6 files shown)


zurdi15:
Yesh it was starting to be a pain to run the full suit lol

zurdi15 (approved):

PR #3565, opened 2026-06-21
feat: opt-in per-platform subfolder scanning

### LINKED ISSUE #2050: [Feature] Support platform level subfolders
It is useful to organize large ROM collections using subdirectories at the platform level, like this:

```
Platform/
   |- Hacks/
   |- Homebrew/
   |- Translations/
        |- rom0.7z
   |- rom1.7z
   |- rom2.7z
```

Romm now supports subfolders at the ROM level, but not at the platform level meaning that ROMs contained in subdirectories are not picked up properly during a scan.

l would like to define a list of platform level subdirectories in the config.yml that Romm would treat as **containing** more ROMs rather than as multifile ROMs themselves.

Reorganizing files to match Romms' currently supported structure is an effective solution for smaller collections. However, this is not viable for large collections that support other systems running different emulation frontends over a network.

**Additional context**
The above directory structure is useful when organizing a large collection because each directorys' contents can be mapped to an individual community-maintained DAT file and updated using purpose built tooling. A useful practice to keep parts of a collection that may change rapidly like 'Translations' and 'Hacks' up to date and de-duplicated.

Additionally, if a collection is supporting multiple frontends it is preferable to be able to configure each frontend to support one directory structure rather than have different structures with duplicate contents per frontend.

androosio:
**Description**

Closes #2050.

Adds **opt-in, per-platform subfolder scanning**. When enabled for a platform, RomM recurses that platform's subfolders and treats each nested file as its own ROM (storing its real path on the entry), instead of collapsing a subfolder into a single multi-file ROM. This lets users keep libraries organized into `Hacks/`, `Translations/`, `Homebrew/`, etc. without flattening them — the original request in #2050, and aligned with @gantoine's "per-system option, keep it simple" direction.

Enable per platform in `config.yml` (default **off**, so existing setups are unaffected):

```yaml
scan:
  subfolders:
    nes: true
    snes: true
```

**Why this matters (the default isn't just incomplete — it's misleading)**

Tested on a real library: a `Hacks/` folder containing 3 different romhacks (two Castlevania hacks + a Chrono Trigger hack) is collapsed into one entry today and hash-matched to *one* of them — the other two are invisible and Play is ambiguous. With this flag enabled the same folder becomes 3 separate, correctly-identified games (cover art, metadata, even translated/Japanese titles).

**Implementation**

- **Per-platform, opt-in** via `scan.subfolders` (`fs_slug -> bool`); platforms left out behave exactly as before. No DB schema change / migration.
- **Path-based identity.** `get_roms_by_fs_name` keys results on full path and `mark_missing_roms` compares full paths, so identically-named files in different subfolders stay distinct. The `(platform_id, fs_name)` index is already non-unique. File resolution (download/delete/hash) uses `rom.fs_path`.
- **Recursion** skips hidden (dot-prefixed) folders and excluded media folders at every level.
- **Frontend:** the Files tab now shows a ROM's on-disk **Location** (click-to-copy), handy for ROMs found inside subfolders. New i18n keys added to all locales.

**Mitigations to reduce surprises when enabling**

- A subfolder containing an `.m3u` playlist is kept whole as a single multi-file ROM (a multi-disc game), so it isn't split into one ROM per disc even on an enabled platform.
- The scan log flags ROMs that became "missing" because a folder is now recursed (previously a single multi-file ROM), so the stale entry is easy to clean up.

**Behavior & limitations** (also documented in `config.example.yml`)

- On an enabled platform every subfolder is recursed; folder-based multi-file ROMs need an `.m3u` (or use flat disc-tagged files) to stay whole.
- Identity is path-based, so moving a ROM between subfolders (or turning the flag back off) reads as remove + re-add — saves/play history/favorites/collection membership won't follow.

**Testing**

- Backend unit tests for recursion (incl. collisions, hidden folders, `.m3u` kept whole), `get_roms_by_fs_name` full-path keying, and `mark_missing_roms`.
- Full backend suite passes locally; frontend typecheck / tests / build / i18n parity all pass.
- Verified end-to-end against a real library with metadata providers: scan → identify → download → UI, with a side-by-side of an opted-in vs default platform.

**Note on scope:** intentionally narrow vs the custom-format/template system @thekiefs is working on — a small, low-risk step that closes the original ask here and shouldn't conflict with the larger effort. Happy to adjust or fold it in.

> **AI disclosure:** this PR was written primarily with Claude Code (AI assistance) — code, tests, and docs — driven and reviewed by me.

**Checklist**

### Before — default behavior (a folder = one game)

A `Hacks/` folder that actually contains **3 different romhacks** (two Castlevania hacks + a Chrono Trigger hack) collapses into a single entry "Hack" — the other two are invisible and Play is ambiguous:

Across the platform it's worse: 4 SNES subfolders each masquerade as one game (e.g. "Hack", "Alcahest", "3 Ninjas Kick Back"), hiding ~15 games total:

### After — `scan.subfolders: { nes: true }`

Same source files, platform opted in: every game from `Hacks/`, `Translations/`, `Homebrew/`, etc. is individually scanned and identified with cover art:

A translated romhack from `Translated (NES)/`, now a first-class identified game — note the on-disk **Location** shown in the Files tab:

### Side by side (one scan, same library)

NES (opted in) → 27 individual games; SNES (default) → 9 collapsed entries:

### DIFF
--- a/backend/config/config_manager.py
+++ b/backend/config/config_manager.py
@@ -132,7 +132,7 @@ class Config:
     SCAN_REGION_PRIORITY: list[str]
     SCAN_LANGUAGE_PRIORITY: list[str]
     SCAN_MEDIA: list[str]
-    SCAN_SUBFOLDERS: dict[str, bool]
+    SCAN_SUBFOLDERS: dict[str, bool | list[str]]
     GAMELIST_MEDIA_THUMBNAIL: MetadataMediaType
     GAMELIST_MEDIA_IMAGE: MetadataMediaType
 
@@ -161,13 +161,23 @@ class Config:
 
         return False
 
-    def should_scan_subfolders(self, fs_slug: str) -> bool:
-        """Whether the scanner should recurse into a platform's subfolders,
-        treating each subfolder as a group of ROMs rather than as a single
-        multi-file ROM. Opt-in per platform via `scan.subfolders` in config.yml.
+    def subfolder_scan_spec(self, fs_slug: str) -> bool | frozenset[str]:
+        """How the scanner should recurse into a platform's subfolders.
+
+        Opt-in per platform via `scan.subfolders` in config.yml, where the
+        value is either:
+          - ``True`` -> recurse every subfolder (each becomes a group of roms);
+          - a list of folder names -> recurse only those folders, leaving every
+            other folder as a single multi-file rom (so folder-based multi-file
+            games elsewhere on the platform stay intact);
+          - ``False`` / omitted -> don't recurse (default behavior).
+
+        Returns ``True``, a ``frozenset`` of folder names, or ``False``.
         """
-        subfolders = getattr(self, "SCAN_SUBFOLDERS", {})
-        return bool(subfolders.get(fs_slug, False))
+        value = getattr(self, "SCAN_SUBFOLDERS", {}).get(fs_slug, False)
+        if isinstance(value, list):
+            return frozenset(value)
+        return bool(value)
 
 
 class ConfigManager:
@@ -674,10 +684,15 @@ class ConfigManager:
             log.critical("Invalid config.yml: scan.subfolders must be a dictionary")
             sys.exit(3)
         else:
-            for fs_slug, enabled in self.config.SCAN_SUBFOLDERS.items():
-                if not isinstance(enabled, bool):
+            for fs_slug, value in self.config.SCAN_SUBFOLDERS.items():
+                is_bool = isinstance(value, bool)
+                is_str_list = isinstance(value, list) and all(
+                    isinstance(name, str) for name in value
+                )
+                if not (is_bool or is_str_list):
                     log.critical(
-                        f"Invalid config.yml: scan.subfolders.{fs_slug} must be a boolean"
+                        f"Invalid config.yml: scan.subfolders.{fs_slug} must be a "
+                        "boolean or a list of folder names"
                     )
                     sys.exit(3)
 
--- a/backend/config/config_manager.py
+++ b/backend/config/config_manager.py
@@ -132,6 +132,7 @@ class Config:
     SCAN_REGION_PRIORITY: list[str]
     SCAN_LANGUAGE_PRIORITY: list[str]
     SCAN_MEDIA: list[str]
+    SCAN_SUBFOLDERS: dict[str, bool]
     GAMELIST_MEDIA_THUMBNAIL: MetadataMediaType
     GAMELIST_MEDIA_IMAGE: MetadataMediaType
 
@@ -159,6 +160,14 @@ class Config:
                 return True
 
         return False
+
+    def should_scan_subfolders(self, fs_slug: str) -> bool:
+        """Whether the scanner should recurse into a platform's subfolders,
+        treating each subfolder as a group of ROMs rather than as a single
+        multi-file ROM. Opt-in per platform via `scan.subfolders` in config.yml.
+        """
+        subfolders = getattr(self, "SCAN_SUBFOLDERS", {})
+        return bool(subfolders.get(fs_slug, False))
 
 
 class ConfigManager:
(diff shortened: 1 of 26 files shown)


zurdi15:
What happen if you have a muli-file game inside one of those folders?

androosio:
Good question — I just checked. Right now a folder inside an enabled platform is only kept whole as one game if it contains an `.m3u` playlist, so a multi-disc game with an `.m3u` survives, but a multi-file game without one (e.g. a `.cue`/`.bin` pair) would currently be split into separate ROMs:

```
Hacks/Final Fantasy VII/ (.m3u)  -> 1 ROM ✓
Hacks/Some CD Game/ (cue+bin)    -> split ✗
```

I can extend the "keep whole" check to recognize disc/playlist descriptors (`.m3u`, `.cue`, `.gdi`, `.ccd`) instead of just `.m3u`, which preserves cue/bin, GDI and multi-disc folders while still recursing genuine groups of standalone games. That covers descriptor-based multi-file games; folder-based formats with no descriptor (some console dump layouts) still can't be auto-distinguished from a group, so those would be expected to sit outside grouping folders. Does that approach match what you'd want, or would you prefer a different rule? Happy to implement it.

androosio:
Following up — I checked against the [folder-structure docs](https://docs.romm.app/latest/getting-started/folder-structure/), and you're right to push on this. A documented multi-file game is just a marker-less folder of disc files:

```
roms/ps/
├─ game_5/
│  ├─ game_5_cd_1.iso
│  └─ game_5_cd_2.iso
└─ game_6.iso
```

That's structurally identical to a *grouping* folder (`Hacks/hack1.zip, hack2.zip`), so "recurse every folder" can't tell them apart — it splits `game_5/` into two ROMs. And since there's no `.m3u`/`.cue` in the documented layout, even a descriptor check wouldn't reliably keep it whole. So my earlier `.m3u`-based answer doesn't fully cover it — thanks for catching that.

The only thing that disambiguates reliably is naming the grouping folders (the original ask in this issue). So I'd propose supporting **both** forms, letting the value be a bool *or* a list:

```yaml
scan:
  subfolders:
    # recurse every subfolder — zero-maintenance, but folder-style
    # multi-file games on this platform get split (use when your
    # multi-disc games are flat / .m3u, not folders)
    nes: true

    # recurse only these folders; every other folder stays a normal
    # multi-file game, so documented multi-disc folders are preserved
    psx: ["Hacks", "Translations"]
```

- `true` → for libraries whose multi-file games are flat (e.g. `.m3u` playlists at the platform root), with a documented caveat.
- list → coexists cleanly with folder-style multi-file games, since anything not named follows RomM's normal "folder = one game" convention.

Does that direction work for you, or would you rather drop the bool form entirely and only allow the explicit list? Happy to build whichever you prefer.

> *Reply drafted with Claude Code (AI assistance), reviewed by me — per the disclosure on the PR.*

androosio:
I made some updates I'm happy with in my latest commit — still testing against my own library, but the design feels right now.

**Summary of what changed:** `scan.subfolders` now accepts a **bool *or* a list** per platform. `true` recurses every subfolder; a list recurses only the named folders and leaves everything else as a normal multi-file game. This directly resolves the multi-file question you raised — details below.

**A multi-file game inside a recursed folder:**

1. **Has an `.m3u` playlist** → kept whole as one multi-disc ROM; recursion stops there. Safe under either form.
2. **Marker-less folder of disc files** (RomM's documented `game_5/cd_1.iso, cd_2.iso` layout) → under `true` it **gets split**, since it's structurally identical to a grouping folder. Under the **list form** it stays intact, because only named folders are recursed; everything else follows the normal "folder = one game" rule.

So `true` can't safely coexist with folder-style multi-file games on the same platform — which is why I added the list form.

```yaml
scan:
  subfolders:
    # recurse every subfolder. Safe when this platform's
    # multi-file games are flat / .m3u, not folders.
    nes: true

    # recurse only these folders; every other folder stays a
    # normal multi-file game, so documented multi-disc folders
    # (game_5/cd1,cd2) are preserved.
    psx: ["Hacks", "Translations"]
```

The two forms cover different platforms: `true` for ones that are purely grouping folders (zero maintenance), the list for ones that mix grouping folders with folder-style multi-file games.

Default is still off, no schema migration, identity stays path-based. Kept config-only by design — it matches the original request and the "per-system, keep it simple" direction. A per-platform UI toggle would be a cleaner follow-up than more config, if wanted later.

*Reply drafted with Claude Code (AI assistance), reviewed by me — per the disclosure on the PR.*

PR #3568, opened 2026-06-21
fix: guard SiblingBadge against missing sibling_roms (scan crash with Group ROMs)

### LINKED ISSUE #3567: [Bug] v2 gallery crashes during scan when "Group ROMs" is enabled (SiblingBadge reads undefined sibling_roms)
**RomM version**
Current `master` (unreleased, post-4.9.2). Introduced by [`bd4bf4f4`](https://github.com/rommapp/romm/commit/bd4bf4f40633769460d83082036846471132faa0) ("feat: enhance collection and game details UI") — `SiblingBadge.vue` doesn't exist in 4.9.2, so released versions are unaffected.

**Describe the bug**
On the v2 UI, with the **"Group ROMs"** setting enabled, running a scan while viewing a v2 surface that shows the scanned games (gallery / Home recently-added / Scan view) throws a render-time `TypeError` for every affected card, flooding the console and making the UI janky:

```
TypeError: Cannot read properties of undefined (reading 'length')
 > ComputedRefImpl.fn  src/v2/components/GameCard/SiblingBadge.vue:46:59
    46 |  () => groupRoms.value === true && props.rom.sibling_roms.length > 0,
       |                                                          ^
 > Proxy._sfc_render  src/v2/components/GameCard/SiblingBadge.vue:88:10
 > renderComponentRoot -> componentUpdateFn -> ... -> <Home> -> <RouterView name="v2">
```

**To Reproduce**
1. Use the v2 UI; enable **Settings -> User Interface -> Gallery -> "Group ROMs"**.
2. Open a platform gallery (or Home) with DevTools Console open.
3. Run a scan that re-identifies roms (a Complete scan, or any scan that streams `scan:scanning_rom` events).
4. As roms stream in -> console fills with the `TypeError` above; affected cards fail to render.

**A/B (the toggle is the gate)** — same scan, same roms streaming:
- Group ROMs **ON** -> `TypeError` storm.
- Group ROMs **OFF** -> 0 errors, clean.

**Expected behavior**
Scanning with "Group ROMs" enabled should not crash gallery cards; `SiblingBadge` should tolerate a rom whose `sibling_roms` isn't populated.

**Root cause**
`SiblingBadge.vue` reads `props.rom.sibling_roms.length` / `.map(...)` (lines 46, 49, 68) without guarding. But the scan socket emit — `backend/endpoints/sockets/scan.py` (`scan:scanning_rom`, ~L355 & L483) — does `SimpleRomSchema...model_dump(exclude={..., "sibling_roms"})`, so the emitted rom has **no `sibling_roms` key** (`undefined`). `SimpleRomSchema` types `sibling_roms` as required, so TypeScript doesn't catch the unguarded access; it only fails at runtime on socket-fed roms. The list/detail REST endpoints return `sibling_roms: []` (per #3490), so only the **socket path** is affected. Trigger requires all of: Group ROMs on, a live scan streaming roms, and a v2 surface rendering them.

**Server:** any build from current `master`.
**Client:** any browser (v2 UI).

**Additional context**
Reproduced on a local `master` build with a clean on/off A/B (captured stack + logs). Proposed fix: guard `SiblingBadge` against an absent `sibling_roms`. PR to follow.

> _Investigation and write-up done with Claude Code (AI assistance), verified by me._

androosio:
**Description**

Fixes a v2 gallery crash during scans when **"Group ROMs"** is enabled (#3567).

`SiblingBadge` reads `props.rom.sibling_roms.length` / `.map`, but the `scan:scanning_rom` socket emit strips `sibling_roms` from the payload (`backend/endpoints/sockets/scan.py` — `SimpleRomSchema...model_dump(exclude={..., "sibling_roms"})`). So a freshly-scanned rom reaches the component with the field `undefined`, even though `SimpleRomSchema` types it as required — TypeScript can't catch it, and it throws `TypeError: Cannot read properties of undefined (reading 'length')` at runtime, crashing every affected gallery card while a scan streams roms.

REST payloads always include `sibling_roms` (possibly `[]`), so only the socket path is affected.

**Fix:** default `sibling_roms` to `[]` via a `siblings` computed and read through it.

Closes #3567

**Checklist**

> **AI disclosure:** this fix was written with Claude Code (AI assistance), reviewed by me.

### DIFF
--- a/frontend/src/v2/components/GameCard/SiblingBadge.vue
+++ b/frontend/src/v2/components/GameCard/SiblingBadge.vue
@@ -42,11 +42,18 @@ const { groupRoms } = useUISettings();
 // "pinned" pattern GameActionBtn uses for `more` / `status` menus).
 const menuOpen = ref(false);
 
+// The scan socket (`scan:scanning_rom`) emits roms with `sibling_roms`
+// stripped from the payload, so it can be undefined here even though the
+// type marks it required. Default to an empty list — otherwise reading
+// `.length`/`.map` throws while a scan streams roms and `groupRoms` is on.
+// REST payloads always include it (possibly empty).
+const siblings = computed(() => props.rom.sibling_roms ?? []);
+
 const visible = computed(
-  () => groupRoms.value === true && props.rom.sibling_roms.length > 0,
+  () => groupRoms.value === true && siblings.value.length > 0,
 );
 
-const totalCount = computed(() => props.rom.sibling_roms.length + 1);
+const totalCount = computed(() => siblings.value.length + 1);
 
 const tooltipText = computed(() =>
   t("rom.versions-count", { n: totalCount.value }),
@@ -65,7 +72,7 @@ const versions = computed(() => [
     current: true,
     main: props.rom.rom_user?.is_main_sibling === true,
   },
-  ...props.rom.sibling_roms.map((s) => ({
+  ...siblings.value.map((s) => ({
     id: s.id,
     label: s.fs_name_no_ext,
     current: false,


zurdi15:
What's the ``.spec.ts`` file purpose? We don't have any .spec file in the v2 tree

androosio:
Fair point. It should have been `.test.ts` but even then it would have been the only .test.ts in v2/components.
Also, it's a bit overkill for this PR so I dropped it.

zurdi15 (approved):

PR #3571, opened 2026-06-21
feat: Per-platform custom library structure with non-destructive identity

### LINKED ISSUE #2050: [Feature] Support platform level subfolders

gantoine:
**Closes #2050.** A **library structure** template system, modeled on [Retrom's](https://github.com/JMBeresford/retrom/wiki/Library-Structure#custom), with **content-hash (non-destructive) identity** so reorganizing a library doesn't orphan saves/favorites/history/collections.

> Supersedes the earlier `scan.subfolders` flag (also on this branch), replaced by the template system per maintainer direction.

## What it does

`filesystem.structure` describes the whole library layout, relative to the library root. `{platform}` marks the platform folder and the final `{game}` marks where a game begins: a file there is a game of its own, a folder there is one multi-file game. Any other braced section is a wildcard folder level, purely organizational.

```yaml
filesystem:
  structure:
    default: "roms/{platform}/{game}"   # the shipped default
    firmware: "bios/{platform}"
    ps3: "roms/{platform}/{category}/{game}"   # roms/ps3/Disc/Game/, roms/ps3/PSN/Game/
    snes: "roms/{platform}/{region}/{game}"    # roms/snes/USA/foo.sfc
```

Sections before `{platform}` must be literal folder names, so the folder platforms are enumerated in is a single known path. A per-platform override has to agree with `default` on where the platform folder sits; relocating one platform elsewhere needs platform discovery and the watcher to union several roots, and is not in this PR.

## Breaking changes

- **The `{platform}/roms` layout is no longer auto-detected.** It becomes opt-in via `default: "{platform}/roms/{game}"`. An instance laid out that way **refuses to start** with the exact config to paste, rather than scanning an empty library and marking every ROM missing.
- **`filesystem.roms_folder` and `filesystem.firmware_folder` are gone.** Each named one path segment that a template now contains. Setting either **refuses to start** with the equivalent template.
- **The `/api/setup/library` response** replaces the `detected_structure` (`struct_a`/`struct_b`/`null`) discriminator with `library_ready: boolean` plus `library_structure: string`, so the setup wizard shows the layout actually in force instead of guessing it client-side. This is the one change that touches the frozen v1 UI (`views/Auth/Setup.vue`); without it v1's wizard reads a field that no longer exists.
- **Adds a DB column.** Migration `0126` adds `roms.full_path_hash` and moves the unique index from `(platform_id, fs_name)` to `(platform_id, full_path_hash)`.

## Template syntax

A `/`-separated sequence of path sections, relative to the library root:

- a **braced** section is a macro; a **bare** section is a literal folder matched exactly (case-sensitively);
- `{platform}` appears exactly once, and every section before it must be literal;
- the **last** section is the terminal `{game}`;
- any other braced section (`{region}`, `{category}`, …) is a wildcard directory level, matching any folder;
- `{library}`, `.` and `..` are rejected with a helpful message.

Hidden (dot-prefixed) folders are never descended into or surfaced. Macro names other than `{platform}`/`{game}` carry no meaning; they document the level for whoever reads the config.

A single `{game}` terminal replaces the earlier `{gameFile}`/`{gameDir}` pair. The information a terminal carried was the **depth**, not the kind, and file-vs-folder at a declared depth is a `stat`. That also removes the need for two templates to describe a level holding both.

### Mixed layouts (list of templates)

A platform whose games sit **both directly in its folder and inside grouping subfolders** needs two depths, so its value may be a **list**; discovery is their **union**, deduplicated by full path:

```yaml
filesystem:
  structure:
    nes:
      - "roms/{platform}/{game}"
      - "roms/{platform}/{category}/{game}"   # Hacks/, Translations/, …
```

Where the two disagree about a folder (a game to one, a grouping level to the other), the folder is treated as grouping, decided by what is actually on disk rather than by template order.

## Non-destructive identity

Identity is reconciled by **content hash** during the scan: moving or renaming a game within the structure is matched against a now-missing ROM and **relocated in place**, so its saves, play history, favorites and collection membership follow it. Falls back to path identity when hashing is unavailable (`skip_hash_calculation` / non-hashable platform).

## Frontend

The Files tab shows a ROM's on-disk **Location** (click-to-copy), useful for nested layouts, as a shared `LocationChip` mirroring `HashChip`'s `RTag`-based pattern. The setup wizard's banner renders the configured template verbatim. New and changed i18n keys in all locales; parity and sort checks pass.

## Implementation

- `config`: `parse_structure_template` / `parse_firmware_template` / `parse_platform_templates`, validated at load. `Config.default_structure`, `firmware_structure`, `platform_structure()` and `platforms_dir` resolve them; two startup guards refuse the retired settings.
- `fs handler`: `_discover_structured_roms` walks literal/wildcard levels to the terminal, collecting files and folders alike; `_collect_fs_roms` unions a platform's templates and drops folders another template descends through. Each ROM records its real `fs_path`.
- `db`: ROM lookups and `mark_missing_roms` key on full path so identically-named files in different folders stay distinct; `get_roms_for_relocation` for hash matching.
- Platform discovery, the firmware folder, the filesystem watcher and the Pegasus/gamelist exporters all read the template instead of the dropped A/B detection.

## Testing

- Backend: template and firmware parsers (valid and invalid, string and list forms), the retired-setting and retired-layout guards, per-platform config loading, structured discovery (wildcard/literal levels, depth > 1, hidden-folder skip, cross-folder collisions, mixed loose-plus-grouped union, grouping-conflict resolution), upload-path derivation, full-path keying, digest derivation and hash-based relocation.
- Frontend: `vue-tsc`, eslint, full `vitest`, `npm run build`, and i18n parity/sort all pass.
- The setup wizard was driven in a browser against a running instance across four payloads (default, platform-first, a custom nested template, and an empty library), in a 320px to 1440px sweep. Light theme is unverified: the wizard card renders dark regardless of theme, which predates this PR.

> **AI disclosure:** authored with Claude Code, driven and reviewed by the maintainer.


androosio:
Hey @gantoine, the hash identity is clearly a better approach than mine 👍 However, the templating is a bit too restrictive in its current form and doesn't take into account libraries that contain roms in the main platform folder and 'grouping' subfolders below that. I'm sure I'm not the only one with a setup like this, e.g.

```
/nes
  game 01
  game 02
  Hacks subfolder
    game 03
    game 04
```

templates:
- `{gameFile}` > the loose top-level games, but nothing inside the grouping folders.
- `{category}/{gameFile}` > the grouping folders' contents, but every loose top-level game is dropped. 

With the template approach, I'd have to add at least one subfolder to each platform containing just the top-level roms

My `scan.subfolders: true` in #3565 was the middle-ground option to handle libraries with my structure. Hoping your implementation can keep that flexibility.

gantoine:
@androosio check "Mixed layouts" section and lmk if that works for ya?

androosio:
> @androosio check "Mixed layouts" section and lmk if that works for ya?

Perfect! thanks very much.

thekiefs:
@gantoine thanks much for pushing this out way faster than I could get to it! I started something here: https://github.com/thekiefs/romm/tree/feature/dynamic-libraries ... but I think your hash approach is much more elegant :) 

Slight note, I was designing for two things I don't think this PR covers. 

1. this looks like it scopes templates to within a platform's ROM folder. RomM still resolves the library root → roms_folder → platform directory itself, using the existing conventions. I see {library} and {platform} aren't valid template variables. 
2. this also doesn't cover multiple libraries on different root disks - but I realize this is probably a rarely needed feature

For 1, the path it can handle looks like:
`{LIBRARY_BASE_PATH}/roms/Atari - 2600/{whatever the template describes}`

But it can't handle:
`{LIBRARY_BASE_PATH}/data/games/Atari - 2600`

`/roms/` is what every rommapper uses today and I understand you'd like to keep existing path resolution, but this might also cause new users some heartburn - especially those with large libraries and existing setups/applications built around their existing directory.

Instead, could you offer a per-platform override that looks something like this:

```
filesystem:
  structure:
     atari-2600:
      path: "/data/games/Atari - 2600"
```

gantoine:
@thekiefs totally reasonable, will implement some version of that! (when i circle back to this after 5.0)

NotaInutilis:
Adding 2 cents from a QA (i.e. trying to break stuff) side, listing some stuff that I will definitely mess with to find out the limits of the feature when it releases.

## Template syntax

Mixing gameFolder and gameFile in that way
```
filesystem:
  structure:
    nes:
      - "{gameFolder}"
      - "{category}/{gameFile}"
```

Trying to recreate the same behavior as the default scan, but nested within categories
```
filesystem:
  structure:
    nes:
      - "{category}/{gameFolder}"
      - "{category}/{gameFile}"
```

## Non-destructive identity

> A cheap size pre-filter avoids hashing unrelated files.

I'd test that when a move includes compressing/extracting archive files with big compression ratios. For curiosity's sake: is it necessary to fully extract archives to know their content, or do they have an index of the files they contain and their uncompressed size?

> Only runs for platforms with a custom structure.

Testing that reverting a platform from a custom structure to the defaults.

It looks like a real useful feature so an option to manually apply to selected/all platforms, or enable it punctually on a manual scan would be nice.

PR #3573, opened 2026-06-21
feat(v2): render cover art at its natural aspect ratio

gantoine:
## What

Gallery covers no longer get forced into a fixed box ratio. Every card now shares one **height** and varies in **width** to match its cover image's natural aspect — no cropping. The grid becomes a flow-packed, wrapping layout (ragged right edge).

## How

- **`GameCover`** measures the rendered image's natural ratio on load (`naturalWidth/naturalHeight`) and drives its own `aspect-ratio` from it. The style ratio is kept only as a pre-load seed so box-art covers don't jump. Emits the measured ratio.
- **`GameCard`** default (gallery) card: fixed art height, natural width. Size tiers and the hero keep their fixed footprints. Forwards the ratio event (keyed by rom id).
- **Gallery grid** (`useGalleryVirtualItems`): replaced fixed-column chunking with `packFlowRows` — greedily packs same-height / natural-width cards into a row until the next would overflow, then wraps. Ratios are measured client-side, cached by rom id, and re-packed on a debounced version bump. Because every card shares one height, **rows stay uniform height** — so `RVirtualScroller` stays exact-offset and AlphaStrip jumps, scroll restoration, and grid-nav are unchanged.
- **`useResponsiveColumns`** additionally exposes `usableWidth` (additive) for the width-based packer.

No backend changes: aspect ratios are read from the loaded images on the client.

## Trade-offs

- Brief reflow as images stream in — box-art covers are correct immediately (2/3 seed matches), oddly-shaped covers nudge their row once their image loads.

## Verification

- `vue-tsc` clean, eslint 0 errors, `npm run build` succeeds.
- Full Vitest suite green (incl. new `packFlowRows` unit tests).
- ⚠️ Not yet eyeballed in a live browser — worth a manual look at the ragged wrap and the label-width clamp on narrow covers.


frontend/src/v2/components/Gallery/GalleryShell.vue
```diff
@@ -988,9 +1014,18 @@ defineExpose({
...
+   so `nowrap` is safe — a brief over/under-fill during the load transient
+   (before measured ratios settle) just leaves a ragged right edge or a
+   hair of overflow, corrected on the next re-pack. `align-items:
+   flex-start` keeps every card pinned to the same top. The 12px column gap
+   matches the packer's `gap`; the 18px row gap matches the chrome math. */
 .r-v2-shell__row {
-  display: grid;
-  gap: 18px 12px;
+  display: flex;
+  flex-wrap: nowrap;
```
gantoine, line 1000: The row is `display: flex; flex-wrap: nowrap` with cards at `width: auto` and the default `flex-shrink: 1`. The packer guarantees the natural-width sum fits, so normally there's no shrink — but `cardHeight * ratio` is floating-point and the rendered widths are rounded by the browser, so a row that the packer computed as "just fits" can render a hair over `usableWidth` and trigger flex-shrink on every card in it. Because card height is fixed, shrinking the width crops the cover slightly (object-fit). It's sub-pixel and likely imperceptible, but if you'd rather guarantee covers never distort, `flex-shrink: 0` here trades the rare crop for a hair of ragged overflow instead. Non-blocking — just flagging the interaction.

---

frontend/src/v2/components/Gallery/GalleryShell.vue
```diff
@@ -242,26 +242,53 @@ const { groupBy, layout, toolbarPosition } = useGalleryMode();
...
     isBoxartStyle(boxartStyle.value) ? boxartStyle.value : "cover_path",
   ),
 );
 
+// Measured natural cover ratios, keyed by rom id (stable across gallery
+// context switches, so the cache survives navigation). GameCard reports
+// each cover's ratio once its image loads; the flow-packer reads them via
+// `ratioAt(position)`. Updates are batched behind `ratioVersion` so a burst
+// of image loads triggers a single re-pack instead of one per cover.
+const ratioByRomId = new Map<number, number>();
```
gantoine: `ratioByRomId` is keyed by rom id and intentionally survives gallery context switches, but it's never pruned, so it grows for the lifetime of the page as a user browses different platforms/collections. Each entry is just two numbers so the memory cost is trivial even at tens of thousands of ROMs — noting it only so it's a conscious choice rather than an oversight. No action needed unless you want a ceiling.

---

frontend/src/v2/composables/useGalleryVirtualItems/index.ts
```diff
@@ -259,47 +307,64 @@ export function useGalleryVirtualItems(opts: Options) {
...
+          ratioAt,
+        )) {
           items.push({
             kind: "row",
-            key: `row-${range.letter}-${r}`,
-            rowIndex: r,
-            startPosition: rowStart,
-            endPosition: rowEnd,
+            key: `row-${row.start}`,
+            rowIndex: 0,
```
gantoine: Now that rows are variable-length contiguous runs, `rowIndex` is hardcoded to `0` at both build sites and nothing reads it anymore (confirmed no consumers in v2). Consider dropping the field from the `row` item type so a future reader doesn't mistake `0` for a meaningful index. Minor cleanup.

---

gantoine (commented):
Reviewed the natural-aspect-ratio gallery change (the 12 files GitHub attributes to this PR — the much larger local diff is just a stale merge-base picking up the activity/save-visibility work already on the base branch).

This is a clean, well-structured change. Highlights:

- **The uniform-height invariant is preserved correctly.** Every default card shares a fixed art height (`--r-card-art-w / 0.6667`, mirrored by `cardHeight()` in JS), so rows stay equal-height and `RVirtualScroller` keeps exact offsets — AlphaStrip jumps, scroll restoration, and grid-nav don't need to change. Good call anchoring the whole design on that.
- **`packFlowRows` is pure, correct, and unit-tested** (the over-wide-card-gets-its-own-row edge case is handled, and the tests' arithmetic checks out). Nice to see it factored out and tested rather than inlined.
- **The flat-mode letter→row remap** to a single O(rows + letters) forward walk is correct given both lists are ascending, and degrades safely (one-card rows) before the container is measured.
- **The `aspect_ratio` API removal is clean** — I grepped both trees and found zero remaining consumers (v1 or v2), so dropping it from `PlatformSchema` and the update endpoint breaks nothing. The model column + `DEFAULT_COVER_ASPECT_RATIO` + migration 0027 are retained, so the backend is unaffected; they're now dead weight you could drop in a follow-up migration if you want.
- **Reactivity is handled deliberately** — the non-reactive `ratioByRomId` Map plus a debounced `ratioVersion` bump to coalesce a burst of image loads into one re-pack is a sound pattern, and the timer is cleaned up on unmount.

Three minor, non-blocking notes inline: a flex-shrink/sub-pixel interaction on the wrapping row, the unbounded (but trivially small) ratio cache, and the now-dead `rowIndex` field.

The main thing still outstanding is the one the PR already calls out: it hasn't been eyeballed in a live browser. The ragged-wrap behavior, the label-width clamp on narrow covers, and the load-transient reflow are all worth a manual look across both themes and a phone width before merge — the logic is sound but this is a very visual change. LGTM on the code.

---

Issue #3564, opened 2026-06-20
[Feature] Expand OIDC/SSO: add an "Allow Registration" toggle and support multiple providers (GitHub, Google, Microsoft, Discord) at once


Spinnich:
**Is your feature request related to a problem? Please describe.**

RomM's single sign-on (SSO) support currently has two limitations that make it hard to run securely and flexibly in a self-hosted setup. Today RomM uses one generic OpenID Connect (OIDC) provider configured through `OIDC_*` environment variables, registered once in `backend/decorators/auth.py`.

1. **There is no way to stop SSO from automatically creating accounts.** In `OpenIDHandler.get_current_active_user_from_openid_token()` (`backend/handler/auth/base_handler.py`), if no RomM user matches the login email, RomM *always* creates a brand-new account. There is no setting to turn this off. In practice this means **anyone who can log into the identity provider also gets a RomM account automatically** (as a Viewer by default if no role claims are configured). For people who point RomM at a shared company/family login or a large identity provider, this is a real problem — they want SSO for the handful of users who *already* have a RomM account, not open self-registration for everyone in the directory.

2. **Only one login provider can be configured at a time.** RomM registers a single OIDC client named `openid`, so admins must pick *one* provider. They can't offer, say, "Sign in with GitHub" **and** "Sign in with Google" side by side. Worse, popular providers like **GitHub and Discord don't implement standard OIDC at all**, so RomM literally cannot connect to them with its current OIDC-only approach — even though these are some of the most common logins self-hosters already have.

For comparison, the project **Pingvin Share X** (https://github.com/smp46/pingvin-share-x) has solved both of these. Its OAuth module (`backend/src/oauth/`) supports GitHub, Google, Microsoft, Discord, and generic OIDC **all enabled at the same time**, behind a clean provider interface, and includes an explicit "allow registration" switch.

**Describe the solution you'd like**

Two concrete improvements, ideally in this priority order:

1. **An "Allow Registration" toggle for SSO** (e.g. `OIDC_ALLOW_REGISTRATION` / a UI toggle). When **off**, an SSO login that doesn't match an existing RomM user is rejected with a clear "no account, ask an admin" message instead of silently creating one. When **on**, behavior stays as it is today. This lets admins run SSO in a safe "existing users only" mode. Pingvin does exactly this via its `oauth.allowRegistration` config, checked before any account is created.

[...]

PR #3577, opened 2026-06-22
feat(auth): add OIDC_ALLOW_REGISTRATION toggle to gate OIDC auto-registration

Spinnich:
**Description**

### Summary

Adds a new `OIDC_ALLOW_REGISTRATION` environment variable that controls whether a
successful OIDC/SSO login is allowed to **automatically create a new RomM account**.

Today, when a user logs in via OIDC and no RomM account matches their email, RomM
*always* provisions a new account, with no way to turn it off. That means anyone who can
authenticate to the configured identity provider effectively gets a RomM account
(a Viewer by default). This PR makes that behavior configurable so admins can run OIDC
in an "existing users only" mode.

**Behavior**
- `OIDC_ALLOW_REGISTRATION=true` (**default**) → unchanged from today: an unknown email
  auto-creates an account.
- `OIDC_ALLOW_REGISTRATION=false` → an unknown email is rejected with a `403` and a clear
  message; no account is created. **Existing users are unaffected** and continue to log
  in normally.

> [!NOTE]
> The default is `true`, so this change is **backward-compatible** — existing OIDC
> deployments keep auto-provisioning after upgrading. Opting into the stricter
> "existing users only" behavior is an explicit `OIDC_ALLOW_REGISTRATION=false`.

### Related issue

Partially addresses #3564. That issue requests two OIDC/SSO improvements:

1. **An "Allow Registration" toggle** — ✅ implemented by this PR.
2. **Support for multiple, first-class login providers (GitHub, Google, Microsoft,
   Discord) at once** — ❌ not included here; still open for a follow-up.

This PR delivers item 1 only, so it intentionally uses "addresses" rather than a closing
keyword — #3564 should remain open to track the remaining multi-provider work.

### Why

Self-hosters frequently point RomM at a shared or large identity provider and want SSO
for the handful of people who already have accounts — not open self-registration for
everyone in the directory. This makes auto-provisioning a configurable choice while
keeping the current behavior as the default.

### Files changed

| File | Change |
| --- | --- |
| `backend/config/__init__.py` | New `OIDC_ALLOW_REGISTRATION` constant, `safe_str_to_bool(_get_env("OIDC_ALLOW_REGISTRATION", "true"))` — defaults to `true`, same building blocks as the neighboring `OIDC_*` settings. |
| `backend/handler/auth/base_handler.py` | In `OpenIDHandler.get_current_active_user_from_openid_token`, gate account creation: when no user matches the email and registration is disabled, log and raise `HTTPException(403)` instead of creating the user. Mirrors the existing "no roles" 403 in the same method. |
| `backend/tests/handler/auth/test_oidc.py` | Two fixtures (`mock_oidc_allow_registration_enabled` / `_disabled`) plus three tests: unknown user rejected when disabled, unknown user created when enabled, existing user still logs in when disabled. Updated `test_oidc_valid_add_user` to set the toggle explicitly so it doesn't rely on the global default. |
| `env.template` | Documents `OIDC_ALLOW_REGISTRATION=true` in the OpenID Connect section. |
| `docs/BACKEND_ARCHITECTURE.md` | Adds the variable (default `true`) to the OIDC env-var reference table. |

No database migrations, model changes, or API schema changes.

### Testing notes

- OIDC handler tests: **17 passed** (14 existing + 3 new). Built with TDD — verified the
  new tests fail before the implementation and pass after, and re-ran them after changing
  the default to `true`.
- Full backend suite run locally: **1721 passed** (single-process). The toggle only
  affects the OIDC login path, which the OIDC tests cover.
- `mypy` clean on the changed source files (no new type errors).
- Manual: with `OIDC_ENABLED=true` and `OIDC_ALLOW_REGISTRATION=false`, a brand-new IdP
  user is rejected with the 403; an existing user logs in fine. With the default (`true`),
  the new user is created as before.

### Reviewer attention

- **Default is `true` (backward-compatible).** No behavior change on upgrade; the stricter
  mode is opt-in via `OIDC_ALLOW_REGISTRATION=false`. Flagging in case maintainers would
  prefer `false` (secure-by-default) instead — easy to flip.
- The flag is intentionally **not exposed in `/heartbeat`** (it's enforced backend-side),
  so the login page can't yet tailor its messaging — a possible follow-up.
- A rejected login surfaces as a raw JSON `403` from the OIDC callback, consistent with
  the existing "user has no roles" rejection rather than a styled page.
- User-facing docs (docs.romm.app) may want a matching entry beyond `docs/`.

**Checklist**

N/A — backend-only change, no UI.

### DIFF
--- a/backend/handler/auth/base_handler.py
+++ b/backend/handler/auth/base_handler.py
@@ -13,6 +13,7 @@ from starlette.requests import HTTPConnection
 
 from config import (
     INVITE_TOKEN_EXPIRY_SECONDS,
+    OIDC_ALLOW_REGISTRATION,
     OIDC_CLAIM_ROLES,
     OIDC_ENABLED,
     OIDC_ROLE_ADMIN,
@@ -436,6 +437,15 @@ class OpenIDHandler:
 
         user = db_user_handler.get_user_by_email(email)
         if user is None:
+            if not OIDC_ALLOW_REGISTRATION:
+                log.error(
+                    "User with email '%s' not found and OIDC registration is disabled",
+                    hl(email, color=CYAN),
+                )
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail="User registration is disabled. Please contact an administrator to create an account.",
+                )
             log.info(
                 "User with email '%s' not found, creating new user",
                 hl(email, color=CYAN),
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -142,6 +142,9 @@ INVITE_TOKEN_EXPIRY_SECONDS: Final[int] = safe_int(
 # OIDC
 OIDC_ENABLED: Final[bool] = safe_str_to_bool(_get_env("OIDC_ENABLED"))
 OIDC_AUTOLOGIN: Final[bool] = safe_str_to_bool(_get_env("OIDC_AUTOLOGIN"))
+OIDC_ALLOW_REGISTRATION: Final[bool] = safe_str_to_bool(
+    _get_env("OIDC_ALLOW_REGISTRATION", "true")
+)
 OIDC_PROVIDER: Final[str] = _get_env("OIDC_PROVIDER", "")
 OIDC_CLIENT_ID: Final[str] = _get_env("OIDC_CLIENT_ID", "")
 OIDC_CLIENT_SECRET: Final[str] = _get_env("OIDC_CLIENT_SECRET", "")
--- a/backend/tests/handler/auth/test_oidc.py
+++ b/backend/tests/handler/auth/test_oidc.py
@@ -24,6 +24,16 @@ def mock_oidc_enabled(mocker):
 
 
 @pytest.fixture
+def mock_oidc_allow_registration_enabled(mocker):
+    mocker.patch("handler.auth.base_handler.OIDC_ALLOW_REGISTRATION", True)
+
+
+@pytest.fixture
+def mock_oidc_allow_registration_disabled(mocker):
+    mocker.patch("handler.auth.base_handler.OIDC_ALLOW_REGISTRATION", False)
+
+
+@pytest.fixture
 def mock_token():
     return {
         "access_token": "",
@@ -171,6 +181,7 @@ async def test_oidc_valid_token_decoding(
 async def test_oidc_valid_add_user(
     mocker,
     mock_oidc_enabled,
+    mock_oidc_allow_registration_enabled,
     mock_token,
     mock_openid_configuration,
     config_override,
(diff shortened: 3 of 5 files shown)


Spinnich:
P.S. Just to specifically point it out again, I'm not able to test these changes locally as I don't have a great way to test OIDC in my dev. environment. The main change is ~13 lines however and seems straight forward enough giving me enough confidence to submit the PR without having tested it.

gantoine (approved):

backend/handler/auth/base_handler.py
```diff
@@ -436,6 +437,15 @@ async def get_current_active_user_from_openid_token(self, token: Any):
 
         user = db_user_handler.get_user_by_email(email)
         if user is None:
+            if not OIDC_ALLOW_REGISTRATION:
+                log.error(
+                    "User with email '%s' not found and OIDC registration is disabled",
```
gantoine, line 442: Minor/optional: a rejected login when registration is disabled is an expected operational outcome (the admin opted into "existing users only"), not a server fault. `log.error` here will accumulate error-level noise if a large IdP directory has people trying SSO — `log.warning` reads more accurately. That said, this is consistent with the neighboring "user has no roles" rejection just above (which also uses `log.error`), so matching that style is a defensible choice too. Up to you.

---

gantoine (commented):
Reviewed the full change — small, focused, and well-executed.

What I verified:

- **The gate is placed correctly.** It sits exactly at the `user is None` branch in `get_current_active_user_from_openid_token`, before `add_user`, and I confirmed that's the only auto-provisioning site in the OIDC path. The role-claim 403 still runs first, so the ordering is sensible.
- **Backward-compatible default.** `safe_str_to_bool(_get_env("OIDC_ALLOW_REGISTRATION", "true"))` resolves to `True` when the var is unset, so existing deployments keep auto-provisioning after upgrade — matches the PR's claim.
- **Existing users are unaffected** — the gate is inside the `user is None` branch only, and there's a test (`test_oidc_registration_disabled_allows_existing_user`) proving it.
- **Tests are thorough and honest** — the three new cases cover reject-when-disabled, create-when-enabled, and existing-user-when-disabled, each asserting `add_user` was/wasn't called. Nice touch updating the pre-existing `test_oidc_valid_add_user` to set the toggle explicitly rather than leaning on the global default.
- **Docs + env.template** both updated, placed alongside the other `OIDC_*` entries.

One minor, optional inline note about the log level (`error` vs `warning`) for the rejection path — not blocking.

On the secure-by-default question you raised: defaulting to `true` is the right call for a behavior-preserving change like this; flipping the default to `false` would silently lock out auto-provisioned users on upgrade, which is a surprising break. Keeping it opt-in is the safer choice. LGTM.

---

PR #3587, opened 2026-06-24
Permission fix for exported game lists and Pegasus platform slugs mapping

ametis70:
**Description**

This PR improves fixes the permissions on exported game lists and adds a shortname mapping for platforms between RomM and pegasus.

- Use `os.chmod` (644) on `FSHandler._atomic_write` so exported files like `gamelist.xml` or `pegasus.metadata.txt` have read permission for groups and others (before those were 600). This matches the permissions on uploaded roms, and it's useful for mounting an NFS share or when RomM is runs in a container with a user/group that is not the same as the user that will be launching games from EmulationStation or Pegasus.
- Add dict and function that maps RomM platform slugs (UPS values) to canonical Pegasus (collection name, shortname) pairs. Without this, Pegasus can't find the right logos for platforms in its themes when the RomM slug does not match the Pegasus one.

**Checklist**

### DIFF
--- a/backend/handler/filesystem/base_handler.py
+++ b/backend/handler/filesystem/base_handler.py
@@ -244,10 +244,7 @@ class FSHandler:
 
         try:
             yield temp_path
-            # mkstemp creates files with 0600; normalise to 0644 so that
-            # exported files (metadata.pegasus.txt, gamelist.xml, …) are
-            # readable by other users on a shared/mounted volume — matching
-            # the permissions applied to uploaded ROM files.
+            # mkstemp creates files with 0600 permissions
             os.chmod(temp_path, 0o644)
             os.replace(str(temp_path), str(target_path))
 
--- a/backend/handler/filesystem/base_handler.py
+++ b/backend/handler/filesystem/base_handler.py
@@ -244,6 +244,11 @@ class FSHandler:
 
         try:
             yield temp_path
+            # mkstemp creates files with 0600; normalise to 0644 so that
+            # exported files (metadata.pegasus.txt, gamelist.xml, …) are
+            # readable by other users on a shared/mounted volume — matching
+            # the permissions applied to uploaded ROM files.
+            os.chmod(temp_path, 0o644)
             os.replace(str(temp_path), str(target_path))
 
         except Exception:
--- a/backend/utils/pegasus_exporter.py
+++ b/backend/utils/pegasus_exporter.py
@@ -9,6 +9,138 @@ from logger.logger import log
 from models.rom import Rom
 from utils.filesystem import link_or_copy_file
 
+# Map RomM platform slugs (UPS values) to canonical Pegasus (collection name, shortname) pairs.
+# Source: https://www.pegasus-frontend.org/docs/user-guide/meta-files/
+# Only platforms present in the official Pegasus "Common platform names" table are listed.
+# When a platform is not in this dict the exporter falls back to the RomM platform name and slug.
+SLUG_TO_PEGASUS: dict[str, tuple[str, str]] = {
+    # ------------------------------------------------------------------ #
+    # Atari
+    # ------------------------------------------------------------------ #
+    "atari2600": ("Atari 2600", "atari2600"),
+    "atari5200": ("Atari 5200", "atari5200"),
+    "atari7800": ("Atari 7800", "atari7800"),
+    "atari800": ("Atari 800", "atari800"),
+    "atari-jaguar-cd": ("Atari Jaguar CD", "atarijaguarcd"),
+    "jaguar": ("Atari Jaguar", "atarijaguar"),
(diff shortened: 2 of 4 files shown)


backend/utils/pegasus_exporter.py
```diff
@@ -5,10 +5,115 @@
...
+    UPS.NEO_GEO_POCKET: ("Neo Geo Pocket", "ngp"),
+    UPS.NEO_GEO_POCKET_COLOR: ("Neo Geo Pocket Color", "ngpc"),
+    # Commodore / Amiga
+    UPS.AMIGA: ("Amiga", "amiga"),
+    UPS.AMIGA_CD32: ("Amiga CD32", "amigacd32"),
+    UPS.COMMODORE_CDTV: ("Amiga CDTV", "amigacdtv"),
+    UPS.C64: ("Commodore 64", "c64"),
+    # Amstrad / Sharp / other home computers
+    UPS.ACPC: ("Amstrad CPC", "amstradcpc"),
+    UPS.SHARP_X68000: ("Sharp X6800", "x68000"),
```
gantoine: Typo in the collection name: `"Sharp X6800"` is missing a digit — the Pegasus canonical name is **Sharp X68000** (the shortname `x68000` is already correct). Since the whole point of the map is to match Pegasus' theme keys, the display name being off by a character is worth fixing.

---

backend/utils/pegasus_exporter.py
```diff
@@ -31,6 +136,14 @@ class PegasusExporter:
     def __init__(self, local_export: bool = False):
         self.local_export = local_export
 
+    @staticmethod
+    def _resolve_collection(platform: Platform) -> tuple[str, str]:
+        """Return (collection_name, shortname) for a platform."""
+        if platform.slug in SLUG_TO_PEGASUS:
+            return SLUG_TO_PEGASUS[UPS(platform.slug)]
+
+        return (platform.custom_name or platform.name, platform.slug)
```
gantoine, line 145: Optional simplification: the `UPS(...)` round-trip isn't needed. `SLUG_TO_PEGASUS` is keyed by `UniversalPlatformSlug`, which is a `StrEnum`, so its keys hash/compare equal to their plain-string values — a `str` slug looks them up directly. This collapses the membership check + re-conversion into one lookup:

```python
return SLUG_TO_PEGASUS.get(platform.slug) or (platform.custom_name or platform.name, platform.slug)
```

Functionally identical (and avoids the `UPS(platform.slug)` call, which would raise `ValueError` on a custom slug if it ever ran outside the `in` guard). Not a bug as written — just tidier.

---

backend/handler/filesystem/base_handler.py
```diff
@@ -244,6 +244,8 @@ async def _atomic_write(self, target_path: Path):
 
         try:
             yield temp_path
+            # mkstemp creates files with 0600 permissions
+            os.chmod(temp_path, 0o644)
```
gantoine, line 248: Worth being explicit about the blast radius: `_atomic_write` backs `write_file`, whose other callers are RA-asset downloads (`ra_handler`) and uploads, not just the gamelist/Pegasus exporters — so this makes every atomically-written file 0644 (world-readable), not only the export metadata. That's harmless here (none of those are secrets, and it matches the goal), just confirming it's intended.

One small thing: hardcoding `0o644` ignores the process umask, so an operator who deliberately runs with a stricter umask (e.g. wanting group-only `0640`) can no longer get it. Given the NFS/cross-user rationale in the description that's a reasonable trade-off — flagging only so it's a conscious choice. `os` is already imported, so no import change needed.

---

gantoine (commented):
Reviewed both changes — a small, well-scoped PR with good test coverage.

**Permissions fix (`_atomic_write`):** correct and safe. I traced the callers — `_atomic_write` only backs `write_file` (RA assets + the gamelist/Pegasus exporters), none of which are sensitive, so widening to 0644 doesn't expose anything that should stay private. The chmod is correctly applied to the temp file before `os.replace`, so the final file lands with the right mode atomically. Left a note confirming the blast radius and the umask trade-off — neither blocking.

**Pegasus slug mapping:** clean implementation. `_resolve_collection` correctly prefers the canonical Pegasus pair and falls back to `custom_name`/`name` + raw slug, and the three new tests cover the mapped / unmapped-with-custom-name / unmapped-no-custom-name branches well. I verified all 82 `UPS.*` members referenced in the map actually exist in the enum (a typo there would crash the module at import — they all resolve), and confirmed the `gba`-slug `test_multiline_description` still passes since it doesn't assert on the collection name.

Two minor items inline: a typo in the `Sharp X68000` display name, and an optional simplification of the `UPS(...)` round-trip in `_resolve_collection`. Only the typo is worth fixing before merge. Nice change.

---

gantoine (approved):

PR #3588, opened 2026-06-24
Test Build GitHub Action improvements

ametis70:
**Description**

This PR improves the GitHub Action used to create test builds. Previously, forks could not push images because the image name was hardcoded to `rommapp/romm-testing`. This has been updated to use `${{ github.repository_owner }}/romm-testing`, allowing forked repositories to publish images under their own namespace.

Additionally, support for pushing images to GitHub Container Registry (GHCR) has been added. When manually running the workflow, users can now select the target registry from a dropdown menu (Docker Hub, GHCR, or both). Pushing to Docker Hub still requires the appropriate Docker Hub secrets to be configured.

**Checklist**

- [X] I've added unit tests that cover the changes (does not apply)

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -7,10 +7,10 @@ on:
     inputs:
       # trunk-ignore(checkov/CKV_GHA_7)
       branch:
-        description: "Git branch to build"
+        description: "Git branch"
         required: true
       registry:
-        description: "Container registry to push to"
+        description: "Container registry"
         required: true
         default: "ghcr"
         type: choice
@@ -106,7 +106,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           target: full-image
 
-      - name: Comment PR with Docker image link
+      # PR builds always push to GHCR only, so the image link is hardcoded to GHCR.
+      - name: Comment PR with GHCR image link
         if: github.event_name == 'pull_request'
         uses: actions/github-script@v7
         env:
@@ -115,12 +116,9 @@ jobs:
           script: |
             const owner = context.repo.owner;
             const tag = process.env.HEAD_REF;
-            const images = [
-              `ghcr.io/${owner}/romm-testing:${tag}`,
-            ];
             github.rest.issues.updateComment({
               comment_id: ${{ steps.build-comment.outputs.comment-id }},
               owner: owner,
               repo: context.repo.repo,
-              body: `✅ Preview build completed!\n\nDocker image: \`${images[0]}\``
+              body: `✅ Preview build completed!\n\nDocker image: \`ghcr.io/${owner}/romm-testing:${tag}\``
             })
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build.yml
```diff
@@ -69,7 +90,8 @@ jobs:
         uses: docker/metadata-action@v5.8.0
         with:
           images: |
-            name=rommapp/romm-testing
+            ${{ env.USE_GHCR == 'true' && format('name=ghcr.io/{0}/romm-testing', github.repository_owner) || '' }}
+            ${{ env.USE_DOCKERHUB == 'true' && format('name={0}/romm-testing', github.repository_owner) || '' }}
```
gantoine: The Docker Hub namespace here is derived from `github.repository_owner`, but the login below authenticates with `secrets.DOCKER_USERNAME` — those are only guaranteed to line up on the canonical repo (owner `rommapp`, and a token with push access to the `rommapp` Docker Hub org). A fork that selects `dockerhub`/`both` would push to `<their-gh-owner>/romm-testing`, which only works if their Docker Hub account happens to own a namespace matching their GitHub owner name; otherwise the push 403s. Not blocking — Docker Hub is opt-in and needs secrets anyway — but a fork owner whose GH name ≠ DH namespace can't use it. If you wanted to make it robust, keying the Docker Hub namespace off `secrets.DOCKER_USERNAME` (with a fallback) rather than `github.repository_owner` would match the credential being used.

(For GHCR this is a non-issue: `docker/metadata-action` lowercases image names, so mixed-case GitHub owners resolve fine.)

---

gantoine (commented):
Reviewed the workflow change end to end. This is a clean, sensible improvement and the logic holds up:

- **`USE_GHCR` / `USE_DOCKERHUB` gating is correct.** Non-`workflow_dispatch` events (i.e. the labeled-PR path) resolve to GHCR-only, and dispatch honors the dropdown — so PR builds always go to GHCR, which is exactly what the hardcoded comment link assumes. The Docker Hub login is now gated behind `USE_DOCKERHUB`, so forks/repos without DH secrets no longer hit a failing login step. 
- **The fork story works for the `workflow_dispatch` path:** dispatching on a fork sets `github.repository_owner` to the fork owner and pushes to `ghcr.io/<owner>/romm-testing` using the fork's own `GITHUB_TOKEN` (which has `packages: write` on its own namespace). 
- **The conditional `images:` ternary is safe** — `format()` always yields a non-empty string on the true branch, and `docker/metadata-action` skips the empty line produced on the false branch, so a single-registry build still gets exactly one image name.
- **The PR-comment rewrite is correct** and the inline note explaining why the link is hardcoded to GHCR is a nice touch.

One minor, non-blocking note inline about the Docker Hub namespace being derived from `github.repository_owner` rather than the `DOCKER_USERNAME` credential — only matters for a fork opting into Docker Hub whose GH owner name differs from its DH account.

Worth keeping in mind (pre-existing, not introduced here): the trigger is `pull_request`, so a labeled PR *from a fork* still runs with a read-only token and can't push to GHCR or update the comment — the fork benefit is really via `workflow_dispatch`. No change requested; just flagging so expectations match. LGTM.

---

gantoine (approved):

PR #3595, opened 2026-06-25
perf(roms): speed up the gallery/search list endpoint on large libraries

gantoine:
## Problem

On a 100k-rom library, every `/api/roms` list/search request took **4-5 seconds**. The endpoint runs several O(library) computations per request, and the dominant ones were uncached or doing far more work than needed.

All numbers below were measured against a real 100k-rom MariaDB instance.

### 1. Cache `rom_id_index` (the big one)
The full ordered id list backing the gallery's virtual scroll was recomputed on **every** request (running the sibling-dedup window over the whole library) and shipped uncached, even for `limit=1`. It's now memoised for the unscoped library scan under the same versioned cache as the other sidecars (`char_index`/`filter_values`), invalidated by the same version bump on scans/rom writes.

**2815ms → 6.7ms on cache hit**, identical results.

### 2. Slim the sibling-dedup query
The `group_by_meta_id` dedup built its inner derived table from `select(Rom)`, materializing **all ~50 columns including every JSON metadata blob** for 100k rows. It also carried a wide unused `fs_name_no_ext` VARCHAR(450) through the window function's temp table (spilling the sort to disk), plus a pointless inner `ORDER BY` the window re-sorts. Now it selects only the 11 columns the window needs, and only `id` + `row_num` flow downstream.

**2.79s → 0.86s in-app**, identical 92,793 ids, no schema change.

### 3. Rewrite `with_char_index`
Replaced `row_number()` over the entire non-hidden set (full materialization + a second `GROUP BY` filesort) with a per-letter `COUNT(*)` and a Python accumulate over the ~26 resulting rows. Identical output (verified on the real DB), drops a filesort layer.

### 4. Covering index for the `sibling_roms` self-join
New migration `0090_roms_sibling_cover_index` adds `idx_roms_sibling_cover (platform_id, igdb_id, moby_id, ss_id, launchbox_id, ra_id, hasheous_id, tgdb_id, id)`. The view's 7-way metadata-id `OR` can't be index-driven from the joined row, so each parent previously scanned its whole platform and read wide rows; now it resolves from an index-only scan.

**77ms → 9.6ms warm on sibling-dense pages** (`Using index` instead of reading wide rows); far larger improvement cold.

## Net effect
| | Before | After |
|---|---|---|
| Unscoped gallery + virtual scroll | 4-5s/request | ~7ms warm / ~0.86s cold |
| Sibling hydration (dense page) | 77ms warm / ~5s cold | 9.6ms |

Scoped/searched requests (computed live by design) get the same ~3.2x dedup speedup.

## Considered but rejected
Precomputing a `sibling_group_key` column to eliminate the dedup window: profiling showed the partition-key expression costs only ~90ms; the rest is the unavoidable sort by `is_main_sibling`/`fs_name_no_ext`. A stored column would save ~90ms in exchange for a migration, backfill, and perpetual write-path maintenance. Not worth it.

## Testing
- 64 backend tests pass (handler + filter-cache + endpoint), incl. a new `test_get_rom_id_index_hit_matches_miss`.
- Migration `0090` applies/downgrades cleanly; single linear alembic head.
- Before/after timings and identical-result checks run against the live 100k-rom DB.

## AI assistance
Written with Claude Code (Opus 4.8): diagnosis via `EXPLAIN ANALYZE`/profiling, the query rewrites, the migration, and the tests. Reviewed by the author.

### DIFF
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -618,9 +618,21 @@ def get_roms(
         # trunk-ignore(mypy/typeddict-item)
         filter_values = RomFiltersDict(**query_filters)
 
-    # Get all ROM IDs in order for the additional data
+    # The full ordered id list backs virtual scroll, so it's computed over the
+    # whole result set on every request. Memoise the unscoped library scan (same
+    # key scheme as the other sidecars); scoped/searched sets stay live.
+    rom_id_index_cache_key = (
+        f"all:u{request.user.id}"
+        f":o{order_by.lower()}:d{order_dir.lower()}:g{int(group_by_meta_id)}"
+        if is_unscoped
+        else None
+    )
+    rom_id_index = db_rom_handler.get_rom_id_index(
+        query=query, cache_key=rom_id_index_cache_key
+    )
+
+    # Hydrate the requested page and its additional data
     with sync_session.begin() as session:
-        rom_id_index = session.scalars(query.with_only_columns(Rom.id)).all()  # type: ignore
 
         def _transform(items: Sequence[Rom]) -> list[SimpleRomSchema]:
             rom_ids = [i.id for i in items]
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -573,13 +573,37 @@ def get_roms(
         include_file_stats=True,
     )
 
-    # Cache only the unscoped library scan; scoped/searched sets are narrower and computed live.
+    # Cache only the fully unscoped library scan; any narrowing parameter makes
+    # the result set narrower, so it is computed live. The sidecar cache key
+    # encodes only user/order/grouping, not the filters, so every filter applied
+    # to `query` below must gate caching here or a narrowed list leaks under the
+    # shared "all" key. Bool flags use `is not None` since False is an active
+    # filter. Logic operators are omitted: they only matter when their list
+    # filter is set, which is already covered.
     is_unscoped = not (
         search_term
         or platform_ids
         or collection_id
         or virtual_collection_id
         or smart_collection_id
+        or genres
+        or franchises
+        or collections
+        or companies
+        or age_ratings
+        or statuses
+        or regions
+        or languages
+        or player_counts
+        or updated_after
+        or matched is not None
+        or favorite is not None
+        or duplicate is not None
+        or last_played is not None
+        or playable is not None
+        or has_ra is not None
+        or missing is not None
+        or verified is not None
     )
 
     # Get the char index for the roms
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -101,6 +101,27 @@ def safe_int_or_none(value: Any) -> int | None:
     return safe_int(value)
 
 
+def build_unscoped_sidecar_cache_key(
+    user_id: int,
+    order_by: str,
+    order_dir: str,
+    group_by_meta_id: bool,
+    is_unscoped: bool,
+) -> str | None:
+    """Cache key for the unscoped library sidecars (char index, filter values,
+    rom id index). Returns None for scoped/searched sets, which are computed live.
+    The computed values depend on user, ordering and grouping, so all are part
+    of the key.
+    """
+    if not is_unscoped:
+        return None
+
+    return (
+        f"all:u{user_id}"
+        f":o{order_by.lower()}:d{order_dir.lower()}:g{int(group_by_meta_id)}"
+    )
+
+
 class RomUpdateForm(BaseModel):
     igdb_id: str | None = Field(default=None, description="IGDB game ID.")
     sgdb_id: str | None = Field(default=None, description="SteamGridDB game ID.")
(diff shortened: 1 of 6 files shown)


zurdi15 (approved):

zurdi15:
I can feel the speedness manual testing it, great job

backend/handler/database/roms_handler.py
```diff
@@ -1125,45 +1135,64 @@ def with_char_index(
...
-        result = [(letter, int(position)) for letter, position in rows]
+        # Walk the letters in the same direction the client paginates over, so
+        # each letter's offset is the count of rows that sort before it.
+        result: list[tuple[str, int]] = []
+        offset = 0
+        for value, count in session.execute(counts).all():
+            if value is not None:
+                result.append((value, offset))
+            offset += count
+        result.sort(key=lambda entry: entry[0])
```
gantoine, line 1160: The rewrite is a nice simplification, but the equivalence to the old `row_number()` approach rests on two collation assumptions worth a sanity check: (1) sorting by `order_by_attr` groups rows into contiguous first-character blocks (true for prefix-consistent collations), and (2) NULL-letter rows accumulate into the offset at the same position they occupied under the old full-sort `row_number()` — i.e. `ORDER BY letter` places NULLs the same way `ORDER BY order_by_attr` did (NULLs first asc / last desc by default). Both hold for the standard MariaDB/Postgres collations, but a case/accent-insensitive collation or a non-default NULL ordering could shift an offset by the size of the NULL/edge group.

The PR notes this was verified identical on the live DB, which is the real assurance — but the added automated test only covers the trivial single-letter `{"t": 0}` shape. A small fixture-level case with mixed first letters, a NULL `name_sort_key` row, and `order_dir=desc` would lock the equivalence in against future collation/ordering changes.

---

backend/endpoints/roms/__init__.py
```diff
@@ -618,9 +618,21 @@ def get_roms(
...
+    # whole result set on every request. Memoise the unscoped library scan (same
+    # key scheme as the other sidecars); scoped/searched sets stay live.
+    rom_id_index_cache_key = (
+        f"all:u{request.user.id}"
+        f":o{order_by.lower()}:d{order_dir.lower()}:g{int(group_by_meta_id)}"
+        if is_unscoped
+        else None
+    )
+    rom_id_index = db_rom_handler.get_rom_id_index(
+        query=query, cache_key=rom_id_index_cache_key
```
gantoine: **The cache key for `rom_id_index` doesn't capture all the filters baked into `query`, so this can serve/poison wrong results.**

`rom_id_index_cache_key` is gated on `is_unscoped` (defined just above), which only excludes `search_term / platform_ids / collection_id / virtual_collection_id / smart_collection_id`. But `query` (built at line 515 via `filter_roms`) *also* has the metadata filters applied — `genres / franchises / companies / age_ratings / statuses / regions / languages / player_counts` (+ logic operators) — and the frontend does send those to this endpoint (`services/api/rom.ts`). So a filter on the all-games view (e.g. `?genres=…` with no platform/collection/search) is still `is_unscoped == True`.

Because the key (`all:u{id}:o..:d..:g..`) encodes none of those filters:
- a genre/status-filtered request computes a *narrowed* id list and stores it under the shared "all" key, and
- a later unfiltered (or differently-filtered) request reads it back.

Since `rom_id_index` is the full ordered id list backing virtual scroll, the gallery shows the wrong set/count of games whenever a user filters the unscoped view, and two different filters collide on one key. `with_char_index` uses the same `query` + same gate, so it has the same exposure (not introduced here, but this PR extends the pattern to the most correctness-critical sidecar). Fix for both: either fold the active metadata-filter signature into the cache key, or treat those filters like scope/search in `is_unscoped` and compute live. (`filter_values` is fine — it deliberately uses `unfiltered_query`.)

---

gantoine (commented):
Reviewed the four optimizations, the migration, and the cache wiring. The performance work is well-reasoned and mostly clean — but I think there's one cache-correctness issue worth resolving before merge.

**Main concern (inline):** the new `rom_id_index` cache is gated on `is_unscoped`, which doesn't account for the metadata filters (`genres / statuses / regions / languages / franchises / companies / age_ratings / player_counts`) that `filter_roms` bakes into `query` and that the frontend actually sends. A filter applied on the all-games view is still "unscoped", so its narrowed id list gets stored under — or read from — the shared `all:u{id}:…` key. Since `rom_id_index` is the id list backing virtual scroll, that surfaces as the gallery showing the wrong set/count of games. `with_char_index` shares the same gate, so the fix should cover both (fold the filter signature into the key, or disqualify those filters from `is_unscoped`). Note `filter_values` is unaffected since it uses `unfiltered_query`.

**The parts that look solid:**
- **Dedup slimming (#2):** correct — only `id` + `row_num` flow downstream, the partition/order inputs are all still present in the narrowed `base_subquery`, and dropping the inner `ORDER BY` is safe because the window defines its own ordering. The "identical ids" claim holds.
- **`with_char_index` rewrite (#3):** the count-and-accumulate is equivalent to the old `row_number → min` for standard collations; left a note on the NULL-ordering / collation assumptions and that the added test only covers the trivial case.
- **Covering index (#4):** migration `0090` chains cleanly off `0089` (verified single linear head — 0087→0088→0089→0090), and the `__table_args__` declaration matches the migration's index name + columns exactly, so model-driven and migration-driven schemas won't drift. Omitting `flashpoint_id` is consistent with the 7-way OR in the `sibling_roms` view.
- **`get_rom_id_index` caching mechanics:** correctly reuses the versioned-cache scheme, and the version bumps on rom writes / hidden-toggle / scan, so the *unscoped* case invalidates correctly. The new hit==miss + version-invalidation test is a good addition.

Once the `is_unscoped` / filter-key gap is addressed (ideally for `char_index` too), this looks great — the diagnosis and the index work are exactly right.

---

PR #3605, opened 2026-06-25
feat: granular permission system (groups, overrides, per-entity hiding)

zurdi15:
## Summary

Replaces RomM's role + fixed-JWT-scopes authorization with a **DB-backed, per-request** model:

- **Two user kinds only**: `admin` (bypasses everything) and `user` (granular).
- **Permission groups**: a named read/write/delete matrix over entity types, with one configurable **default group** for new users.
- **Per-user overrides**: tri-state grant/revoke on top of the group.
- **Opt-out visibility (denylist)**: an admin can hide specific platforms/games from a **user or a whole group**; hiding a platform cascades to its ROMs/firmware at query time.

`User.oauth_scopes` becomes a coarse projection of the new model, so the existing scope-based enforcement chain, client tokens, OAuth `/token`, kiosk, and the legacy v1 UI keep working. Upgrade is access-preserving: seeded "Viewer (legacy)" / "Editor (legacy)" groups reproduce each old role exactly (including delete, which previously gated on `*_WRITE`).

## What's included

**Backend**
- New models: `permission_groups`, `permission_group_grants`, `user_permission_overrides`, `hidden_entities` (single consolidated migration `0091_permission_system`).
- Resolver + coarse-scope projection (`handler/auth/permissions.py`), enforcement helpers (`assert_can`, `require_permission`, visibility filtering + 404-masking).
- `/permissions/me`, admin CRUD under `/api/permissions` (catalog, groups, user assignment/overrides, hidden entities), `permissions:changed` socket broadcast, `ActionKey` exposed in OpenAPI.
- **Role narrowed to `admin`/`user`** (migration `0092`): VARCHAR-backed enum, `Role.coerce` for legacy strings, group-default fallback for group-less users, OIDC editor/viewer claims map to `user`.

**Frontend**
- v2 admin UI: Permission Groups tab, group form (matrix + color + group-level hiding), merged user editor (admin toggle + group + overrides matrix + hidden platforms/games), colored access pills, shared `permissionGroups` Pinia store.
- `RCheckbox` gains a generic multi-state mode (used by the permission matrices); `RTable` gains a subtle row entrance animation + per-column select-all.
- Hidden platforms/games pickers (icon/cover, multi-select, list view).
- v1 user management adapted to `admin`/`user`. Regenerated types; i18n added across all 18 locales.

## Migration / deploy notes

- Migrations auto-apply on backend startup. `0092` converts the native `role` enum to VARCHAR and normalizes `ADMIN→admin`, `VIEWER/EDITOR→user` (verified on MariaDB).
- No user gains or loses access on upgrade (parity test enforces it).

## Verification

- Backend: full `tests/endpoints` + `tests/handler` suite green on MariaDB (incl. parity, resolver, OIDC, visibility, admin CRUD). Migration conversion probe-tested.
- Frontend: `vue-tsc` typecheck, Vitest + Storybook play tests, Trunk lint, and i18n parity all pass; browser-verified in light + dark themes.

## AI assistance disclosure

This PR was developed with substantial assistance from an AI agent (Claude Code / Claude Opus). The AI did the bulk of the implementation, test updates, and verification under human direction and review. Per `CONTRIBUTING.md`, disclosing this is required.


zurdi15:
## Code review: granular permission system

Reviewed adversarially across four dimensions (auth core/resolver, migrations, endpoint enforcement + query visibility, frontend). The architecture is sound and the parity story (legacy groups reproducing old roles, projection feeding the coarse layer) holds up well. There is, however, **one CI blocker** and a cluster of **visibility "fail-open" gaps** that should be closed before this ships, since opt-out hiding is a core promise of the feature.

Items marked _(verified)_ I reproduced or confirmed against the code/CI directly. Items marked _(reported)_ come from the review pass and are worth a maintainer look but I did not individually reproduce.

---

### Blocker (CI is red)

**1. Backend `pytest` fails on MariaDB and PostgreSQL** at `tests/models/test_user.py::test_admin`. _(verified)_
The admin scope projection returns `sorted(FULL_SCOPES, key=lambda s: s.value)` (alphabetical) at `backend/handler/auth/permissions.py:156`, while the non-admin path uses `order_scopes(...)` (canonical order) and the test asserts list-equality against the `FULL_SCOPES` constant order. Same 20 scopes, different order, so `==` fails. The scope *set* is correct, so this is cosmetic, but it blocks merge.
- Fix: admin branch should `return order_scopes(FULL_SCOPES)` (consistent with the non-admin path), or return the constant directly; alternatively compare as sets in the test.
- Process note: `tests/models/` was outside the local verification sweep (`tests/endpoints` + `tests/handler`), which is why this slipped through. Worth running the full `uv run pytest` before declaring green.

---

### High (security: hidden entities leak / privilege escalation)

**2. Default-group hides fail open for users with no explicit group.** _(verified)_ `backend/handler/auth/permissions.py:127-132`
`_resolve_grant_map` correctly falls back to the default group's *grants* when `user.permission_group_id is None`, but `_resolve_non_admin` passes the **raw** `user.permission_group_id` (None) to `get_hidden_entity_ids`. So a platform/ROM hidden "from the default group" is not hidden from any user without an explicitly assigned group, which is the common case (`add_user` and `set_user_group(..., None)` both leave it NULL). Resolve the default group id before the hidden lookup, mirroring `_resolve_grant_map`.

**3. `/roms/by-hash` and `/roms/by-metadata-provider` bypass the hidden-ROM mask.** _(verified)_ `backend/endpoints/roms/__init__.py:855-915, 924-952`
Every sibling read path (`/{id}`, `/{id}/simple`, content, download) 404-masks via `can_see_rom`, but these two do not. Any `ROMS_READ` user (the default) can dump a hidden ROM's full `DetailedRomSchema` by enumerating a small-integer provider id or supplying a known hash. Add the same `is_authenticated and not can_see_rom(...)` mask.

**4. Granting the `USERS` entity to a non-admin enables admin creation (escalation footgun).** _(verified chain)_ `permissions_map.py:69`, `endpoints/permissions.py:65`, `endpoints/user.py:60-66,101`
The catalog exposes `list(PermEntity)` (including `USERS`/`TASKS`/`LOGS`) as grantable. `(USERS, WRITE)` projects to `Scope.USERS_WRITE`. `add_user` gates on that **scope** (not `assert_admin`) and does `role=Role.coerce(role)`, accepting `role="admin"`. So an admin who grants a user "Users: write" (intending user-list management) inadvertently lets them `POST /users` with `role="admin"` and mint a full admin. Pre-rework `USERS_WRITE` was admin-only. Consider excluding `USERS`/`TASKS`/`LOGS` from delegable entities, or requiring real admin to set `role="admin"` in `add_user`.

---

### Medium (secondary read paths not visibility-filtered; one parity change)

These all leak the existence/metadata/counts of hidden platforms or ROMs to a user they are hidden from (content downloads remain 404-masked, but ids, names, and counts leak). _(all reported)_
- **Library feeds** (`backend/endpoints/feeds.py`, many lines) call `get_platforms()` / `get_roms_scalar()` with no hidden sets; under `DISABLE_DOWNLOAD_ENDPOINT_AUTH` the generated URLs are also usable.
- **Sibling ROMs** (`roms_handler.py:299-352` via `DetailedRomSchema.sibling_roms`) can include a hidden sibling.
- **`/stats`** (`endpoints/stats.py:11-34`) totals include hidden entities and `include_platform_stats=true` leaks hidden platform names.
- **Collection `rom_ids` / `rom_count`** (`responses/collection.py`, `endpoints/collections.py`) expose hidden ROM ids and counts.
- **`search.py` `search_rom`** fetches a ROM by id with no `can_see_rom` check.

**`can_access` ownership short-circuit defeats `granted=False` revokes on owned resources.** `backend/handler/auth/dependencies.py:46-48` _(reported)_ Returns `True` for `owner_id == user_id` before consulting grants, so an admin's explicit revoke of (e.g.) own-delete is unenforceable. Since DELETE projects to no coarse scope, there is no `@protected_route` backstop.

**OIDC `editor` claims and in-flight `EDITOR` invites now provision viewer-level access.** `base_handler.py:428-464`, `endpoints/user.py:204-210` _(reported)_ New OIDC/invite editors get `role=USER` with no group, so they resolve to the default (Viewer) group, losing library-wide write. The migration preserves *existing* editors, so this is a deliberate-looking divergence for newly provisioned ones, but it breaks the "no access change" framing for that path. Confirm it is intended.

---

### Low

- `delete_firmware` (`endpoints/firmware.py:298-303`) and `update_platform` PUT (`endpoints/platform.py:123-140`) do not 404-mask hidden entities, unlike their read/delete siblings. _(reported)_
- Hiding a platform/ROM does not invalidate the per-user unscoped sidecar cache, so hidden ids linger in `total`/`rom_id_index` until TTL. `roms/__init__.py:106-124`. _(reported)_
- `update_permission_group` rename onto an existing name hits the unique constraint as a 500 (create checks first, update does not). `endpoints/permissions.py:120-145`. _(reported)_
- No "last default group" guard: `update_group(is_default=False)` on the sole default leaves no default (fail-closed, but a footgun). _(reported)_
- Kiosk read-only rests entirely on the coarse layer: `resolve_permissions` applies no kiosk cap, so a future mutating route gated only by `assert_can` (no coarse write scope) would be writable in kiosk mode. `permissions.py:96-110`. _(reported)_
- Perf: `oauth_scopes` is now 2-3 DB queries per authenticated non-admin request, and the resolver re-resolves the grant map independently for fine-grained handlers (no per-request sharing). Admins short-circuit. `models/user.py:133-140`. _(reported)_
- A11y: `HiddenGamesPicker.vue:126` search rows are `<li @click>` with no keyboard handler. `npx eslint` reports 2 errors (keyboard/gamepad users cannot add games, violating v2 universal-input). Note: this does **not** fail CI (Trunk Check is green), but it should use a real button / `role=button` + `@keydown`. _(verified)_
- Frontend dead code: `EditUserDialog` `originalRole` is set but never read; `permissionsApi` is an unused export in `services/api/permissions.ts`; `stores/permissions.ts` `hidden` state is write-only (no reader yet). _(reported)_
- `InviteLinkDialog.vue:109` renders role labels via `charAt(0).toUpperCase()` instead of the existing `role-admin`/`role-user` i18n keys, so they stay untranslated. _(reported)_
- `v2/utils/groupColor.ts` hardcodes a hex palette (the one hex-literal spot in v2) and the swatch `aria-label` is the raw hex rather than a name. _(reported)_

### Nits
- Migration legacy-group grant matrices are hand-typed literals duplicating `permissions_map.py` rather than derived from it (kept in sync only by `test_permissions_parity.py`). _(reported)_
- `add_hidden_entity` accepts non-cascading entity types (e.g. `COLLECTIONS`, `FIRMWARE`) that the resolver ignores, and does not validate `entity_id` exists (silent no-op). _(reported)_
- `get_admin_users()` counts disabled admins in the last-admin guards; consider filtering `enabled=True`. _(reported)_
- Em-dashes appear in several new comments (CLAUDE.md forbids them).

---

### Investigated and refuted (so nobody chases these)

- **0092 enum -> VARCHAR needs `postgresql_using` / breaks Postgres.** Refuted. _(verified)_ I reproduced the exact conversion on real Postgres 15: `ALTER TABLE users ALTER COLUMN role TYPE varchar(20)` (no `USING`) succeeds, and the normalization UPDATEs produce `admin`/`user` correctly. Both the dedicated `migrate-postgres` job and the Postgres pytest suite (which runs `alembic upgrade head`) are green. Not a blocker. (Minor: the native `role` enum type is left orphaned after conversion, which only matters for a downgrade round-trip, which is itself best-effort.)
- **HiddenGamesPicker eslint errors break CI lint.** Refuted. Trunk (the actual CI/pre-commit lint gate) passes; it is a real a11y gap but not a CI failure (see Low above).

---

### Solid / verified OK

- Resolver precedence (admin bypass -> per-user override -> group grant -> default group) is correct; `granted=False` overrides `pop()` correctly; group-less users fall back to the default group; missing default group is guarded.
- Projection parity is exact: legacy Viewer/Editor grants are set-equal to `WRITE_SCOPES`/`EDIT_SCOPES`, including the DELETE grants that the old `*_WRITE` scopes implied, so no one loses delete on upgrade and no one gains admin scope.
- Migration chain is a single head (`0090 -> 0091 -> 0092`), revision ids are all <= 32 chars, exactly one `is_default` group is seeded, backfill ordering is correct (groups before assignment), and SQLite/MariaDB/Postgres upgrade paths all run.
- Every `/api/permissions` admin route is `assert_admin`-gated (a non-admin with `USERS_*` scope still cannot read the catalog or mutate groups/overrides/hides); `update_user` has no mass-assignment of `permission_group_id`/`is_admin`; self-demote and last-admin (primary paths) are guarded.
- `get_roms` list totals and filter sidebars are correctly hidden-filtered (counts match the filtered query).
- Frontend: `vue-tsc` typecheck and i18n parity (all 17 non-default locales) pass; `role-map.ts` deleted cleanly with no lingering refs; `RCheckbox`/`RTable` primitive purity intact; the override (inherit/grant/grant_own/revoke) and group (none/full/own) matrices round-trip and emit well-formed payloads; v1 changes are minimal and additive.

Nice work overall. The blocker is a one-line fix; the High visibility gaps are the main thing to close before merge given hiding is a headline feature.

<sub>Review assisted by Claude Code (Claude Opus): four parallel adversarial review agents plus manual verification of the blocker, both High visibility findings, the escalation chain, and the Postgres-migration refutation.</sub>

gantoine (commented):
Reviewed the authorization core in depth — resolver, scope projection, enforcement helpers, per-entity visibility/404-masking, the admin CRUD, and the migrations. (I left the v2 admin UI mostly aside; backend is the source of truth.) This is a well-architected rework and the core is sound; the issues are visibility-coverage gaps on a few endpoints plus one Postgres migration concern.

## Verified correct
- **Resolver is fail-closed** (`handler/auth/permissions.py`): no group / no default group / unknown role / missing overrides all resolve to an empty grant set → every `allows()` returns False. Admin bypass is exact (`user.role == Role.ADMIN`) and a normal `user` never constructs `is_admin=True`.
- **Override precedence** is right: explicit revoke `pop`s the group grant, explicit grant adds. (One sharp edge: a per-user grant override silently replaces the group's `own_only` because override identity is `(entity, action)` without `own_only`; both directions are admin-initiated and the narrowing direction fails closed, so not exploitable — just a granularity limitation worth a comment.)
- **Scope projection is a strict subset** (`permissions_map.py`): `DELETE` projects to *no* coarse scope, so delete is double-gated (coarse `*_WRITE` necessary + authoritative `assert_can(DELETE)`). The legacy parity matrices match the `0091` seed grants exactly.
- **Permission-admin CRUD is admin-only**: every mutating route in `endpoints/permissions.py` calls `assert_admin(request)` in its body, so the coarse `[Scope.USERS_WRITE]` decorator can't be abused by a non-admin who's merely been granted `USERS` write. `assert_admin` correctly 403s unless `perms.is_admin`.
- **Read-path masking is broad and consistent** on the endpoints this PR touched: list (`get_roms` threads `hidden_*` into the query; sidecars/counts/`rom_id_index` derive from the filtered query; sidecar cache key is per-user), single-entity GETs/HEAD/content (roms, platform, firmware), `download_roms`, all `feeds.py` variants, `collections` (incl. fixing `rom_count`), `search`, and `stats` counts via `_exclude_hidden`. Delete paths combine `assert_can(DELETE)` + `can_see_*`. Kiosk mode locks non-admins to read-only at both the scope and fine layers.

## Should fix — visibility-coverage gaps
These mutation endpoints fetch by id and 404 **only on non-existence**, with no `can_see_*` mask, unlike their siblings (`delete_roms`, `update_platform`) — so a hidden entity is editable and its existence is confirmed:
- **`update_rom`** (PUT `/roms/{id}`, `endpoints/roms/__init__.py:1316-1319`) — medium: an editor can edit metadata/cover of a ROM hidden from them. Mirror `delete_roms`: `if not rom or not perms.can_see_rom(rom.id, rom.platform_id): raise RomNotFoundInDatabaseException(id)`.
- **`update_rom_user`** (PUT `/roms/{id}/props`, `:1828-1831`) — low/medium: writes the caller's own RomUser row + confirms existence of a hidden ROM.
- **`add_firmware`** (`endpoints/firmware.py:48-52`) — low: upload firmware to a hidden platform (platform-hide should cascade to firmware).

## Should fix — gaps in *pre-existing* endpoints the new model doesn't cover
These aren't in this PR's diff (so no inline comment), but the hidden-entity feature is incomplete without them — flagging because they defeat the feature's purpose:
- **`patch_rom`** (`endpoints/roms/patch.py`, gated only `[Scope.ROMS_READ]`) — **high**: looks up a `RomFile` by id and **streams the patched ROM bytes back** with no `can_see_rom` check. Any user can download the *content* of a hidden ROM by file id. This is the only path that leaks actual file bytes; worth resolving in this PR since it's the feature's headline guarantee. Fix: resolve `rom_file.rom` and 404 unless `can_see_rom(rom.id, rom.platform_id)` (and the same for the patch file).
- **`get_all_activity` / `get_rom_activity`** (`endpoints/activity.py`) — medium: return `rom_name`/`rom_cover_path`/`platform_*` for live sessions across all users with no hidden filter, leaking hidden ROM/platform metadata.

## Migration — `0092` is not Postgres-safe
`alembic/versions/0092_narrow_role_to_admin_user.py` converts `role` from a native `Enum` to `VARCHAR` with no `postgresql_using` and never drops the orphaned PG `role` type. The repo's own migrations show both required patterns (`0010_igdb_id_integerr.py` uses `postgresql_using="...::text"`; `0026`/`0033`/`0075` do `ENUM(name=...).drop(connection)`). Consequences on Postgres: the upgrade likely fails the enum→varchar cast without `postgresql_using="role::text"` (and migrations auto-apply on startup, so that would block boot), and `downgrade()`'s `sa.Enum(name="role")` recreation collides with the lingering type ("type role already exists"). The PR notes verification on MariaDB only — please test `0092` up+down on a real Postgres instance and mirror the existing precedent. (Chain itself is a single linear head; only the dialect handling is the issue. Minor: two files share the `0091_` filename prefix — harmless but confusing.)

Net: the design and the bulk of enforcement are solid. The `can_see_*` masking just needs to be applied uniformly across every entity read/write path (the three in-diff endpoints above, plus `patch_rom`/activity), and `0092` needs Postgres dialect handling, before this is safe to merge.

---

zurdi15:
@gantoine thanks for the thorough review. All points addressed in `7d4579540`.

### Visibility-coverage gaps (in-diff)
- **`update_rom`** (PUT `/roms/{id}`) and **`update_rom_user`** (PUT `/roms/{id}/props`) now 404-mask hidden roms, mirroring `delete_roms`.
- **`add_firmware`** now 404-masks uploads to a hidden platform (platform-hide cascades to firmware).

### Pre-existing endpoints
- **`patch_rom`** (high) — resolves the parent rom of **both** the base file and the patch file and 404s unless `can_see_rom`, before any filesystem access. The hidden-rom byte leak is closed.
- **`get_all_activity` / `get_rom_activity`** — drop sessions whose rom is hidden from the caller (via `get_hidden_rom_ids_among`, so the platform cascade is honored too).

### Migration Postgres-safety
The role enum -> VARCHAR narrowing is now Postgres-safe:
- cast uses `postgresql_using="role::text"`;
- the orphaned native `role` enum type is dropped after the alter (`checkfirst=True`);
- `downgrade()` recreates the type explicitly with `create_type=False` before re-typing the column, casting back with `postgresql_using="role::role"`.

Verified locally on a real Postgres 16 with an **up -> down -> up** cycle: column ends as `varchar` with no orphaned type, downgrade restores the enum, and re-upgrade has no "type already exists" collision. The PR's `migrate-postgres` and `pytest (postgresql)` CI jobs are green.

The two migrations are also collapsed into a single `0092_permission_system` (so the duplicate `0091_` filename prefix is gone), and I added a comment in the resolver documenting the override `own_only`-replacement granularity limit you flagged.

Added tests for the newly-masked `update_rom` / `update_rom_user` / `patch_rom` paths in `test_permissions_visibility.py`.

_AI assistance: these fixes were implemented with Claude Code._

docs/permission-system-redesign.md
```diff
@@ -0,0 +1,216 @@
...
+**Delete:** `frontend/src/v2/composables/useCan/role-map.ts` (PR 6).
+
+---
+
+## Verification (end-to-end)
+
+- **Backend**: `pytest` on resolver (precedence, ownership, own_only), projection-parity, migration up/down + backfill assertions, visibility filtering (list + count + 404-mask), last-admin guard, OIDC/invite mapping, kiosk read-only.
+- **The parity test is the linchpin**: for a DB seeded with admin/editor/viewer users, assert each user's projected `oauth_scopes` equals their pre-migration scopes exactly — this proves the upgrade changes nobody's access.
+- **Frontend** (CLAUDE.md §VII): `npm run typecheck`, `npm run lint`, `npm run test`, `npm run generate` after backend API changes. Browser-test `uiVersion=v2`: a `user` with restricted group sees gated controls hidden (`v-if`), a hidden platform absent from their gallery; admin sees everything; live update when an admin changes their permissions (socket). Both themes, four modalities.
+- **Manual upgrade test**: run the migration against a populated dev DB; confirm existing editors/viewers retain identical capabilities (including delete) and visibility.
```
gantoine: we don't store these typically

backend/models/user.py
```diff
@@ -109,16 +132,12 @@ def kiosk_mode_user(cls) -> User:
...
-
-        if self.role == Role.EDITOR:
-            return EDIT_SCOPES
-
-        if KIOSK_MODE:
-            return READ_SCOPES
+        # Derived from the granular permission model (groups + overrides),
+        # projected onto the legacy coarse scopes. Resolved per access via its
+        # own session; admins short-circuit without touching the DB.
+        from handler.auth.permissions import compute_oauth_scopes
```
gantoine, line 139: can be imported at the top of the file?

backend/models/permission.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True)
+    user_id: Mapped[int] = mapped_column(
+        ForeignKey("users.id", ondelete="CASCADE"), index=True
+    )
+    entity: Mapped[PermEntity] = mapped_column(_str_enum(PermEntity, 20))
+    action: Mapped[PermAction] = mapped_column(_str_enum(PermAction, 10))
+    granted: Mapped[bool] = mapped_column(Boolean)
+    own_only: Mapped[bool] = mapped_column(Boolean, default=False)
```
gantoine, line 136: why an override instead of directly editing the user's matrix?

backend/endpoints/responses/rom.py
```diff
@@ -406,6 +406,22 @@ def from_rom(cls, rom: Rom, *, is_main_sibling: bool = False) -> SiblingRomSchem
...
+    """`db_rom.sibling_roms` minus any sibling hidden from the caller.
+
+    Single-rom endpoints (detail / simple fallback) read siblings off the
+    eager-loaded relationship, which bypasses the list query's hidden filter.
+    """
+    siblings = list(db_rom.sibling_roms)
+    if not request.user.is_authenticated:
+        return siblings
+
+    from handler.auth.dependencies import get_permissions
```
gantoine, line 420: move to top of file?

backend/endpoints/roms/__init__.py
```diff
@@ -987,6 +1033,12 @@ def get_rom(
     if not rom:
         raise RomNotFoundInDatabaseException(id)
 
+    # 404-mask roms hidden from the caller (skip when unauthenticated download).
+    if request.user.is_authenticated and not get_permissions(request).can_see_rom(
+        rom.id, rom.platform_id
+    ):
+        raise RomNotFoundInDatabaseException(id)
```
gantoine: this is repeated a lot, should probably be a decorator

backend/endpoints/collections.py
```diff
@@ -42,6 +43,43 @@
...
+        list(perms.hidden_platform_ids),
+        list(perms.hidden_rom_ids),
+    )
+    if not hidden:
+        return schemas
+    for s in schemas:
+        visible = set(s.rom_ids) - hidden
+        if len(visible) != len(s.rom_ids):
+            s.rom_ids = visible
+            s.rom_count = len(visible)
```
gantoine, line 80: dunno how i feel about this pattern, but let's merge and get the AI to refactor later if needed!

backend/endpoints/firmware.py
```diff
@@ -190,6 +212,15 @@ def head_firmware_content(request: Request, id: int, file_name: str):
...
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=error)
 
+    # Firmware inherits its platform's visibility; 404-mask when hidden.
+    if request.user.is_authenticated and not get_permissions(request).can_see_platform(
+        firmware.platform_id
+    ):
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Firmware with ID {id} not found",
+        )
```
gantoine: same decorator idea here

gantoine (changes requested):

docs/permission-system-redesign.md
zurdi15: oops

backend/endpoints/responses/rom.py
zurdi15: same circular import issue

backend/models/permission.py
zurdi15: because that way any change made to the default group wll be reflected in that user permissions. Otherwise we are freezing the group permissions at the time of assign them to the user

backend/models/user.py
zurdi15: it's a local import to break a circular import

`models.user` → `handler.auth.permissions` → `decorators.database/models.user`

backend/endpoints/roms/__init__.py
zurdi15: instead of a decorator I think it's better to build a helper in this case, a decorator would make us call the `db_rom_handler.get_rom(id)` twice

gantoine (approved):

Issue #3600, opened 2026-06-25
[Bug] PSP games in `.cso` (compressed ISO) format never match on RetroAchievements - RAHasher fails with "Could not open track"


Spinnich:
**RomM version**
5.0.0-alpha.1

**Describe the bug**

PSP games stored as `.cso` files (a compressed ISO format) never get a
RetroAchievements (RA) match during a scan, even when the game is correct and
plays fine in PPSSPP.

For example, scanning `3rd Birthday, The (Europe).cso` does not match the
expected RA entry at https://retroachievements.org/game/3511/hashes. The
backend log shows:

```
[rahasher] RAHasher failed with code 1. stderr='Could not open track\n'
```

RomM treats this failed run as "no hash" and moves on, so the game is left
without a RetroAchievements link.

**To Reproduce**

1. Put a PSP game in `.cso` (compressed ISO) format into your PSP library
   folder, e.g. `3rd Birthday, The (Europe).cso`.
2. Run a scan with RetroAchievements / hash calculation enabled.
3. Open the scan/backend logs.
4. See the error: `RAHasher failed with code 1. stderr='Could not open track\n'`,
   and note the game has no RetroAchievements match.

**Expected behavior**

The `.cso` should match the same RetroAchievements game it matches in PPSSPP
(here, game 3511). RA hashes for PSP are based on the contents of the disc, so
a `.cso` and the equivalent uncompressed `.iso` produce the same RA hash. RomM
should be able to identify a PSP `.cso` just like an `.iso`.

**Additional context**

Root cause (for maintainers):

RomM shells out to the external `RAHasher` binary to compute the RA hash. For a
single `.cso` file on a disc-based platform (PSP), the file path is passed to
`RAHasher` unchanged. `RAHasher` (built from RALibretro, using rcheevos' disc
reader) understands raw `.iso`, `.bin`/`.cue`, and `.chd`, but it does **not**
understand the CSO / CISO compressed-ISO container. So it cannot open the disc
and exits with `Could not open track`.

This works in PPSSPP because standalone PPSSPP has its own CSO support: it reads
the CSO as a virtual disc and decompresses only the blocks it needs, then hashes.
The standalone `RAHasher` CLI has no such layer.

Why RomM's existing guards don't catch it:
- `.cso` is not in `COMPRESSED_FILE_EXTENSIONS`
  (`backend/utils/filesystem.py`), which only lists generic archives
[...]

PR #3609, opened 2026-06-26
fix(ra): hash PSP compressed-ISO containers natively (#3600)

gantoine:
PSP games stored as .cso (and .ciso/.zso/.dax) never matched on RetroAchievements because RomM hands these files to the external RAHasher binary, which can't read CSO/ZISO/DAX compressed-ISO containers and fails with "Could not open track". The game was then left without an RA link.

The PSP RA hash does not depend on the whole disc: rcheevos computes it as MD5(PSP_GAME/PARAM.SFO + PSP_GAME/SYSDIR/EBOOT.BIN), each file capped at 64 MiB. Those containers are block-indexed, so we can decompress just the blocks those two files occupy and hash them in-process, producing the identical hash RA stores (and PPSSPP computes).

Add utils/psp_hasher.py: a seekable block-decoder for CISO/ZISO (deflate and LZ4) and DAX containers, a minimal ISO9660 reader to locate the two files, and calculate_psp_ra_hash(). RAHasherService now computes the hash natively for PSP .cso/.ciso/.zso/.dax files, falling back to RAHasher on any failure. Raw .iso PSP discs still go to RAHasher, which reads them fine.

Claude-Session: https://claude.ai/code/session_01KsyrT6n8WYWWxR2XGoEASs

**Description**

**Checklist**

### DIFF
--- a/backend/utils/psp_hasher.py
+++ b/backend/utils/psp_hasher.py
@@ -350,7 +350,14 @@ def calculate_psp_ra_hash(file_path: str) -> str:
         iso = _Iso9660(image)
         param_sfo = _read_iso_file(iso, image, _PARAM_SFO_PATH)
         eboot_bin = _read_iso_file(iso, image, _EBOOT_BIN_PATH)
-    except (OSError, PspHashError, struct.error) as exc:
+    except (
+        OSError,
+        PspHashError,
+        struct.error,
+        zlib.error,
+        ValueError,
+        IndexError,
+    ) as exc:
         log.warning(
             f"Failed to read PSP files from {hl(file_path)} for native "
             f"{hl('RA', color=LIGHTMAGENTA)} hashing: {exc}"
--- a/backend/utils/psp_hasher.py
+++ b/backend/utils/psp_hasher.py
@@ -1,0 +1,377 @@
+"""Native RetroAchievements hashing for PSP compressed-ISO containers.
+
+RAHasher (RALibretro/rcheevos) reads raw ``.iso``, ``.bin``/``.cue`` and
+``.chd`` PSP discs, but not the block-compressed ISO containers PPSSPP uses
+(``.cso``/``.ciso``/``.zso``/``.dax``). For those it fails with "Could not open
+track" and the game is left without a RetroAchievements match.
+
+The PSP RA hash does not depend on the whole disc. rcheevos computes it as
+``MD5(PSP_GAME/PARAM.SFO contents + PSP_GAME/SYSDIR/EBOOT.BIN contents)``, each
+file capped at 64 MiB. Those two files are only a few MB, and the compressed
+containers are block-indexed, so we can decompress just the handful of blocks
+they occupy and hash them in-process, with no subprocess and no multi-GB
+temporary ISO. This produces the identical hash RA stores (and PPSSPP computes),
+so the game matches.
+"""
+
+import hashlib
+import os
+import struct
+import zlib
+from typing import BinaryIO
+
+from logger.formatter import LIGHTMAGENTA
+from logger.formatter import highlight as hl
+from logger.logger import log
+
+# Container extensions RAHasher can't read but which we can hash natively. The
+# real container is still detected by magic; the extension only gates whether we
+# attempt native hashing before falling back to RAHasher.
+PSP_NATIVE_HASH_EXTENSIONS: tuple[str, ...] = (".cso", ".ciso", ".zso", ".dax")
+
+# Files that make up the PSP RA hash, in order, relative to the ISO root.
+_PARAM_SFO_PATH = "PSP_GAME/PARAM.SFO"
+_EBOOT_BIN_PATH = "PSP_GAME/SYSDIR/EBOOT.BIN"
+
+# rcheevos caps each hashed file at MAX_BUFFER_SIZE (64 MiB). We mirror that so
+# our hash matches RA's database byte-for-byte even for oversized EBOOT.BIN.
+_MAX_HASH_BYTES = 64 * 1024 * 1024
+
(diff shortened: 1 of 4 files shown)


backend/utils/psp_hasher.py
```diff
@@ -0,0 +1,377 @@
...
+            f"Could not open PSP container {hl(file_path)} for native "
+            f"{hl('RA', color=LIGHTMAGENTA)} hashing: {exc}"
+        )
+        return ""
+
+    try:
+        iso = _Iso9660(image)
+        param_sfo = _read_iso_file(iso, image, _PARAM_SFO_PATH)
+        eboot_bin = _read_iso_file(iso, image, _EBOOT_BIN_PATH)
+    except (OSError, PspHashError, struct.error) as exc:
```
gantoine: **This `except` is too narrow to deliver the documented "any failure falls back to RAHasher" guarantee.** The block decoders invoked from `_read_iso_file` → `image.read_at` → `_decode_block` raise exception types not listed here:
- `_inflate` raises `zlib.error` on a corrupt/edge deflate block (also possible for CISO files written with `align > 0`, where `next_pos - pos` includes inter-block alignment padding that gets handed to `zlib.decompress`).
- `_lz4_decompress_block` indexes `src[i + 1]` (line 86) with no bounds guard, so a truncated ZISO block raises `IndexError`.

Neither `zlib.error` nor `IndexError` is in `(OSError, PspHashError, struct.error)`, and the caller in `rahasher.py:216` (`native_hash = await asyncio.to_thread(calculate_psp_ra_hash, file_path)`) isn't wrapped either — so instead of returning `""` and falling through to RAHasher, a malformed container propagates the exception out of `calculate_hash` and aborts hashing for that ROM (potentially the scan). That's worse than pre-PR behavior, which the module docstring explicitly says it avoids.

Simplest fix: catch broadly here since the contract is "return '' on any parse failure" —
```python
except (OSError, PspHashError, struct.error, zlib.error, ValueError, IndexError) as exc:
```
or just `except Exception` for this best-effort path (and likewise widen the `_open_disc_image` except at line 342). Belt-and-suspenders, you could also guard the `to_thread` call in `rahasher.py`.

---

backend/utils/psp_hasher.py
```diff
@@ -0,0 +1,377 @@
...
+                i += 1
+                literal_len += b
+                if b != 0xFF:
+                    break
+        out += src[i : i + literal_len]
+        i += literal_len
+        if i >= n:
+            break  # trailing literals: end of block
+
+        offset = src[i] | (src[i + 1] << 8)
```
gantoine, line 86: Minor robustness in the LZ4 decoder: `src[i + 1]` (and the `src[i]` reads in the extended-length loops) have no bounds check, so a truncated/corrupt ZISO block raises `IndexError` rather than producing a short/empty result. Not a problem for well-formed input, but it's one of the exception types that currently escapes `calculate_psp_ra_hash` (see the comment on the `except` clause). Either guard the reads (`if i + 1 >= n: break`) or rely on the broadened `except` to turn it into a clean RAHasher fallback.

---

gantoine (commented):
Nice, well-scoped fix — decompressing only the two blocks that back the PSP RA hash instead of materializing a multi-GB ISO is exactly the right approach, and the synthetic CISO/ZISO/DAX + ISO9660 test builders are a great way to exercise the decoders. I read through the format parsers in detail; they look correct:

- **CISO/ZISO** header + index parsing, the `align` shift, the plain-block flag, and deflate-vs-LZ4 dispatch are right; `_inflate` trying raw-deflate then zlib-wrapped correctly handles both maxcso and classic-`ciso` output.
- **LZ4** block decode handles extended literal/match lengths and overlapping back-references correctly.
- **DAX** fixed-block + NC-area (plain block) handling looks right.
- **ISO9660** walk (PVD root record at offset 156, LBA*2048 addressing, per-sector record iteration with zero-length sector skip, `;version` stripping, dir-vs-file flag, multi-sector directories) is a sound minimal reader.
- The 64 MiB per-file cap and `MD5(PARAM.SFO + EBOOT.BIN)` match the documented rcheevos algorithm, and the integration is correctly gated on PSP `ra_id` + extension with a documented RAHasher fallback.

One issue worth fixing before this leaves draft, plus two notes:

1. **Exception handling doesn't actually guarantee the RAHasher fallback** (inline). `calculate_psp_ra_hash` only catches `(OSError, PspHashError, struct.error)`, but the block decoders raise `zlib.error` (corrupt/edge deflate block, or `align>0` padding) and `IndexError` (truncated LZ4), and the caller doesn't wrap the `to_thread` call. So a malformed container raises out of `calculate_hash` instead of returning `""` and falling back — worse than pre-PR behavior, which the docstring says it avoids. Broaden the `except` (the contract is "return '' on any parse failure").

2. **Ground-truth validation is missing.** The tests are a closed loop (the test encoder feeds the decoder), so they prove internal consistency but not that the output equals the hash RetroAchievements actually stores. The PR's whole value rests on "produces the identical hash RA stores," and the checklist has *tested locally* unchecked. Before un-drafting, worth confirming against at least one real `.cso` whose RA hash you can compare to the live RA database / PPSSPP. (This is also why it's still a draft, I assume — just flagging.)

3. **`align > 0` CISO padding** (folded into #1): for large (>2 GB) CSOs that use index alignment, `next_pos - pos` includes padding bytes passed to `zlib.decompress`; the tests only cover `align=0`. Broadening the `except` makes this degrade gracefully, but if you want native hashing to actually work on those files, trim to the real compressed length.

Overall the core logic is solid; mainly the failure-isolation needs tightening so the "no worse than before" promise holds.

---

Spinnich:
You're the man! Did not expect to see this get rolled into v5.0.

PR #3616, opened 2026-06-26
fix(screenscraper): stop persisting SS media download URLs

### LINKED ISSUE #3612: [Bug] Vestigial ScreenScraper download URLs are still saved in the database after assets are downloaded locally
**RomM version**
5.0.0-alpha.1

**Describe the bug**
When RomM scrapes a game from ScreenScraper, it saves the ScreenScraper *download URLs* for the cover, manual, screenshots, and the extra artwork (box art, logos, bezels, videos, etc.) into the database. They are stored in several places on the `roms` table:

- `url_cover`
- `url_manual`
- `url_screenshots`
- the `ss_metadata` JSON blob, which holds ~18 `*_url` entries (`box2d_url`, `screenshot_url`, `manual_url`, `video_url`, and so on)

The problem: during that very same scan, RomM already **downloads every one of those assets and saves a local copy** (the matching `path_cover_*`, `path_screenshots`, `path_manual`, and `ss_metadata.*_path` fields). The web UI only ever displays those local copies. The saved ScreenScraper URLs are never used to show anything to the user afterward, so for the `ss_metadata.*_url` entries we're keeping data that nothing reads.

This looks like a leftover from an earlier design, when the plan was to point at ScreenScraper's servers on demand instead of keeping local copies. That design changed, but the URL-saving stuck around.

These ScreenScraper URLs are not plain links. They carry authentication details needed to fetch from ScreenScraper. Keeping them in the database (and including them in API responses) means that sensitive information is written into a place it no longer needs to live, which is an avoidable risk for anyone who backs up or shares their database, logs, or support bundles. Removing what we don't need is a straightforward security and privacy improvement.

**To Reproduce / How to observe it**
1. Add a game and scan it with ScreenScraper enabled.
2. Inspect the `roms` row for that game (or the `/api/roms/{id}` response).
3. Note that `url_cover`, `url_manual`, and the `ss_metadata` `*_url` fields contain full ScreenScraper URLs, including query parameters.
4. Confirm the game's images also exist as local files (`path_cover_*`, `path_screenshots`, `ss_metadata.*_path`) and that the UI renders from those local copies, not from the stored URLs.

**Expected behavior**
RomM should not retain ScreenScraper download URLs (or any embedded authentication parameters) once the asset has been saved locally. Concretely:

- The ScreenScraper-only `ss_metadata.*_url` fields appear safe to stop storing entirely — nothing reads them after the download step. Existing rows could be cleaned up via a migration.
- The shared `url_cover` / `url_manual` / `url_screenshots` columns are also used by other metadata providers and by the "has this image changed since last scan?" check on re-scan, so they shouldn't be dropped. But they don't need to keep any authentication query parameters — those can be stripped (or the field reduced to just what the change-check needs) before saving.

**Additional context**
- The asset download already happens in the scan flow, so this is about *not persisting* something we currently throw away anyway, not about losing functionality.
- The stored URLs are also returned in API responses (`url_cover`, `url_manual`, and the full `ss_metadata` object), so this affects API output as well as the database.
- A fix should include cleaning up data already stored in existing databases, not just changing new scans.

Spinnich:
**Description**

ScreenScraper media URLs were saved into the database during a scan, even though every asset is also downloaded to a local file that the UI actually renders from. They lived in two places on the `roms` table:

- the shared `url_cover` / `url_manual` / `url_screenshots` columns, and
- ~18 `*_url` entries inside the `ss_metadata` JSON blob.

These URLs are ScreenScraper API queries that carry authentication parameters (`devid`/`devpassword`/`ssid`/`sspassword`) and are never read after the in-scan download. Persisting them writes authentication detail into the DB (and into `/api/roms/*` responses) for no functional gain — an avoidable risk for anyone who backs up or shares their database, logs, or support bundles. This closes #3612.

**The approach:** the SS URLs are now held only in memory long enough to drive the download, then dropped before anything is persisted. Because the assets are keyed by the stored `ss_id` (non-sensitive) and every scan that downloads SS art already re-queries ScreenScraper, the URLs are always re-derivable and never need to be stored.

The one thing the stored URL was still used for after a scan is the re-scan "has this asset changed?" check. To preserve that without storing a query, we now persist a tiny, non-sensitive **variant tag** instead — the resolved `media` descriptor, e.g. `box-2D(wor)` — in `ss_metadata` (`cover_media` / `manual_media` / `screenshot_media`). On a re-scan, comparing that tag detects when a config change (e.g. a new region priority) selects a different asset, so UPDATE-scan refresh keeps working. Other metadata providers' URLs (IGDB, Moby, …) are plain public CDN links and are left untouched.

### What changed and why

- **Stop persisting SS URLs.** `ss_metadata.*_url` is dropped at the persist boundary; SS-sourced `url_cover`/`url_manual`/`url_screenshots` are blanked. Non-SS provider URLs are kept as valid cover fallbacks.
- **Keep the UPDATE-scan change check** via a non-sensitive variant tag rather than the URL, so config-driven art changes still re-download.
- **Re-attach credentials only at download time**, host-gated to `screenscraper.fr`, and never store them anywhere.
- **Clean up existing rows** with a data migration.

### Files modified

- `backend/handler/metadata/ss_handler.py` — Core logic. Strip the full SS auth param set (`ssid`/`sspassword`/`devid`/`devpassword`) from in-memory media URLs; re-add the full set in `add_ss_auth_to_url` at download time. New helpers: `pop_ss_media_urls` (drop the `*_url` keys), `_ss_media_descriptor` (extract the `media` tag), public `is_screenscraper_url` (host gate, renamed from a private helper). New `SSMetadata` tag fields and tag computation in `build_ss_game`.
- `backend/endpoints/sockets/scan.py` — Capture the SS media URLs and blank the SS-sourced url columns before `add_rom`; download cover/manual/screenshots and special media from the captured URLs; change the cover/manual/screenshots `overwrite` decision to "stored URL changed (non-SS) OR variant tag changed (SS)."
- `backend/endpoints/roms/__init__.py` — Manual-edit / `ss_id`-change flow: drop `*_url` and blank SS-sourced `url_cover`/`url_manual` before `update_rom` (after the download has consumed them).
- `backend/alembic/versions/0092_strip_ss_media_urls.py` — Batched data migration: drop every `*_url` from `ss_metadata`, and blank `url_cover`/`url_manual` / drop SS entries from `url_screenshots` only when they point at ScreenScraper. Cross-dialect (typed `CustomJSON`). Downgrade is intentionally a no-op (the stripped data can't be reconstructed; local assets remain valid).
- `backend/tests/handler/metadata/test_ss_handler.py` — Tests for auth stripping, dev-cred re-attachment, `pop_ss_media_urls`, `is_screenscraper_url`, `_ss_media_descriptor`, and `build_ss_game` tag output.
- `backend/tests/alembic/test_0092_strip_ss_media_urls.py` (+ `__init__.py`) — Unit tests for the migration's transform helpers (host gating, URL blanking, `*_url` removal).
- `frontend/src/__generated__/models/RomSSMetadata.ts` — Regenerated types; `RomSSMetadata` gained the three optional tag fields (additive, backward-compatible).

### Testing notes

- Full backend suite green (1817 passed); `trunk fmt && trunk check` (ruff/black/isort/mypy/bandit) clean.
- Migration verified on the dev MariaDB: `upgrade head` → `downgrade -1` → `upgrade head`.
- Verified against full real scans (ScreenScraper enabled):
  - Initial/COMPLETE: art downloads succeed, credentials re-attached and redacted in logs, `url_cover` stored blank, `cover_media` tag stored, `ss_metadata.*_url` absent.
  - UPDATE with no config change: ScreenScraper re-queried for ~107 roms, **zero** redundant asset downloads (the tag check correctly skipped everything).
  - UPDATE after changing region priority: only the assets whose resolved region actually changed re-downloaded (all `200`); covers stayed put as expected, since RomM's region selection is dominated by each rom's own filename region tag.

### Reviewer attention

- **The `overwrite` decision in `scan.py`** is the core new logic (`url changed OR SS variant tag changed`). Most relevant on UPDATE scans.
- **One-time self-heal:** the migration does not back-fill the new tags (they can't be derived without re-querying SS), so the first UPDATE scan after upgrade re-downloads each SS rom's art once, then settles.
- **Deliberate scope boundary:** SS "special" media (miximage, logo, bezel, marquee) still has no change check (`store_media_file` skips if the file exists), so it refreshes only on a Complete rescan while cover/screenshots refresh on UPDATE. This asymmetry is pre-existing and intentionally left out of scope; happy to file a follow-up.
- **v1 UI:** for SS roms, the legacy "re-download cover/manual" buttons and the cover "source" link key off the now-blank URL and will hide. Viewing/downloading the assets is unaffected (those use `path_*`).
- **API shape:** `ss_metadata` no longer contains `*_url` values and gains the three optional tag fields; existing consumers read `*_path`, not `*_url`.

**Checklist**

---

**AI assistance disclosure:** This PR — the code, the migration, the tests, and this description — was written primarily by Claude Code (Opus), under my direction and review. I verified the behavior against real scans locally as described above.

### DIFF
--- a/backend/alembic/versions/0092_strip_ss_media_urls.py
+++ b/backend/alembic/versions/0092_strip_ss_media_urls.py
@@ -1,0 +1,128 @@
+"""Drop stored ScreenScraper media download URLs
+
+ScreenScraper media URLs were saved into roms.ss_metadata (the ``*_url`` entries)
+and into url_cover / url_manual / url_screenshots during a scan, even though the
+assets are downloaded to local files (``*_path``) that the UI renders from. The
+stored URLs are SS API queries (carrying auth params, re-derivable from the
+stored ss_id) and are never read after the download, so this migration removes
+them from existing rows (issue #3612):
+
+- drop every ``*_url`` key from ss_metadata
+- blank url_cover / url_manual and drop screenshot entries that point at
+  ScreenScraper; other providers' public URLs are left untouched as they are
+  valid cover fallbacks
+
+The per-asset change-detection tags (ss_metadata.cover_media, etc.) are not
+back-filled here as they cannot be derived without re-querying ScreenScraper;
+the next scan repopulates them.
+
+Revision ID: 0092_strip_ss_media_urls
+Revises: 0091_unique_platform_fs_name
+Create Date: 2026-06-26 00:00:00.000000
+
+"""
+
+from urllib.parse import urlparse
+
+import sqlalchemy as sa
+from alembic import op
+
+from utils.database import CustomJSON
+
+# revision identifiers, used by Alembic.
+revision = "0092_strip_ss_media_urls"
+down_revision = "0091_unique_platform_fs_name"
+branch_labels = None
+depends_on = None
+
+BATCH_SIZE = 1000
+
+roms_table = sa.table(
+    "roms",
+    sa.column("id", sa.Integer),
+    sa.column("ss_metadata", CustomJSON()),
+    sa.column("url_cover", sa.Text),
+    sa.column("url_manual", sa.Text),
+    sa.column("url_screenshots", CustomJSON()),
+)
+
+
+def _is_screenscraper_url(url: object) -> bool:
+    if not isinstance(url, str) or not url:
+        return False
+    try:
+        host = urlparse(url).hostname
+    except ValueError:
+        return False
+    if not host:
+        return False
+    host = host.lower()
+    return host == "screenscraper.fr" or host.endswith(".screenscraper.fr")
+
+
+def _clean_ss_metadata(ss_metadata: object) -> object:
+    """Return ss_metadata without any media download URL (``*_url``) keys."""
+    if not isinstance(ss_metadata, dict):
+        return ss_metadata
+    return {
+        key: value for key, value in ss_metadata.items() if not key.endswith("_url")
+    }
+
+
+def upgrade() -> None:
+    connection = op.get_bind()
+
+    last_id = 0
+    while True:
+        rows = connection.execute(
+            sa.select(
+                roms_table.c.id,
+                roms_table.c.ss_metadata,
+                roms_table.c.url_cover,
+                roms_table.c.url_manual,
+                roms_table.c.url_screenshots,
+            )
+            .where(roms_table.c.id > last_id)
+            .order_by(roms_table.c.id)
+            .limit(BATCH_SIZE)
+        ).all()
+        if not rows:
+            break
+
+        for row in rows:
+            updates: dict = {}
+
+            if isinstance(row.ss_metadata, dict):
+                cleaned_meta = _clean_ss_metadata(row.ss_metadata)
+                if cleaned_meta != row.ss_metadata:
+                    updates["ss_metadata"] = cleaned_meta
+
+            if _is_screenscraper_url(row.url_cover):
+                updates["url_cover"] = ""
+
+            if _is_screenscraper_url(row.url_manual):
+                updates["url_manual"] = ""
+
+            if isinstance(row.url_screenshots, list):
+                kept_shots = [
+                    shot
+                    for shot in row.url_screenshots
+                    if not _is_screenscraper_url(shot)
+                ]
+                if kept_shots != row.url_screenshots:
+                    updates["url_screenshots"] = kept_shots
+
+            if updates:
+                connection.execute(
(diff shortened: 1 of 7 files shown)


gantoine:
after a quick glance this seems both too specific to ssfr and too complicated for what it's trying to accomplish. i've set it as draft and will revisit after 5.0

PR #3640, opened 2026-07-01
Add a music query API for soundtracks

tmgast:
## Description

RomM stores soundtracks but exposes no way to consume them as a music library: a
client has to enumerate ROMs and pull each one's tracks individually, and the
per-track metadata lives in the `rom_files.audio_meta` JSON blob, which is
unindexable on MariaDB, so any library-wide filter or sort degrades to a full
scan.

This adds a queryable, music-first API for soundtracks, backed by a typed and
indexed `track_meta` table migrated from that JSON.

It also lets users upload soundtracks, manuals, and screenshots to any ROM, not
only folder-based ones. When the target is a single-file ROM, the client warns
that the upload will migrate it to a multi-file (folder) structure, then performs
that migration in place automatically.

### Changes

- **`track_meta` table.** Soundtrack audio metadata moves out of the
  `rom_files.audio_meta` JSON into a typed, indexed table with a 1:1 relationship
  to its `rom_file` (`rom_file_id` PK, `rom_id` FK, both `ON DELETE CASCADE`);
  `year`/`track`/`disc` are parsed to integers. Migration `0093` backfills every
  existing row, verifies row and cover-path counts, and drops `audio_meta` only
  after the check passes. All writers (upload, scan, cover-persist, rescan
  rebuild) populate the new table. Note the response change: the rom-file
  `audio_meta` field is renamed to `track_meta`, and `year`/`track`/`disc` are
  now integers rather than strings.
- **`/api/music`.** A read-only, `roms.read`-scoped surface over `track_meta`:
  `/tracks` (paginated; `search` over title/artist/album, plus `artist`, `album`,
  `genre`, `year`, and `min_duration`/`max_duration` filters; sortable) and facet
  endpoints `/artists`, `/albums`, `/genres`, `/years` returning `{value, count}`
  for typeahead. All results are visibility-filtered against the caller's hidden
  platforms and ROMs. `GET /api/roms` also gains a `has_soundtrack` filter.
- **Single-file to folder conversion.** Uploading a soundtrack, manual, or
  screenshot to a single-file ROM promotes it to folder layout in place (same
  `rom.id`, no rescan) instead of returning 400. A shared
  `promote_single_file_to_folder` handler backs both the upload paths and a
  standalone idempotent `POST /api/roms/{id}/convert-to-folder`; the frontend
  confirms before converting.

### Tests

- Migration `0093`: backfill correctness (including messy tags and cover
  preservation), idempotent re-run, and `upgrade`/`downgrade` on both MariaDB and
  PostgreSQL.
- `/api/music`: track filtering, sorting, and pagination; facet distinct/count
  and typeahead; the duration bound; scope enforcement and per-caller visibility
  filtering.
- Conversion: id-stable in-place promotion, idempotent no-op when already
  foldered, folder-name collision to 409, extensionless staging, and auto-convert
  on the soundtrack, manual, and screenshot upload paths.
- Parser unit coverage for tag-to-column mapping; upload and scan paths both
  persist `track_meta`.

---

> AI Disclosure

**Checklist**


backend/handler/database/roms_handler.py
```diff
@@ -1603,6 +1636,224 @@ def update_rom_file(
...
+            year=year,
+            min_duration=min_duration,
+            max_duration=max_duration,
+            exclude_field=own,
+        )
+        where.append(col.is_not(None))
+        if field != "years":
+            where.append(func.length(func.trim(col)) > 0)
+        if search:
+            where.append(func.lower(col).like(f"%{search.lower()}%"))
```
gantoine: **Likely bug on PostgreSQL.** When the `/years` facet receives a `search` value, this runs `func.lower(col)` with `col = TrackMeta.year`, a `SmallInteger`. PostgreSQL has no `lower(integer)` overload and raises `function lower(integer) does not exist`. MariaDB and SQLite implicitly cast the int to text, so the existing tests (and CI) stay green.

Note the adjacent facet-value guard a few lines up is already conditioned with `if field != "years"` for exactly this reason (`trim`/`length` on an int) but the `search` branch here is not. RomM supports Postgres, and there is no `/years?search=` test to catch it.

Suggestion: skip/guard the years facet, e.g. cast to text (`func.lower(cast(col, String))`) or, since it is a year, match `TrackMeta.year == int(search)` when the input parses as a number and return nothing otherwise.

backend/models/rom.py
```diff
@@ -120,12 +125,15 @@ class RomFile(BaseModel):
...
-        CustomJSON(), default=None, nullable=True
-    )
     missing_from_fs: Mapped[bool] = mapped_column(default=False, nullable=False)
 
     rom: Mapped[Rom] = relationship(back_populates="files")
+    track_meta: Mapped[TrackMeta | None] = relationship(
+        back_populates="rom_file",
+        uselist=False,
+        cascade="all, delete-orphan",
+        lazy="selectin",
```
gantoine: **Perf/design (low).** `lazy="selectin"` here means every load of a `RomFile` collection (downloads, file listings, etc.) fires an extra batched `SELECT` against `track_meta`, even though only `SOUNDTRACK` files ever have a row and most of those callers do not need audio metadata. The old `audio_meta` column came along for free with the row.

Since `get_rom` already opts in explicitly via `selectinload(RomFile.track_meta)`, consider leaving the relationship at the default (lazy) and eager-loading only where the metadata is actually serialized. Minor, but it adds a query to some hot paths.

backend/endpoints/music.py
```diff
@@ -0,0 +1,259 @@
...
+        year=row.year,
+        track=row.track,
+        disc=row.disc,
+        duration_seconds=row.duration_seconds,
+        has_embedded_cover=row.has_embedded_cover,
+        game_name=row.game_name,
+        platform_id=row.platform_id,
+        platform_slug=row.platform_slug,
+        platform_name=row.platform_name,
+        stream_url=f"/api/roms/{row.rom_file_id}/files/content/{quote(row.file_name)}",
```
gantoine, line 52: **Consistency (low).** The `/api/music` endpoints carefully exclude the caller's hidden platforms/ROMs, but the `stream_url` (and `cover_url`) handed back here point at `/api/roms/{id}/files/content/...`, which enforces `ROMS_READ` scope but not the same per-user hidden-platform/ROM visibility filter. Discovery is already filtered so exposure is limited (a user would need to guess a `rom_file_id`), and this mirrors existing download behavior. Flagging only to confirm it matches the intended visibility semantics for the new music surface.

gantoine (commented):
Reviewed the full diff plus the surrounding code (the content-download route, the `Rom.has_soundtrack` column_property, the test setup). This is a well-executed PR. A few things stood out as genuinely good:

- **Migration `0093` is careful:** keyset batches, row + cover-path count verification before it drops `audio_meta`, idempotent guards on re-run, and up/down exercised on both MariaDB and Postgres.
- **Security hardening in the download path** (deriving media type / disposition from the trusted DB record rather than the client-supplied filename) is correct, and the `inline` audio disposition for `<audio>` Range seeking is a nice touch.
- **The commit-2 rollback refactor** (`_restore`) makes the folder conversion properly recover the filesystem on a DB failure, and it is covered by a test.

The inline comments below are the issues worth raising. The only one I would call a likely bug is #1 (a Postgres portability break on the `/years` facet); the rest are low-severity design/consistency notes.

_Disclosure: this review was AI-assisted (Claude, via PostHog Code)._

backend/handler/database/roms_handler.py
```diff
@@ -1603,6 +1636,224 @@ def update_rom_file(
...
+        max_duration: float | None = None,
+        exclude_field: str | None = None,
+    ) -> list[Any]:
+        clauses: list[Any] = []
+        if hidden_platform_ids:
+            clauses.append(Rom.platform_id.not_in(hidden_platform_ids))
+        if hidden_rom_ids:
+            clauses.append(Rom.id.not_in(hidden_rom_ids))
+        if search:
+            like = f"%{search.lower()}%"
```
gantoine: **Nit.** `search` is interpolated straight into the `like()` pattern without escaping `%`/`_`, so a value like `50%` or `a_b` is treated as a wildcard pattern rather than a literal. It is parameterized so there is no injection risk; it just produces surprising matches. Same applies to the facet `search` branch. Escaping the LIKE metacharacters (or documenting the wildcard behavior) would tighten this up.

backend/handler/rom_conversion.py
```diff
@@ -0,0 +1,75 @@
...
+    folder = rom.fs_name_no_ext
+    fs_path = rom.fs_path
+    fs_name = rom.fs_name
+    origin = f"{fs_path}/{fs_name}"
+    dest_dir = f"{fs_path}/{folder}"
+    final = f"{dest_dir}/{fs_name}"
+    staged = f"{fs_path}/{_STAGE_PREFIX}{fs_name}"
+    extensionless = folder == fs_name
+
+    if not extensionless and fs_rom_handler.validate_path(dest_dir).exists():
```
gantoine: **Edge case (low).** The `RomAlreadyExistsException` collision check is skipped in the `extensionless` branch (`folder == fs_name`). If a directory already occupies the destination name for an extensionless ROM, the subsequent `make_directory`/`move_file_or_folder` may merge into or collide with it rather than failing cleanly like the extension case does. Worth a guard (or at least a test) for the extensionless-collision path.

backend/alembic/versions/0094_track_meta_table.py
```diff
@@ -0,0 +1,200 @@
...
+            batch_op.create_index("idx_track_meta_year", ["year"])
+            batch_op.create_index("idx_track_meta_artist", ["artist"])
+            batch_op.create_index("idx_track_meta_album", ["album"])
+
+    # Backfill in keyset batches; clear first so a partial re-run starts clean.
+    src, dst = _src_rom_files(), _track_meta()
+    conn.execute(sa.delete(dst))
+    expected = expected_covers = 0
+    last_id = 0
+    while True:
```
gantoine, line 114: Follow-up on readability: keeping the transform in Python here is the right call and matches every other data-migration in the repo (`0019`, `0040`, `0057`, `0084` all iterate rows in Python; none push the mapping into SQL, which wouldn't be portable across the four supported backends anyway).

The one thing that reads as heavier than the repo norm is the manual keyset paging (`last_id`, `id > last_id`, `order_by`, `limit`). The established idiom for batching a large backfill is `Result.fetchmany(N)` on a single query — see `0084_add_roms_search_index.py` (the migration right before this one), which backfills a Python-computed column (`compute_name_sort_key`, directly analogous to `track_meta_columns`) like this:

```python
result = bind.execute(sa.select(roms.c.id, roms.c.name))
while True:
    batch = result.fetchmany(_BACKFILL_BATCH)
    if not batch:
        break
    bind.execute(update_stmt, [
        {"_id": row.id, "_key": compute_name_sort_key(row.name)} for row in batch
    ])
```

Applied here that drops `last_id`, the `id > last_id` filter, the `order_by`, and the `limit`, while keeping the batched insert and the row/cover count verification:

```python
result = conn.execute(
    sa.select(src.c.id, src.c.rom_id, src.c.audio_meta).where(src.c.audio_meta.isnot(None))
)
expected = expected_covers = 0
while True:
    batch = result.fetchmany(_BATCH)
    if not batch:
        break
    values = []
    for row in batch:
        meta = row.audio_meta
        if isinstance(meta, str):
            meta = json.loads(meta)
        cols = track_meta_columns(meta or {})
        expected_covers += cols["cover_path"] is not None
        values.append({"rom_file_id": row.id, "rom_id": row.rom_id, **cols})
    conn.execute(sa.insert(dst), values)
    expected += len(values)
```

One real trade-off to be aware of: `fetchmany` (and `0084`) buffer the full result client-side on the common MySQL/MariaDB drivers and only chunk the *inserts*, whereas your keyset `LIMIT` genuinely bounds each fetch regardless of driver. So if bounded memory on a huge library was a deliberate goal, the current loop is defensible — it just deserves a one-line comment saying that's why it's keyset-paged rather than a plain query. If it was only about batching the writes (what `0084` does), matching the `fetchmany` idiom is cleaner and more consistent with the rest of the repo.

gantoine (commented):
One follow-up note on the backfill loop, per the discussion about the `while True` block.

backend/alembic/versions/0094_track_meta_table.py
```diff
@@ -0,0 +1,200 @@
...
+                "cover_path": row.cover_path,
+            }
+            conn.execute(
+                sa.update(dst)
+                .where(dst.c.id == row.rom_file_id)
+                .values(audio_meta=meta)
+            )
+        last_id = batch[-1].rom_file_id
+
+    op.drop_table("track_meta")
```
gantoine, line 194: needs to be updated to 0094_x

backend/endpoints/music.py
tmgast: Yeah, intended...

The content route uses the same visibility model every download already does (scope-gated, not per-user hidden-filtered), and `/api/music` just hands back URLs into it. Discovery is filtered, so you can't pull hidden tracks from the music surface, you'd have to already know a `rom_file_id`, which is true of the entire existing download surface, not something this adds.

gantoine (approved):
nice! putting this on-hold until @zurdi15 merges the mobile design update

Issue #3650, opened 2026-07-01
[Bug] Wii games in `.rvz` (Dolphin compressed) format never match on RetroAchievements


Spinnich:
**RomM version**
5.0.0-alpha.2 (also present on current `master`)

**Describe the bug**

Wii games stored as `.rvz` files (the compressed disc format Dolphin creates)
never get a RetroAchievements (RA) match during a scan, even though the same file
plays fine and unlocks achievements in Dolphin.

Scanning a Wii `.rvz` leaves the game with no RetroAchievements link, and the
backend log shows:

```
[rahasher] RAHasher failed with code 1. stderr='Not a supported Wii file\n'
```

RomM treats this failed run as "no hash" and moves on, so the game is left
without a RetroAchievements link.

**To Reproduce**

1. Put a Wii game in `.rvz` format into your Wii library folder.
2. Run a scan with RetroAchievements / hash calculation enabled.
3. Open the scan/backend logs.
4. See `RAHasher failed with code 1. stderr='Not a supported Wii file'`, and note
   the game has no RetroAchievements match.

**Expected behavior**

A `.rvz` should match the same RetroAchievements game its uncompressed `.iso`
equivalent matches (and that Dolphin matches with RA enabled). RomM should be
able to identify a Wii `.rvz` just like an `.iso`.

**Additional context**

This is the Wii counterpart of the PSP `.cso` problem in #3600 and the companion
of the GameCube `.rvz` issue. It is filed separately because a correct Wii fix
has to reproduce the disc's encryption, which the GameCube case does not.

---

Root cause (for maintainers):

RomM shells out to the external `RAHasher` binary (RALibretro 1.8.3, built on
rcheevos) and hands it the `.rvz` path unchanged.

`RAHasher` identifies a Wii disc by a raw "magic word" at a fixed offset:
`0x5D1C9EA3` at byte `0x18` (confirmed in rcheevos `rc_hash_wii`). `.rvz` is a
Dolphin-specific, block-compressed container whose header does not expose that
magic, so `RAHasher` opens the file, fails the check, and exits code 1 with
`Not a supported Wii file` (verified against the shipped `RAHasher 1.8.3`).

This works in Dolphin because Dolphin's own disc layer (DiscIO) decompresses (and
re-encrypts) RVZ on the fly. The standalone `RAHasher` CLI has no such layer.

Why RomM's existing guards don't catch it: identical to the GameCube `.rvz`
[...]

PR #3657, opened 2026-07-02
fix(retroachievements): hash GameCube and Wii RVZ/WIA images natively

### LINKED ISSUE #3649: [Bug] GameCube games in `.rvz` (Dolphin compressed) format never match on RetroAchievements
**RomM version**
5.0.0-alpha.2 (also present on current `master`)

**Describe the bug**

GameCube games stored as `.rvz` files (the compressed disc format Dolphin
creates) never get a RetroAchievements (RA) match during a scan, even though the
same file plays fine and unlocks achievements in Dolphin.

Scanning a GameCube `.rvz` leaves the game with no RetroAchievements link, and
the backend log shows:

```
[rahasher] RAHasher failed with code 1. stderr='Not a Gamecube disc\n'
```

RomM treats this failed run as "no hash" and moves on, so the game is left
without a RetroAchievements link.

**To Reproduce**

1. Put a GameCube game in `.rvz` format into your GameCube library folder.
2. Run a scan with RetroAchievements / hash calculation enabled.
3. Open the scan/backend logs.
4. See `RAHasher failed with code 1. stderr='Not a Gamecube disc'`, and note the
   game has no RetroAchievements match.

**Expected behavior**

A `.rvz` should match the same RetroAchievements game its uncompressed
`.iso`/`.gcm` equivalent matches (and that Dolphin matches with RA enabled). The
GameCube RA hash is derived from the game's boot code, not from the compression,
so a `.rvz` and its equivalent `.iso` produce the same RA hash. RomM should be
able to identify a `.rvz` just like an `.iso`.

**Additional context**

This is the GameCube counterpart of the PSP `.cso` problem in #3600: the game
works in the emulator but never matches on RetroAchievements because the external
`RAHasher` binary can't read the compressed disc container.

---

Root cause (for maintainers):

RomM shells out to the external `RAHasher` binary (RALibretro 1.8.3, built on
rcheevos) to compute the RA hash, and hands it the `.rvz` path unchanged.

`RAHasher` identifies a GameCube disc by a raw "magic word" at a fixed offset:
`0xC2339F3D` at byte `0x1C` (confirmed in rcheevos `rc_hash_gamecube`). `.rvz` is
a Dolphin-specific, block-compressed container (WIA/RVZ family) whose header does
not expose that magic, so `RAHasher` opens the file, fails the magic-word check,
and exits code 1 with `Not a Gamecube disc` (verified against the shipped
`RAHasher 1.8.3`). A raw disc image carrying the magic word at `0x1C` is accepted
and hashed, so it's the RVZ container, not the game data, that it can't read.

This works in Dolphin because Dolphin's own disc layer (DiscIO) decompresses RVZ
on the fly. The standalone `RAHasher` CLI has no such layer.

Why RomM's existing guards don't catch it (same shape as #3600):
- `.rvz` is not in `COMPRESSED_FILE_EXTENSIONS` (`backend/utils/filesystem.py`),
  which only lists generic archives, so the "skip RAHasher for archived disc
  platforms" short-circuit never triggers.
- `.rvz` is not in `ARCHIVE_READERS` and is not detected as CHD, so in
  `backend/handler/filesystem/roms_handler.py` it falls into the plain
  single-file branch, which passes the raw `.rvz` path straight to
  `RAHasherService.calculate_hash()`.
- The native PSP hashing path added in #3600 only runs for the PSP RA id, so
  GameCube doesn't benefit.

A `RAHasher` version bump won't help: RVZ reading is an open, unimplemented
feature request (RetroAchievements/RALibretro#415), with no public C/C++ RVZ
library, so it would have to be ported from Dolphin. The fix belongs in RomM.

Notes toward a fix:

The GameCube RA hash is small and content-based, not a hash of the whole disc.
Per rcheevos (`rc_hash_nintendo_disc_partition`), it is:

```
MD5( boot/partition header (incl. apploader)  +  each of the up to 18 main.dol
     segments (7 code + 11 data) referenced by the apploader )
```

That's only a few MB near the start of the disc, and GameCube discs are not
encrypted, so no crypto is involved. So, exactly like the PSP `.cso` fix in
#3600, compute this natively and in-process:

1. Wrap the `.rvz` in a small seekable, random-access reader that decompresses
   only the groups a byte range touches. RVZ is group-indexed (the "128k" in the
   common "[zstd-19-128k]" sets is the group size), so this is cheap.
2. Read the boot header / apploader / DOL segments through that reader and run
   the same apploader + DOL walk `RAHasher` does.

Do not decompress the whole file to a temporary `.iso`. Dolphin doesn't, and it
isn't necessary: RVZ is block-indexed precisely so the few MB of boot code can be
read with random access. This keeps the fix subprocess-free, temp-file-free, and
fast, and it produces the identical hash RA stores.

Decompression note: RVZ groups can use Zstandard (the default and by far the most
common codec, requiring a `zstandard` dependency), plus bzip2, LZMA/LZMA2, PURGE,
or NONE (bz2/lzma are in the Python stdlib).

The same approach also covers `.wia` (RVZ's predecessor, same container family).
The Wii case is tracked separately because it additionally requires reproducing
the disc's encryption; see the companion issue.

Related references:
- Sibling issue for PSP `.cso`: #3600
- Companion issue for Wii `.rvz`: #3650 
- RAHasher RVZ support request: https://github.com/RetroAchievements/RALibretro/issues/415
- RetroAchievements GameCube hashing (apploader + DOL segments):
  https://docs.retroachievements.org/developer-docs/game-identification.html
- Dolphin RVZ/WIA implementation:
  https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/DiscIO/WIABlob.h
- Dolphin RVZ/WIA format spec:
  https://github.com/dolphin-emu/dolphin/blob/master/docs/WiaAndRvz.md

Spinnich:
**Description**

Fixes #3649 and #3650.

GameCube and Wii games stored as `.rvz`/`.wia` (the compressed disc format Dolphin
creates) never matched on RetroAchievements: the external `RAHasher` binary cannot
read the RVZ container and fails with `Not a Gamecube disc` /
`Not a supported Wii file`, so scans left those games without an RA link even
though the same files unlock achievements in Dolphin.

Neither console's RA hash depends on the whole disc, so this PR computes the hash
natively, in-process, with no subprocess and no temporary multi-GB ISO:

- A random-access RVZ/WIA reader decompresses only the groups a byte range
  touches. All codecs are supported (NONE, bzip2, LZMA, LZMA2, Zstandard), plus
  RVZ packing (junk regeneration via the lagged-Fibonacci PRNG) and stored hash
  exceptions. PURGE (legacy WIA) is rejected and falls back.
- **GameCube** (#3649): MD5 of the boot header, apploader, and the up-to-18
  `main.dol` segments, mirroring rcheevos `rc_hash_gamecube`.
- **Wii** (#3650): RA hashes *encrypted* disc bytes, but RVZ stores partition data
  decrypted with the hash tree stripped. The reader recalculates the H0/H1/H2
  hash tree, applies the container's hash exceptions, and re-encrypts clusters
  with the embedded title key (AES-128-CBC, data IV taken from the encrypted
  hash block at 0x3D0), reproducing the exact retail disc bytes rcheevos hashes.
- `RAHasherService.calculate_hash()` tries the native path for `.rvz`/`.wia` on
  the GameCube/Wii platforms and falls back to `RAHasher` on any failure, same
  shape as the PSP `.cso` native path (#3600).

**Files changed**

| File | Change |
| --- | --- |
| `backend/utils/rvz_hasher.py` | New: RVZ/WIA container reader + native GameCube/Wii RA hash implementations. |
| `backend/adapters/services/rahasher.py` | Native-hash gate for NGC/Wii `.rvz`/`.wia` before spawning `RAHasher`, with fallback. |
| `backend/tests/utils/test_rvz_hasher.py` | New: synthetic RVZ/WIA builder, spec-based reference implementations, 27 tests incl. real-ROM integration tests (skip when the dumps are absent). |
| `backend/tests/adapters/services/test_rahasher.py` | 9 new service tests: native path short-circuits the subprocess, falls back on failure, and is not triggered for other platforms or raw `.iso`. |
| `pyproject.toml` / `uv.lock` | Add `zstandard ~= 0.25`; promote `cryptography` (already transitive via joserfc/authlib) to a direct dependency for AES. |

**Testing**

- Synthetic tests cover every codec, WIA containers, groups stored raw, zero
  groups, RVZ junk packing, Wii hash exceptions, the rcheevos TMD cap, and
  update-partition skipping. The expected hashes come from independent
  reference implementations written from the specs (rcheevos `hash_disc.c`,
  Dolphin's `WiaAndRvz.md`, wiibrew), so the production reader and the tests
  meet only at the format boundary.
- Real dumps: *Mario Kart: Double Dash!! (USA)* `.rvz` hashes to
  `adad8934d2586d27ec4b653bae52e47b` and *Super Mario Galaxy (USA)* `.rvz` to
  `4e0d0d2f2c5d3c13d758b027bbcc059f`; both verified present in RA's official
  hash lists (games 7693 and 189) via `API_GetGameList`.
- Cross-check against the shipped binary: `RAHasher` 1.8.3 produces identical
  hashes when run on disc images reconstructed by this same reader, so there is
  no version-skew risk between the native path and the subprocess path.
- Timing on real files: ~0.06 s for the GameCube `.rvz` (zstd-19), ~0.42 s for
  the 3.5 GB Wii `.rvz`.
- Full backend suite: 1929 passed, no regressions.

**Notes for reviewers**

- Two rcheevos quirks are mirrored **deliberately** (matching Dolphin and
  RAHasher, and therefore RA's database): `rc_hash_wii_disc` reads the
  partition-relative data-offset field but seeks with it as an *absolute* file
  offset, and GameCube `main.dol` segment offsets are seeked relative to the
  partition, not the DOL. Both are commented in the code; "fixing" either
  breaks matching.
- The Wii encrypted-byte reconstruction (hash-tree layout, exception mapping,
  IV derivation) is the highest-risk logic; the real-ROM tests and the
  RAHasher cross-check are the strongest evidence it is byte-exact.
- Failure contract: both hash functions log and return `""` on any parse
  failure, and the service falls through to `RAHasher` (no worse than before).
- Decrypted Wii images (header byte 0x61 != 0) are explicitly not supported and
  fall back; retail RVZ dumps are always encrypted-original.
- Malformed containers are handled defensively: chunk size and entry counts are
  bounded, per-group decompression output is capped (decompression-bomb guard),
  and corrupt data (bad zstd frames, out-of-range DOL segments) fails softly to
  the RAHasher fallback rather than raising into the scan.

**Checklist**

---

> **AI assistance disclosure:** This PR was written primarily by Claude Code
> (Claude Fable 5): implementation, tests, and the validation tooling used to
> check the hashes against the RetroAchievements database and the RAHasher
> binary. Direction, test ROMs, and review were provided by the human operator.
> PR responses may also be drafted with AI assistance and will be disclosed as
> such.

### DIFF
--- a/backend/adapters/services/rahasher.py
+++ b/backend/adapters/services/rahasher.py
@@ -9,6 +9,11 @@ from logger.formatter import highlight as hl
 from logger.logger import log
 from utils.filesystem import COMPRESSED_FILE_EXTENSIONS
 from utils.psp_hasher import calculate_psp_ra_hash, is_psp_native_hash_file
+from utils.rvz_hasher import (
+    calculate_gamecube_ra_hash,
+    calculate_wii_ra_hash,
+    is_rvz_native_hash_file,
+)
 
 RAHASHER_VALID_HASH_REGEX = re.compile(r"[0-9a-f]{32}")
 
@@ -143,6 +148,8 @@ RA_BUFFER_HASH_UNSUPPORTED_IDS: frozenset[int] = frozenset(
 )
 
 PSP_RA_ID: int = PLATFORM_SLUG_TO_RETROACHIEVEMENTS_ID[UPS.PSP]
+NGC_RA_ID: int = PLATFORM_SLUG_TO_RETROACHIEVEMENTS_ID[UPS.NGC]
+WII_RA_ID: int = PLATFORM_SLUG_TO_RETROACHIEVEMENTS_ID[UPS.WII]
 
 
 class RAHasherError(Exception): ...
@@ -218,6 +225,26 @@ class RAHasherService:
                 log.debug(
                     f"Computed native {hl('RA', color=LIGHTMAGENTA)} hash for PSP "
                     f"container {hl(file_path)}"
+                )
+                return native_hash
+
+        # GameCube/Wii RVZ and WIA images can't be read by RAHasher ("Not a
+        # Gamecube disc" / "Not a supported Wii file"). Compute the RA hash
+        # natively instead, reconstructing only the disc byte ranges the hash
+        # covers. On failure we fall through to RAHasher (no worse than before).
+        if platform["ra_id"] in (NGC_RA_ID, WII_RA_ID) and is_rvz_native_hash_file(
+            file_path
+        ):
+            if platform["ra_id"] == NGC_RA_ID:
+                native_hash = await asyncio.to_thread(
+                    calculate_gamecube_ra_hash, file_path
+                )
+            else:
+                native_hash = await asyncio.to_thread(calculate_wii_ra_hash, file_path)
+            if native_hash:
+                log.debug(
+                    f"Computed native {hl('RA', color=LIGHTMAGENTA)} hash for "
+                    f"{platform['slug']} RVZ/WIA image {hl(file_path)}"
                 )
                 return native_hash
 
--- a/backend/adapters/services/rahasher.py
+++ b/backend/adapters/services/rahasher.py
@@ -235,12 +235,7 @@ class RAHasherService:
         if platform["ra_id"] in (NGC_RA_ID, WII_RA_ID) and is_rvz_native_hash_file(
             file_path
         ):
-            if platform["ra_id"] == NGC_RA_ID:
-                native_hash = await asyncio.to_thread(
-                    calculate_gamecube_ra_hash, file_path
-                )
-            else:
-                native_hash = await asyncio.to_thread(calculate_wii_ra_hash, file_path)
+           native_hash = await asyncio.to_thread(calculate_gamecube_ra_hash if platform["ra_id"] == NGC_RA_ID else calculate_wii_ra_hash, file_path)
             if native_hash:
                 log.debug(
                     f"Computed native {hl('RA', color=LIGHTMAGENTA)} hash for "
--- a/backend/adapters/services/rahasher.py
+++ b/backend/adapters/services/rahasher.py
@@ -235,7 +235,14 @@ class RAHasherService:
         if platform["ra_id"] in (NGC_RA_ID, WII_RA_ID) and is_rvz_native_hash_file(
             file_path
         ):
-           native_hash = await asyncio.to_thread(calculate_gamecube_ra_hash if platform["ra_id"] == NGC_RA_ID else calculate_wii_ra_hash, file_path)
+            native_hash = await asyncio.to_thread(
+                (
+                    calculate_gamecube_ra_hash
+                    if platform["ra_id"] == NGC_RA_ID
+                    else calculate_wii_ra_hash
+                ),
+                file_path,
+            )
             if native_hash:
                 log.debug(
                     f"Computed native {hl('RA', color=LIGHTMAGENTA)} hash for "
--- a/backend/utils/rvz_hasher.py
+++ b/backend/utils/rvz_hasher.py
@@ -32,6 +32,7 @@ Format references:
 
 import bz2
 import hashlib
+import io
 import lzma
 import os
 import struct
@@ -72,6 +73,17 @@ _MAX_GC_HEADER_SIZE = 1024 * 1024
 _MAX_WII_CLUSTERS = 1024
 _HASH_CHUNK_SIZE = 1024 * 1024
 
+# Real encoders use 32 KiB - 2 MiB chunks; the cap only rejects crafted
+# headers that would otherwise force multi-GB per-group allocations.
+_MAX_CHUNK_SIZE = 64 * 1024 * 1024
+# Dolphin's reader accepts at most 52*64 exceptions per list; used to bound
+# how much decompressed group output can precede the payload.
+_MAX_EXCEPTIONS_PER_LIST = 3328
+_MAX_EXCEPTION_LIST_BYTES = 2 + _MAX_EXCEPTIONS_PER_LIST * 22
+# Generous allowance for RVZ packing descriptors (4-byte run headers plus
+# 68-byte junk seeds) on top of the unpacked payload size.
+_RVZ_PACK_STREAM_SLACK = 0x100000
+
 _PARSE_ERRORS = (
     OSError,
     ValueError,
@@ -79,6 +91,7 @@ _PARSE_ERRORS = (
     EOFError,
     struct.error,
     lzma.LZMAError,
+    zstandard.ZstdError,
 )
 
 
@@ -234,7 +247,10 @@ class _RvzReader:
             _COMPRESSION_ZSTD,
         ):
             raise RvzHashError(f"unknown compression method {self.compression}")
-        if self.chunk_size == 0 or self.chunk_size % _WII_SECTOR_SIZE:
+        if (
+            not _WII_SECTOR_SIZE <= self.chunk_size <= _MAX_CHUNK_SIZE
+            or self.chunk_size % _WII_SECTOR_SIZE
+        ):
             raise RvzHashError("invalid chunk size")
 
         self.disc_header = header_2[0x10:0x90]
@@ -245,7 +261,7 @@ class _RvzReader:
         compr_data_len = header_2[0xD4]
         self._compr_data = header_2[0xD5 : 0xD5 + min(compr_data_len, 7)]
 
-        if n_part > 0x100 or n_raw > 0x10000 or n_groups > 0x1000000:
+        if n_part > 0x100 or n_raw > 0x10000 or n_groups > 0x100000:
             raise RvzHashError("implausible entry counts")
 
         # Partition entries are stored uncompressed. Entries smaller than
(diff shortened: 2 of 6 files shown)


backend/utils/rvz_hasher.py
```diff
@@ -0,0 +1,832 @@
...
+
+        # Per-group hash-exception lists for Wii partition data. One list per
+        # 2 MiB of chunk, or a single list for RVZ chunks below 2 MiB.
+        self._except_lists_per_group = max(1, self.chunk_size // 0x200000)
+        self._sectors_per_chunk = self.chunk_size // _WII_SECTOR_SIZE
+        self._chunk_data_size = self._sectors_per_chunk * _WII_SECTOR_DATA_SIZE
+
+        # Sized so one Wii hash group's chunks stay cached across the data
+        # and exception passes of _encrypted_hash_group.
+        self._group_cache_capacity = max(
```
gantoine: Hardening (low): the decode-group cache is bounded by *entry count* (floor of 16), but each entry can be up to `chunk_size` bytes, and `chunk_size` is validated only up to `_MAX_CHUNK_SIZE` (64 MiB). A container that declares a large `chunk_size` can therefore drive resident memory to ~16 x 64 MiB ≈ 1 GiB in the cache alone (the clear-on-full at line 467 caps the count, not the bytes), even though each individual group is bomb-guarded. Real encoders use 32 KiB–2 MiB chunks so this never triggers in practice and it degrades to a fallback rather than crashing, but since this module is explicitly defending against crafted headers, consider bounding the cache by total bytes instead of entry count.

gantoine (commented):
Focused review of `backend/utils/rvz_hasher.py` (with a pass over the `rahasher.py` gate).

This is genuinely impressive work. The container reader is careful and the failure contract is sound: bad magic, unknown/PURGE compression, truncated tables, out-of-range group indices, and corrupt codec streams all raise `RvzHashError`/`_PARSE_ERRORS`, get caught at the two public entry points, log, return `""`, and fall through to `RAHasher` (no worse than before). The decompression-bomb guards added in the second commit (per-group `max_output`, chunk-size cap, bounded exception lists, chunked zstd reads) are the right defensive posture for parsing attacker-influenceable headers. The test strategy - independent reference implementations written from the specs so production and tests "meet only at the format boundary," plus real-ROM hashes cross-checked against RA's DB and RAHasher 1.8.3 - is about as strong as validation gets for byte-exact reconstruction, and it's the main reason I'm comfortable not trying to re-derive the H0/H1/H2 + AES-CBC layout by hand.

The service gate mirrors the PSP `.cso` path exactly, runs the native hash via `asyncio.to_thread`, and only short-circuits the subprocess when the native path returns a non-empty hash. Good.

Everything below is low-severity: hardening gaps and readability nits, not correctness blockers. The `read_at` "always returns exactly `length`, zero-filling the unmapped" invariant is a nice design choice that makes all the downstream `struct.unpack` calls short-read-safe.

backend/utils/rvz_hasher.py
```diff
@@ -0,0 +1,832 @@
...
+        20-byte replacement digest).
+        """
+        exceptions: list[tuple[int, int, bytes]] = []
+        pos = 0
+        for list_ix in range(self._except_lists_per_group):
+            if pos + 2 > len(stream):
+                raise RvzHashError("truncated hash exception list")
+            (count,) = struct.unpack_from(">H", stream, pos)
+            pos += 2
+            if count > 3328:
```
gantoine: Nit: this uses the literal `3328` where the named constant `_MAX_EXCEPTIONS_PER_LIST` (= 3328) is already defined, and line 399 uses a literal `64` where `_WII_SECTORS_PER_HASH_GROUP` exists. The rest of the module is meticulous about naming its magic numbers, so these two stand out.

backend/utils/rvz_hasher.py
```diff
@@ -0,0 +1,832 @@
...
+    """
+
+    def __init__(self, fh: BinaryIO) -> None:
+        self._fh = fh
+        header_1 = fh.read(0x48)
+        if len(header_1) < 0x48:
+            raise RvzHashError("truncated file header")
+        magic = header_1[:4]
+        if magic not in (_RVZ_MAGIC, _WIA_MAGIC):
+            raise RvzHashError(f"unrecognized container magic {magic[:4]!r}")
```
gantoine: Nit: `magic` is already `header_1[:4]`, so `magic[:4]` in the error message is redundant - just `{magic!r}`.

gantoine (changes requested):

backend/adapters/services/rahasher.py
```diff
@@ -224,6 +228,26 @@ def _find_archives() -> list[Path]:
...
+        # covers. On failure we fall through to RAHasher (no worse than before).
+        if platform["ra_id"] in (NGC_RA_ID, WII_RA_ID) and is_rvz_native_hash_file(
+            file_path
+        ):
+            if platform["ra_id"] == NGC_RA_ID:
+                native_hash = await asyncio.to_thread(
+                    calculate_gamecube_ra_hash, file_path
+                )
+            else:
+                native_hash = await asyncio.to_thread(calculate_wii_ra_hash, file_path)
```
gantoine:
```suggestion
           native_hash = await asyncio.to_thread(calculate_gamecube_ra_hash if platform["ra_id"] == NGC_RA_ID else calculate_wii_ra_hash, file_path)
```

gantoine (approved):

PR #3664, opened 2026-07-03
feat(v2): responsive/mobile pass over the v2 UI

zurdi15:
**Description**

A responsive/mobile pass over the **v2 UI** (`frontend/src/v2/`). The desktop
layouts were already solid; this makes the same surfaces work from ~320px up,
across mouse/touch/keyboard/gamepad and both v2 themes, plus a couple of
fixes uncovered along the way. All layout work uses the existing
`html[data-bp~="…"]` breakpoint system (no raw `@media` for layout).

**Settings tables (mobile)**

- `RTable`: a fixed `rowHeight` no longer crushes the `xs` card-stack.
- Users / Permission Groups: the card-stack "Name" caption now lines up with
  the other rows; the group description wraps multi-line at a smaller size.
- Folder Mappings: card-stack alignment; the Platform/Type selectors group
  with their chevron and truncate long platform names instead of overflowing.
- Excluded: the "Add exclusion" dialog now scrolls (footer no longer overlaps
  the type cards).

**Overlay & form primitives (touch/mobile)**

- `RSelect`: caps the dropdown panel at the viewport (no horizontal overflow on
  phones); the search field autofocuses only on precise-pointer devices, so a
  tap no longer pops the Android keyboard while keeping desktop type-to-filter.
- `RTooltip`: reveals on mouse/pen hover and keyboard focus only (skips touch),
  and dismisses on click — so tapping a button that opens a menu/dialog no
  longer leaves a tooltip stuck open.
- `RCarousel`: constrains the fullscreen grid column so the image, thumbnails
  and counter clamp to the viewport; the thumbnail strip scrolls sideways when
  it overflows.
- `RDialog`: caps the mobile bottom-sheet below the top navbar so a tall sheet
  (Match/Edit ROM) can't ride up over it.
- `RMenu`: More/Status pickers dock as mobile bottom sheets.

**Galleries / indexes (mobile scroll + flicker)**

- Sections use `dvh` and, on `sm-and-down`, cancel the layout's bottom-nav
  padding with a matching negative margin, so content scrolls behind the tab
  bar with a single scroll (no document-level scroll stacked on the internal
  one) and the last row clears the bar.
- Fixed the sticky gallery toolbar flip-flopping while scrolling: the "stuck"
  threshold read a `position: sticky` element's `offsetTop`, which reports the
  *stuck* position once pinned. It now measures a non-sticky sentinel, and the
  toolbar ref is bound with a stable function (an inline arrow ref was
  re-invoked every render, resetting the measurement mid-scroll).

**GameDetails (mobile) + app shell**

- On mobile the whole view scrolls as one document instead of the desktop
  "fixed cover + inner panel scroll" (the Files tab scrolls now too). Title/meta
  centred, cover a touch larger.
- Action ribbon shrinks to the default (40px) size on phones and splits the
  completion/rating/difficulty metrics onto their own row with a hairline; the
  metric pills match the button height.
- `AppNav`: the logo link is pinned to its content (`justify-self: start`) so
  the empty bar space beside it stops acting as a hidden "go home" hit target.
- Scanning pill swaps the "Scanning" label for the radar glyph on `xs`.

**Fixes uncovered along the way**

- **backend** `fix(roms)`: `/roms/{id}/simple` intermittently 500'd with a
  `DetachedInstanceError` — the `with_details` loader fetched siblings without
  `platform_id`, which the per-sibling visibility check then read on a
  session-detached instance. Now eager-loaded.
- **router**: a `?tab=` change (same path) no longer runs a view transition,
  which was snapshotting the GameDetails cover's `view-transition-name` into
  the top layer over the fixed navbar.
- **reachability**: the online `/heartbeat` poll backs off from 10s to 5min so
  a network blip stops aggressively flagging the backend offline (passive
  interceptor events still flip it offline instantly; the offline poll stays
  fast for quick recovery).

> **AI assistance disclosure** (per `CONTRIBUTING.md`): these changes were
> written with AI assistance (Claude Code / Claude Opus). A human reviewed and
> tested the result on-device throughout.

**Checklist**

- [x] I've added unit tests that cover the changes <sup>(mostly responsive CSS/layout; verified via typecheck + lint + the existing Vitest/Storybook suite)</sup>


zurdi15:
## Self-review

Author self-review of this branch (AI-assisted, per `CONTRIBUTING.md`). I re-read the full `master...HEAD` diff with a critical eye rather than just re-stating what it does. Below are the things I'd flag on my own work, ranked, plus what I verified.

### Verified on the merged tree
- `vue-tsc` typecheck: **clean** (exit 0).
- Grep gate (plan requirement): **zero layout `@media (max/min-width)` left in `src/v2`** — the only match is a doc comment in `AppLayout.vue`. All layout breakpoints now go through `html[data-bp~="…"]`.
- `trunk check` (backend + frontend) clean and Vitest suite (453 tests / 18 files) green as of the pre-merge run.
- `master` merged in; one semantic duplicate (`Rom.platform_id` added to the same `load_only` on both sides) resolved.

### Findings on my own changes

**1. `Setup.vue`: a library-info load failure is now silent (behavior change).** `loadLibraryInfo` used to raise a snackbar on error; I replaced it with `console.error`, and `SetupStepPlatforms` now renders unconditionally against an empty `libraryInfo` instead of being gated on the data. The happy path is fine — the initial spinner is guarded by `v-if="loadingLibrary && step === 1"`, so there's no empty flash. The concern is only the **error path**: if `GET /library-info` fails, the user lands on a "0 platforms" step with no on-screen explanation. If the snackbar host isn't mounted under `AuthLayout` (plausible, this runs pre-login) the old snackbar never actually rendered and `console.error` is the honest call; if it did render, this is a regression. Worth confirming, and if so restoring an inline error state on the step.

**2. A desktop feature was removed to fix a mobile overflow (scope).** `SetupStepPlatforms` lost the green "folder path" chip (`roms/{slug}`) next to selected platforms, plus its `platformFolderPath` helper. That drops the affordance on **all** viewports, not just where it overflowed. A more surgical fix would hide it on `xs` only. Flagging so it reads as a conscious decision, not silent collateral.

**3. `.r-v2-two-pane` utility is defined but unused (dead code).** Added to `global.css` as the "documented two-pane convention" from the plan, but no view actually applies the class — `GameDetails` rolls its own flex-column rules. It's currently speculative CSS. I'd either adopt it in at least one view or drop it until something needs it.

**4. Two composables became never-torn-down singletons (by design, with a caveat).** `installBreakpointAttribute` now runs its watcher in a detached `effectScope(true)`, and `useInputModality` no longer removes its window listeners on unmount, so both survive the `AuthLayout ↔ AppLayout` swap (needed — otherwise `data-bp` / `data-input` drop during the transition). Correct for app-lifetime globals, but the `installed` flags now never reset, which is a caveat for HMR and any test expecting teardown.

**5. RSelect autofocus reads `matchMedia` once, non-reactively (minor).** `autofocusSearch` is a computed with no reactive dependency, so a device that gains a precise pointer mid-session (tablet + mouse) won't start autofocusing the search field. Capability rarely changes and the effect is small; leaving as-is, noting it.

**6. Interactive primitives added no `play()` coverage (process).** `RTable`'s card-stack got a Storybook story, but the behavioral changes (`RMenu sheetOnMobile`, `RTooltip` touch-gating, `RSelect` touch-autofocus skip) hinge on device/viewport signals Storybook doesn't emulate cleanly, so they're covered by hand + headless repro only. This is the "added unit tests" box left unchecked — listing the specific gaps.

### For reviewer attention (intentional, not defects)
- **Reachability poll 10s → 5min** (requested). Passive interceptor events still flip the app offline instantly on a failing request, but an idle tab whose backend dies with no in-flight REST traffic can now take up to 5 min to notice. Deliberate tradeoff.
- **The `dvh`/`lvh` + negative-margin single-scroll scheme** (`GalleryShell`, `IndexShell`, `AppLayout`, background layers) is the highest-risk area across mobile browser engines. Verified headless + on-device, but it's the first place to look if a mobile scroll regression turns up.

gantoine (approved):

PR #3689, opened 2026-07-05
feat(playmatch): move integration to API v2

DevYukine:
**Description**
Moves the Playmatch integration to the new API v2 endpoints, since v2 is the version going forward. Older clients (pre 5.0) stay supported on v1, they just won't receive new features.

While at it, this also sends the CRC32 hash as an additional identify lookup key, which v2 supports.

**Disclosure**
I am not a Python developer and am not very familiar with the RomM codebase. I used Claude to help with the endpoint switch and to validate the change against the Playmatch v2 OpenAPI spec. I reviewed the result before submitting.

**Checklist**

- [x] I've added unit tests that cover the changes (existing cassette updated to the v2 URL)

Backend change only, no screenshots needed


gantoine (approved):

gantoine:
changed my mind, this is making the cut lol

PR #3693, opened 2026-07-06
fix(platforms): stop platform variants from renaming the parent platform

### LINKED ISSUE #3157: [Bug] Arcade platform gets renamed to FBneo after I add FBNeo folder and assign platform as Arcade
**RomM version**
4.6.1

**Describe the bug**
After adding multiple arcade folders: ARCADE (for mame) and FBNeo (for FinalBurn Neo roms) and assigning them both as Platform Variant of Arcade platform, the Arcade platform name in platform picker gets renamed to whatever current name of FBNeo platform is. So before adding FBNeo folder i had "Arcade" in platform picker when assigning platform to folder in Library Manager, after adding FBNeo folder and asigning it as Platform: Arcade, Type: Platform Variant - the name of Arcade platform got renamed to FBNeo.

**To Reproduce**

1. Add FBNeo folder to roms folder
2. Run scanner - find FBNeo as unidentified platform
3. Assign FBNeo as Arcade, type: Platform Variant
4. General platform name gets for arcade gets renamed from Arcade to FBNeo

**Expected behavior**
General platform name should remain as Arcade

**Screenshots**

Image

Image

- OS: Windows - docker

- Browser [e.g. Chrome, Safari]

**Additional context**

gantoine:
**Description**

When a folder is mapped as a **Platform Variant** (or Folder Alias) of an existing platform, the scanner resolves that folder to the parent's slug. The database then holds two platform rows sharing the same `slug` (e.g. the real `arcade` folder and an `fbneo` folder both with `slug="arcade"`).

`get_supported_platforms()` built its slug→platform map with a plain dict comprehension:

```python
db_platforms_map = {p.slug: p for p in db_platforms}
```

With two rows sharing a slug, whichever came last (platforms are ordered by name) silently overwrote the parent, so the platform picker and Folder Mappings page displayed the variant's name instead of the parent's. This is what caused Arcade to be "renamed" to FBneo, and the reported WiiWare/WiiVC case.

The fix prefers the **canonical** platform (the one whose `fs_slug` matches its `slug`) when building the map, so a variant/alias folder can no longer shadow its parent.

Note: variants remain their own platform row borrowing the parent's metadata, which is the documented intent of "Platform variant". This PR only fixes the naming collision.

Fixes #3157

**Checklist**

N/A (backend-only change)

---

**AI assistance disclosure**

This change was written with AI assistance (PostHog Code). The AI performed the investigation, implemented the fix, and wrote the accompanying test; the change has been reviewed before submission.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

### DIFF
--- a/backend/tests/utils/test_platforms.py
+++ b/backend/tests/utils/test_platforms.py
@@ -1,0 +1,25 @@
+from handler.database import db_platform_handler
+from models.platform import Platform
+from utils.platforms import get_supported_platforms
+
+
+def test_supported_platform_not_shadowed_by_variant():
+    """A variant/alias folder bound to a parent slug must not shadow the parent.
+
+    Regression for the bug where adding an "fbneo" folder as a variant of
+    "arcade" renamed the Arcade platform to "FBneo" in the platform picker.
+    """
+    # Canonical Arcade platform (folder name matches the slug).
+    db_platform_handler.add_platform(
+        Platform(name="Arcade", slug="arcade", fs_slug="arcade")
+    )
+    # Variant folder resolved to the same slug during scan.
+    db_platform_handler.add_platform(
+        Platform(name="FBneo", slug="arcade", fs_slug="fbneo")
+    )
+
+    supported = get_supported_platforms()
+    arcade = next(p for p in supported if p.slug == "arcade")
+
+    assert arcade.name == "Arcade"
+    assert arcade.fs_slug == "arcade"
--- a/backend/utils/platforms.py
+++ b/backend/utils/platforms.py
@@ -26,7 +26,16 @@ def get_supported_platforms() -> list[PlatformSchema]:
         Flashpoint, and HowLongToBeat.
     """
     db_platforms = db_platform_handler.get_platforms()
-    db_platforms_map = {p.slug: p for p in db_platforms}
+
+    # Multiple folders can resolve to the same slug (a folder alias or a
+    # platform variant bound to a parent platform). When that happens, prefer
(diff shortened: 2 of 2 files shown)


Zeudan:
I think there is still an edge case after #3693.

My Arcade library only contains platform variants (FBNeo, Naomi, Atomiswave, NeoGeo, etc.). I don't have a separate/canonical Arcade platform/folder.

All of these variants are mapped to Arcade as Platform Variant.

In Folder Mappings, RomM Platform is shown as FBNeo for all of them. If I rename the FBNeo platform to Arcade, RomM Platform changes to Arcade for all variants.

It looks like, because there is no platform where fs_slug == slug == arcade, one of the variants (FBNeo in my case) is still being used as the representative for the shared arcade slug.

Is having a canonical Arcade platform expected/required for this setup, or should Folder Mappings still resolve these variants to Arcade when no canonical Arcade folder exists?

estefan3112:
Same issue here (version 6.2): If you run multiple Arcade romsets, e.g. FBNeo besides Mame2003, you do not have an Arcade directory, as you need them in two different directories for FBNeo and Mame2003. After deletion of the unused arcade directory, the behavior described above applies. There is no way to get the Folder Mapping right.

Does this mean that the arcade directory must remain intact, even if empty?

EDIT: Creating an empty arcade directory and rescanning the Arcade platform reinstated the old structure. I assign both FBNeo and Mame2003 to Arcade, but leave the folder empty.

PR #3765, opened 2026-07-16
fix(v2): batch gallery cover fetches into windows

gantoine:
**Description**

The v2 gallery is dramatically slower than v1 to load cover thumbnails on low-power devices. On an Amlogic S905W2 TV box, a grid of ~108 covers takes 1-2 minutes to fill; the v1 UI loaded the same library in 1-2 seconds.

**Root cause:** v2 hydrated each visible card with its own `GET /roms/{id}/simple` request, i.e. one HTTP round trip and one DB lookup per card. On a full grid that fires ~100 requests almost simultaneously. The browser's per-host connection limit (~6) serializes them into ~17 slow waves, and each wave pays full HTTP + JSON + a discrete backend query. v1 instead fetched a whole page (`GET /roms?limit=72&offset=...`) in 1-2 batched requests.

**Fix:** route both layouts' hydration through the store's existing `fetchWindowAt`, which aligns each visible position to its shared 72-item window and dedupes against pending/loaded windows:

- Grid: `GalleryShell.syncFetches` now calls `fetchWindowAt(p)` per visible position instead of per-card `fetchRomAt(p)`. Positions in the same window collapse to a single request (the store adds the offset to `pendingWindows` synchronously before its first `await`, so concurrent calls dedupe).
- List: `GameListRow` mount now calls `fetchWindowAt(position)`.
- Removes the now-unused `fetchRomAt` / `cancelFetchAt` store actions and their per-position `AbortController` bookkeeping. Per-window fetches are still aborted en masse by `invalidateWindows` / `resetGallery` on context switch.

A full viewport now resolves in a handful of paginated requests rather than one per card, matching v1's batched loading.

Note: two secondary GPU costs also compound this on weak GPUs but are out of scope here: the always-on full-viewport `filter: blur(28px)` background layer, and the per-card `filter: blur(16px)` cover reveal animation. Happy to follow up on those separately.

**AI assistance disclosure:** This change was authored with AI assistance (PostHog Code / Claude). The diagnosis, code changes, and comment updates were AI-generated and human-reviewed.

**Checklist**

N/A (no visual change; loading behavior only).

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -7,14 +7,14 @@ cli:
 plugins:
   sources:
     - id: trunk
-      ref: v1.7.5
+      ref: v1.10.2
       uri: https://github.com/trunk-io/plugins
 # Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
 runtimes:
   enabled:
     - go@1.21.0
     - node@22.16.0
-    - python@3.13.3
+    - python@3.14.4
 # This is the section where you manage your linters. (https://docs.trunk.io/check/configuration)
 lint:
   disabled:
@@ -22,27 +22,29 @@ lint:
     - oxipng
     - pyright
   enabled:
+    - grype@0.110.0
+    - pinact@4.0.0
     - dotenv-linter@4.0.0
     - hadolint@2.14.0
-    - markdownlint@0.48.0
-    - eslint@10.0.3
+    - markdownlint@0.49.1
+    - eslint@10.7.0
     - actionlint@1.7.11
     - bandit@1.9.4
-    - black@26.3.0
-    - checkov@3.2.508
+    - black@26.5.1
+    - checkov@3.3.8
     - git-diff-check
     - isort@8.0.1
-    - mypy@1.19.1
+    - mypy@2.3.0
     - osv-scanner@2.3.3
-    - prettier@3.8.1:
+    - prettier@3.9.5:
         packages:
           - "@trivago/prettier-plugin-sort-imports@6.0.2"
-          - "@vue/compiler-sfc@3.5.29"
-    - ruff@0.15.5
+          - "@vue/compiler-sfc@3.5.40"
+    - ruff@0.15.22
     - shellcheck@0.11.0
     - shfmt@3.6.0
     - taplo@0.10.0
-    - trivy@0.69.3
+    - trivy@0.70.0
     - trufflehog@3.93.7
     - yamllint@1.38.0
   ignore:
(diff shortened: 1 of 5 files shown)


Spinnich:
The batching direction gets a strong +1 from me. I independently measured the
same mechanism this morning on a replica of my production DB (92,800 roms) while
investigating #3771: ~44 ms of actual backend work per `/simple` call versus
300-1300 ms observed per request in the browser, i.e. the per-card latency was
connection-cap queuing, exactly as this PR describes. One window query hydrates
72 rows in 0.02s at the DB. Batching is the right fix.

Three things before merge, first one matters most:

**1. Window fetches carry the full sidecar payload and (when scoped) full
sidecar compute.** `_buildRequestParams` doesn't pass `with_char_index=false` /
`with_filter_values=false`, and `rom_id_index` is always serialized into every
`GET /roms` response. So after this PR:

- Warm cache: every 72-card window response is the full bootstrap payload
  (~930 KB decoded / ~330 KB gzipped on my library, confirmed via HAR). That is
  a lot of JSON to parse per scroll window on the low-power devices this PR
  targets.
- Scoped galleries: platform/collection/search sidecars are never cached
  (#3768), so each window fetch in a large platform gallery recomputes filter
  values live, measured at 3-4s per window on my 9,500-rom platform. Grid mode
  previously paid that once at bootstrap; this PR would pay it per window.

Suggestion: once `metadataLoaded` is true (the bootstrap or window 0 already
populated the filter panel, alpha strip, and id index), window fetches should
send `with_char_index=false` and `with_filter_values=false`. `rom_id_index`
needs a backend opt-out first, which is exactly #3770; this PR makes that
param more valuable, worth wiring here once it exists.

**2. Cancel/re-want race strands skeletons.** `cancelWindow` aborts the
controller but the offset stays in `pendingWindows` until the aborted fetch's
`finally` runs. Flick-scroll away and back within that gap and
`syncVisibleWindows` calls `fetchWindowAt(offset)`, which early-returns on the
stale `pendingWindows` entry; the abort then lands, `finally` clears the entry,
and nothing refetches until the next viewport change. The removed per-card
`cancelFetchAt` deleted its controller key synchronously, so it didn't have
this race. Fix: have `cancelWindow` remove the offset from `pendingWindows`
(and the controller map) synchronously, mirroring the old behavior.

**3. Question: list mode's row-mount fetches vs the shell's window sync.**
`GameListRow` fires `fetchWindowAt` on mount (overscan rows included), while
`syncVisibleWindows` cancels any pending window not covering a *visible*
position. If the shell's sync also runs in list mode, an overscan row's window
can be fetched on mount and then immediately cancelled by the next sync, and
since rows only fetch on mount, it won't retry until the viewport changes. If
`collectVisiblePositions` includes the overscan range this is moot; worth
confirming.

Nit: after `MAX_WINDOW_RETRIES` is exhausted, the `retryCounts` entry lingers
until a context switch; and a retry timer that fires after the user stopped
scrolling refetches a window that may no longer be visible. Both harmless,
mentioning for completeness.

(Review AI-assisted, Claude Code; measurements from the #3768 investigation
replica.)

PR #3824, opened 2026-07-19
fix(v2): batch gallery hydration into windows and cap concurrency

### LINKED ISSUE #3800: [Bug] Frequent Crashes with multiple users browsing on 5.0 interface
**RomM version**
5.0

**Describe the bug**
Frequent crashes with multiple active users browsing with the new UI.

**To Reproduce**

Browse through large lists of roms (900+) while another user is doing the same. The faster the scrolling, and presumably the more cover art being requested from the server, the more likely it is to crash. A friend and I experienced multiple crashes, each leading to a package restart, while we could see the red restart warning on the screen.

**Expected behavior**
For that to not happen.

**Screenshots**
None taken

Romm host is a Synology Diskstation 918+ on DSM 7.3.2.

- Device: Mac OS on Firefox, Windows on Chrome
- OS: Mac 26.5.1 Windows unsure

**Additional context**
Both users were accessing this host from WA and FL. The host device is in WA.

gantoine:
**Description**

Fixes #3800 — frequent crashes/restarts when multiple users fast-scroll large ROM lists (900+) on the v2 UI.

**Root cause.** The v2 gallery hydrated each visible card with its own `GET /roms/{id}/simple` request. Fast-scrolling a large library fires dozens of concurrent requests, and two users doing it at once funnel them all through a backend that defaults to a single gunicorn/uvicorn worker (`WEB_SERVER_CONCURRENCY=1`). The worker gets recycled (`--max-requests 1000`) / timed out / OOM-killed under the burst, and the container's watchdog restarts it — the "red restart warning" the reporter saw. Cover image files themselves are static (served by nginx), so the amplifier is the per-card API traffic, not the images.

Note this reverses the direction of `560a75f9` ("hydrate gallery cards per-card"). That revert optimized `/simple` latency but reintroduced the request-volume storm this issue reports.

**Fix — re-batch hydration into windows:**
- Store: replace per-card `fetchRomAt`/`cancelFetchAt` with `fetchWindowAt` + `syncVisibleWindows`. Each visible position aligns to its shared 72-item window and loads via one paginated `getRoms`; positions in the same window collapse to a single request, deduped against loaded/pending windows, with out-of-view windows cancelled and failed ones retried on bounded backoff.
- `GalleryShell.collectVisiblePositions` now covers both grid (`row`) and list (`list-row`) items, so both layouts route through the debounced `syncFetches` → `syncVisibleWindows`.
- `GameListRow` no longer self-fetches on mount; the shell drives it.

**Fix — debounce + cap the client:**
- The list path now goes through the shell's existing 80ms fetch debounce (previously grid-only; list rows fired undebounced on every mount).
- New `MAX_CONCURRENT_WINDOWS = 4` ceiling in the store: windows over the cap queue and drain as in-flight slots free up, so no scroll speed can flood the backend.

A full viewport now resolves in a handful of paginated requests instead of ~one per card.

Out of scope (defense-in-depth, can follow up separately): the single-worker default, the aggressive `--max-requests`, and the untuned DB pool that let the storm escalate into a restart.

**AI assistance disclosure:** This change was authored with AI assistance (Claude Code / Opus 4.8). The investigation, diagnosis, code changes, comments, and tests were AI-generated and human-reviewed.

**Checklist**

N/A (loading behavior only, no visual change).

---
Verification: `npm run typecheck` clean; full vitest suite passes (493 tests, incl. 3 new store tests asserting per-window batching and the concurrency cap); `trunk fmt && trunk check` clean on changed files. Not driven in a live browser — reproducing the symptom needs a 900+ ROM library under concurrent multi-user load; the new unit tests assert the batching and cap invariants directly instead.

### DIFF
--- a/frontend/src/services/api/rom.ts
+++ b/frontend/src/services/api/rom.ts
@@ -183,9 +183,10 @@ export interface GetRomsParams {
   playerCountsLogic?: string | null;
   metadataProvidersLogic?: string | null;
   tagsLogic?: string | null;
-  // Skip the char index / filter-value aggregations server-side
+  // Skip the char index / filter-value / id-index aggregations server-side
   withCharIndex?: boolean;
   withFilterValues?: boolean;
+  withRomIdIndex?: boolean;
   // Cancel an in-flight request
   signal?: AbortSignal;
 }
@@ -236,6 +237,7 @@ async function getRoms({
   tagsLogic = null,
   withCharIndex = undefined,
   withFilterValues = undefined,
+  withRomIdIndex = undefined,
   signal = undefined,
 }: GetRomsParams) {
   const params = {
@@ -345,6 +347,9 @@ async function getRoms({
     ...(withCharIndex !== undefined ? { with_char_index: withCharIndex } : {}),
     ...(withFilterValues !== undefined
       ? { with_filter_values: withFilterValues }
+      : {}),
+    ...(withRomIdIndex !== undefined
+      ? { with_rom_id_index: withRomIdIndex }
       : {}),
   };
 
--- a/frontend/src/v2/components/Gallery/GalleryShell.vue
+++ b/frontend/src/v2/components/Gallery/GalleryShell.vue
@@ -554,23 +554,20 @@ const currentLetter = computed<string>(() => {
   return "";
 });
 
-// Per-card viewport-driven fetch. The shell tracks which positions are
-// currently visible (from the rows in `viewportRange`) and keeps
-// `byPosition` in sync via per-card `GET /roms/{id}/simple` calls — pure
-// by-id lookups on the backend, much faster than the paginated
-// `getRoms(limit/offset)` pipeline. Each fetch is independent, so covers
-// stream in as their individual responses land.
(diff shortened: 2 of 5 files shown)


gantoine:
Good catch — confirmed and fixed in ff5536a.

The interrupted window was indeed marked loaded (and later syncs skipped it → permanent skeletons). While fixing it I found the same staleness class in two adjacent spots: `pendingWindows` membership is ambiguous because a *replacement* fetch for the same offset re-adds it, so the stale fetch's `finally` could also delete the new fetch's controller/pending entry.

Fix: guard the whole post-await path by AbortController identity (`inFlightControllers.get(ctrlKey) === controller`), matching the pattern the bootstrap fetch already uses — bail before applying metadata, use it as the batched-apply relevance signal, re-check before `loadedWindows.add` and before recording a failure/retry, and only clear our own bookkeeping in `finally`.

Added a regression test (`does not mark a window loaded when the context is invalidated mid-apply`) that drives a controllable frame yield, invalidates the context mid-apply, and asserts the window is not marked loaded and the fresh context can refetch offset 0.

PR #3846, opened 2026-07-20
Fix slot-less save upload DB/file desync

### LINKED ISSUE #3833: Slot-less save upload: filename lookup matches rows from any slot, and the upsert never updates file_path — DB row and file on disk diverge
**Environment:** RomM 5.0.0 (both issues reproduced live against my own instance; code refs from the `5.0.0` tag)

`POST /api/saves` without a `slot` param behaves as a filename upsert, and two details combine into silent save corruption. I traced both in the source and reproduced both against a live server:

**1. The filename lookup ignores slots.** `get_save_by_filename` only filters by slot when one is passed, and takes an unordered `.first()` (`backend/handler/database/saves_handler.py:41-46`). A slot-less upload can therefore match — and update in place — a save row that lives in a *named* slot.

*Live repro:* created a save in a named slot (server stored it datetime-tagged), then POSTed different content with that exact filename and **no** slot → the named row was updated in place (same id, `content_hash` flipped to the new upload). A slot-less client can silently mutate another client's named-slot save.

**2. The upsert doesn't move the row's file.** On the update branch only `file_size_bytes` and `content_hash` are written (`backend/endpoints/saves.py:269-276`), while the uploaded bytes go to a path built from the *current* request's `emulator` param (`saves.py:234-243`). If that differs from where the existing row's file lives, the row now carries the new `content_hash` while `file_path` still points at the old bytes.

*Live repro:* POST `test.srm` (content A, no slot, no emulator) → row created, hash = md5(A). POST different content B, same filename, no slot, `emulator=retroarch-mgba` → **same row id**, `content_hash` = md5(B), `file_size_bytes` = B's size — but `GET /api/saves/{id}/content` still returns **content A**, and the row's `emulator` stays null (the B bytes sit orphaned in the `retroarch-mgba/` subfolder). Every client that trusts `content_hash` now disagrees with what it downloads; the web player keeps loading the old save.

**Impact:** This is how danielcopper/decky-romm-sync#1478 happened — my client (in an older version) uploaded a slot-less save with an emulator tag, the untagged row got hijacked, and both the web player and my sync kept serving/downloading stale bytes; to the user it looked like lost progress.

**Question:** Is any of this intended — e.g. the cross-slot filename matching as legacy upsert semantics? The hash-of-new/bytes-of-old state in particular looks hard to intend. Expected would be: the upsert updates `file_path`/`emulator` together with the content (removing the stale file), and/or the lookup is scoped so a slot-less upload can never adopt a row from another slot or another storage path.

gantoine:
**Description**

Fixes #3833.

A slot-less `POST /api/saves` behaves as a filename upsert, and two details combined into silent save corruption (DB row and on-disk file diverging, so clients keep downloading stale bytes while `content_hash` reports the new content).

**1. The filename lookup ignored slots.** `get_save_by_filename` only filtered by slot when one was passed, so a slot-less upload could match, and overwrite, a save row living in a *named* slot. `slot=None` now scopes to the null-slot save (`Save.slot IS NULL`), so a slot-less upload can never adopt a row from another slot.
- `get_save_by_content_hash` is intentionally left as-is: it has a documented, test-pinned cross-slot `slot=None` behavior used only for harmless content dedup, not for the destructive upsert path.

**2. The upsert never moved the row's file.** The update branch only wrote `file_size_bytes` and `content_hash`, while the uploaded bytes landed under a path derived from the *current* request's `emulator`. If that differed from the existing row's location, the row carried the new hash while `file_path` still pointed at the old bytes. The upsert now also persists `file_path`/`emulator`, and removes the now-orphaned file when the row's location changes.

**Behavior change to note:** conflict detection for the (artificial) "a save stored in a named slot, re-uploaded slot-less" path now no longer cross-matches. This is the corrected behavior; a slot-less upload negotiates against the null-slot (archival/web-UI) row, and slotted uploads use the slot-scoped branch. Existing already-diverged rows are not self-healed; a subsequent re-upload reconciles them going forward (happy to add a one-shot maintenance task if wanted).

**AI assistance disclosure**
This PR was written with AI assistance (Claude Code). The AI investigated the issue, implemented the fix, and wrote the tests; changes were reviewed by the author.

**Checklist**

N/A (backend-only)

### DIFF
--- a/backend/endpoints/saves.py
+++ b/backend/endpoints/saves.py
@@ -267,11 +267,7 @@ async def add_save(
             )
 
     if db_save is None:
-        # The write above landed on saves_path/actual_filename, whose bytes can
-        # already belong to a save in another slot (file identity is path+name,
-        # independent of slot). If that row's recorded hash no longer matches
-        # what we just wrote, refresh it instead of inserting a duplicate, so it
-        # does not keep reporting metadata that mismatches the bytes on disk.
+        # Refresh hash if the file already exists to avoid mismatched metadata.
         colliding_save = db_save_handler.get_save_by_path(
             user_id=request.user.id,
             rom_id=rom.id,
@@ -282,10 +278,7 @@ async def add_save(
             db_save = colliding_save
 
     if db_save:
-        # Track where the bytes actually landed: the file is written under a
-        # path derived from this request's emulator, which may differ from the
-        # existing row's location. Persisting file_path/emulator keeps the row's
-        # hash and its on-disk content from drifting apart.
+        # Track file path and emulator to prevent hash-content drift.
         stale_full_path = db_save.full_path
         update_data: dict = {
             "file_size_bytes": scanned_save.file_size_bytes,
@@ -297,9 +290,7 @@ async def add_save(
             update_data["slot"] = slot
         db_save = db_save_handler.update_save(db_save.id, update_data)
 
-        # The row moved to a new path, so the previously stored bytes may now
-        # be orphaned. Slot is not part of the on-disk location, so another
-        # row can still point at the old path; only unlink it if none do.
+        # Delete orphaned bytes only if no other row references the old path.
         if stale_full_path != db_save.full_path:
             still_referenced = any(
                 other.id != db_save.id and other.full_path == stale_full_path
--- a/backend/endpoints/saves.py
+++ b/backend/endpoints/saves.py
@@ -267,13 +267,28 @@ async def add_save(
             )
 
     if db_save:
+        # Track where the bytes actually landed: the file is written under a
+        # path derived from this request's emulator, which may differ from the
+        # existing row's location. Persisting file_path/emulator keeps the row's
+        # hash and its on-disk content from drifting apart.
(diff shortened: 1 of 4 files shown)


danielcopper:
Thanks for the quick turnaround — scoping the lookup to `slot IS NULL` and persisting `file_path`/`emulator` on the upsert is exactly what I was hoping for.

A one-shot maintenance task would definitely be valuable: I have at least one user with an already-diverged row in the wild (the danielcopper/decky-romm-sync#1478 case that led to #3833). For those installs the web player keeps serving the stale bytes until the affected save happens to be re-uploaded through the fixed path — and users have no way to tell *which* of their saves are diverged. A task that re-scans save rows, compares `content_hash` against the file at `file_path`, and reconciles (or at least reports) mismatches would let them heal without manual surgery.

gantoine:
> A task that re-scans save rows, compares content_hash against the file at file_path, and reconciles (or at least reports) mismatches would let them heal without manual surgery.

can you try to prompt claude for this? should be able to run against the database directly (SQL)

PR #3855, opened 2026-07-20
feat(platforms): add editable platform descriptions

### LINKED ISSUE #3402: [Feature] Platform Descriptions
RomM has the ability to create custom platforms and relate them to a platform ID but lacks any way to describe what these custom platforms are for. Sometimes this is obvious to users but other times it isn't.

RomM already allows you to add a description to a collection to help add additional context to what the collection is for / trying to accomplish. Ideally, platforms should have the same function so admin's are not limited to just the platform name to describe it's purpose.

Currently, the platform name customization is the only valid solution I'm aware of.

**Additional context**
Example of what a custom platform and it's description might be used for.
Platform Name: `Sega - Genesis/ Mega Drive (Unofficial)`
Platform Description: `Contains all Aftermarket, Beta, Demo, Proto, Unl, etc. roms`

Spinnich:
**Description**

Closes #3402

RomM lets you create custom platforms and give them a custom name, but the name
is the only thing describing what a platform is for. That's often not enough
context, e.g. a platform called `Sega - Genesis/ Mega Drive (Unofficial)` gives
no hint that it holds aftermarket, beta, proto and unlicensed dumps.

Platforms now carry an optional free-text description, mirroring the description
that collections already have. It's editable from the platform Settings tab and
shown as a chip in the platform header next to the category/family/generation
chips.

Design notes:

- The column is nullable and defaults to `""`, matching `custom_name`. This
  matters because a scan builds a fresh `Platform` and merges it over the
  existing row; since the scan pipeline never populates either field, both
  survive the merge untouched. There's a dedicated test for this.
- `PUT /api/platforms/{id}` patches only the fields present in the body, so an
  omitted `description` is preserved and an explicit `""` clears it.
- v1 gets no new UI (it's frozen), but round-trips the field unchanged through
  the shared API service, so renaming a platform from the v1 drawer will not
  clobber a description set in v2.

**Files modified**

Backend
- `backend/models/platform.py` — add `description` as a nullable `Text` column
  with a `""` default, positioned next to `custom_name` to match its shape.
- `backend/alembic/versions/0099_platform_description.py` — new migration.
  Hand-written rather than autogenerated (autogenerate currently fails on this
  repo with `NoReferencedTableError` on `saves.origin_device_id`, unrelated to
  this change). Uses `if_not_exists` / `if_exists` in line with the surrounding
  migrations.
- `backend/endpoints/platform.py` — `update_platform` accepts `description` as
  an embedded body param, applied only when not `None`.
- `backend/endpoints/responses/platform.py` — `description: str | None = None`
  on `PlatformSchema`, matching the existing `custom_name` field style.
- `backend/tests/endpoints/test_platform.py` — six new endpoint tests: default
  value, set, set alongside `custom_name`, omitted-is-preserved, cleared via
  `""`, and 403 without the write scope.
- `backend/tests/handler/database/test_platforms_handler.py` — new file. Asserts
  a rescan-shaped `session.merge` does not reset `custom_name` / `description`.

Frontend (v2 only)
- `frontend/src/__generated__/models/PlatformSchema.ts`,
  `Body_update_platform_api_platforms__id__put.ts` — regenerated from OpenAPI.
- `frontend/src/services/api/platform.ts` — send `description` on update.
- `frontend/src/v2/components/Gallery/SettingsTab.vue` — the existing name-only
  form now covers both fields. Rather than adding a second save path, the
  existing `nameDirty`/`saveName`/`discardName` machinery was generalised to
  `dirty`/`save`/`discard`, so there is still one dirty check and one submit.
- `frontend/src/v2/components/Gallery/PlatformHead.vue` — new `description` prop
  rendered as a chip in the `#tags` slot, same pattern as `CollectionHead`.
- `frontend/src/v2/views/Gallery/Platform.vue` — wire the prop at both
  `PlatformHead` render branches (library vs firmware/settings).
- `frontend/src/locales/*/platform.json` — new `platform.description` key added
  to all 18 locales. Each locale reuses its own existing translation of
  "Description" from `collection.json` rather than a new or English string.

**Checklist**

**Testing notes**

- Backend: full suite, 2244 passed / 0 failed.
- Frontend: 499 vitest tests passed; `vue-tsc` clean.
- i18n: `check_i18n_locales.py` and `check_i18n_sorted.py` both pass.
- Lint: `trunk fmt` + `trunk check` clean (also enforced by the pre-commit hook).
- Migration: verified `upgrade head` and `downgrade -1` round-trip on MariaDB
  11.3, including confirming the column is dropped and re-added cleanly.
- UI: verified in a browser in both light and dark themes — the description
  renders in the header chip and is correctly bound into the Settings textarea.

**Things a reviewer should look at**

1. **`description` is unbounded.** There is no `max_length` on the body param and
   `Text` caps at 65,535 bytes, so a payload larger than that raises an
   unhandled `sqlalchemy.exc.DataError` and returns a 500 rather than a 422.
   I verified this. It requires the `PLATFORMS_WRITE` scope, so it isn't an
   anonymous DoS. `collections.description` has the same unbounded `Text` shape,
   so this matches existing behaviour rather than introducing a new pattern —
   happy to add `max_length=65535` here (and/or open a separate issue for
   collections) if you'd prefer to close it.
2. **Existing platforms return `null`, new ones return `""`.** The migration
   deliberately doesn't backfill, so migrated rows stay `NULL` until first
   edited. The frontend normalises both with `?? ""`. This mirrors `custom_name`,
   but shout if you'd rather the migration backfilled `''`.
3. **Unsaved edits are discarded across both fields.** The Settings form
   re-seeds both inputs when the platform prop changes, so a socket update that
   only touches `display_name` will also drop an in-progress description edit.
   The single-field version had the same "source of truth changed under us"
   behaviour; this widens it to two fields.
4. **No truncation on the header chip.** A long description renders as one large
   chip. `CollectionHead` behaves the same way, so I left it alone, but it's
   worth deciding whether descriptions should be length-capped in the UI.
5. **Migration numbering.** This lands as `0099`; worth a rebase check if
   another `0099` merges first.

**Note on AI assistance**

This change was written with AI assistance (Claude). The feature design, the
migration, the tests and the v2 UI changes were AI-generated and then reviewed,
executed and verified locally by me (full backend and frontend suites, migration
up/down on MariaDB, and a browser check in both themes).

#### Screenshots
image

image


Spinnich:
Thanks — all three were valid. Addressed in `82caaf87a`, `84dd839a6` and `fa2631ef1`.

**1. Oversized description returns 500** — fixed.

Confirmed the behaviour first: a 70,000-char payload raised an unhandled `sqlalchemy.exc.DataError` and surfaced as a 500. Both user-authored fields are now bounded, with the limits defined next to the columns in `models/platform.py` so the validation can't drift from the schema:

```python
CUSTOM_NAME_MAX_LENGTH = 400
DESCRIPTION_MAX_LENGTH = 65535  # TEXT
```

Scope note for reviewers: `custom_name` had the identical bug and it **predates this PR** — `VARCHAR(400)` with no bound on the body param, 500ing the same way on master. I fixed it here because it's the adjacent parameter and leaving one bounded next to an unbounded one seemed worse, but shout if you'd rather I split it out.

Three tests added: oversize rejected with 422 for each field, plus an exactly-at-max-length case so the bound isn't off by one.

**2. Name update replays stale description** — fixed.

This was the one I'd talked myself out of when writing the PR description; I had it filed as a harmless round-trip and it isn't. Good catch.

Rather than patch the call site, `updatePlatform` now makes `description` opt-in, so the failure mode is structurally impossible rather than dependent on callers holding fresh data:

```ts
updatePlatform({ platform })                 // v1 rename drawer — description never sent
updatePlatform({ platform, description })    // v2 settings — explicitly intended
```

My first attempt at this used a fully explicit `{ id, customName, description }` signature, which was cleaner but forced an edit to the v1 drawer and tripped `check-v1-frozen`. Keeping the platform-object argument means v1 stays byte-identical to master. The explicit shape is the nicer design and is worth revisiting once v1 is deleted.

**3. Canonical update discards both edits** — fixed.

Split into two independent watchers, so a socket update touching only the name no longer throws away an unsaved description, and vice versa.

---

**One more, found in manual testing rather than by review:** saving *only* a description also wrote `custom_name`. The old name-only form bailed unless the name was dirty, so `custom_name` was only ever written on a real rename; merging the two fields into one save made it unconditional. It looked like a no-op because `display_name` falls back to `name`, but it left the platform with an explicit custom name that would shadow later upstream metadata renames. Each field is now sent only when it actually changed, with a component test that fails if the name is written unconditionally.

**Verification:** backend suite 2247 passed, frontend 501 passed, `vue-tsc` clean, trunk clean, and all CI green including the migration jobs on both MariaDB and Postgres (worth noting since `alembic revision --autogenerate` currently fails repo-wide on `saves.origin_device_id` → `devices`, so this migration is hand-written). The v2 edit flow was also exercised manually end to end.

gantoine (approved):

PR #3876, opened 2026-07-22
feat(backend): categorize Switch rom files by title ID

lucid-void:
**Description**
Automatically categorizes Nintendo Switch / Switch 2 ROM files as base game, update, or DLC based on the title ID in their filename, so base + update(s) + DLC belonging to the same game can live loose in a single game folder and still be tagged individually, without requiring update/ or dlc/ subfolders.

Today, a file's RomFileCategory is derived only from the folder it sits in (e.g. a file under update/ becomes an UPDATE). This PR adds a fallback: when a Switch/Switch-2 file isn't already categorized by its directory layout, its 16-hex Nintendo title ID is parsed from the filename and classified per the rules from #2526:

- 4th-to-last nibble odd → DLC
- low 12 bits == 0x800 → UPDATE
- otherwise (low 12 bits cleared, even 4th-to-last nibble) → GAME

Example (Breath of the Wild):

┌────────┬──────────────────┬──────────┐
│  File  │     Title ID     │ Category │
├────────┼──────────────────┼──────────┤
│ Base   │ 01007EF00011E000 │ GAME     │
├────────┼──────────────────┼──────────┤
│ Update │ 01007EF00011E800 │ UPDATE   │
├────────┼──────────────────┼──────────┤
│ DLC 1  │ 01007EF00011F001 │ DLC      │
├────────┼──────────────────┼──────────┤
│ DLC 2  │ 01007EF00011F002 │ DLC      │
└────────┴──────────────────┴──────────┘

Notes / scope:

- Folder-based categorization still wins. Title-ID classification is only a fallback when the directory layout doesn't already assign a category, so existing update//dlc/ layouts are unchanged.
- Grouping onto a single game entry already works via RomM's multi-part ROM model (a directory = one Rom, its inner files = RomFiles). This change makes the loose-files-in-one-folder layout viable by tagging each file correctly.
- Tinfoil integration is unaffected. The Tinfoil feed emits one download URL per RomFile (filtered by extension, not category), so grouping/tagging files under one ROM entry doesn't change what Tinfoil sees.
- Gated strictly to the switch and switch-2 platforms; reuses the existing SWITCH_PRODUCT_ID_REGEX.

Implementation: a switch_title_id_category() helper plus a fallback branch in _build_rom_file, both in backend/handler/filesystem/roms_handler.py.

[...]

PR #3883, opened 2026-07-22
feat(scan): embed Switch title IDs into file names during scan

lucid-void:
**Description**
Adds an opt-in capability that automatically embeds Nintendo Switch title IDs into ROM file names during a scan, so external tools that parse the title ID out of the file name (e.g. CyberFoil) can index them:

Super Mario Odyssey.nsp  →  Super Mario Odyssey [0100000000010000][v0].nsp

RomM already caches the blawar TitleDB and pulls a title ID out of file names for metadata matching. This adds the reverse direction (name → title ID) and uses it to rename files that don't yet have one.

**How it works**

- New env flag ENABLE_SWITCH_TITLE_ID_RENAME (default false). Backend-behavioral only, no UI/schema change.
- The Switch TitleDB update task now also builds a reverse index (romm:switch_name_to_id) mapping a normalized base-game name to its title ID. It only keeps names that resolve to a single base title, so ambiguous names are never renamed to the wrong game. Updates (0x800) and DLC (0x1000) are excluded.
- During a scan, _maybe_add_switch_title_id renames matching files on disk (reusing the existing rename_fs_rom) and, for an already-tracked entry, moves its DB record and RomFiles onto the new name (mirroring the manual rename endpoint).

Conservative by design — it skips files that already carry a title ID, non-served extensions, nested/multi-file folders, and any name that is unknown or ambiguous. Requires no Nintendo keys.

**Scope / follow-ups**

- v1 covers flat, single-file base games (.nsp .xci .nsz .xcz .nro).
- Nested folders and update/DLC files (per-file base±offset ID math + version tags) are left as a deliberate follow-up.
- Composes cleanly with the separate Switch base/update/DLC categorization work: this produces IDs in file names, that consumes them.
- The reverse index populates on the next TitleDB update run; existing caches degrade gracefully (no index → no renames).

**AI assistance disclosure**

This PR was written with substantial AI assistance (Claude Code / Claude Opus 4.8): design, implementation, and tests were AI-generated and human-reviewed.

Checklist

### DIFF
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -213,6 +213,9 @@ SCHEDULED_UPDATE_SWITCH_TITLEDB_CRON: Final[str] = _get_env(
     "SCHEDULED_UPDATE_SWITCH_TITLEDB_CRON",
     "0 4 * * *",  # At 4:00 AM every day
 )
+ENABLE_SWITCH_TITLE_ID_RENAME: Final[bool] = safe_str_to_bool(
+    _get_env("ENABLE_SWITCH_TITLE_ID_RENAME")
+)
 ENABLE_SCHEDULED_UPDATE_LAUNCHBOX_METADATA: Final[bool] = safe_str_to_bool(
     _get_env("ENABLE_SCHEDULED_UPDATE_LAUNCHBOX_METADATA")
 )
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 from dataclasses import dataclass
 from itertools import batched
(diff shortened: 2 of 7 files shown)


tmgast:
Thanks for the PRs, this one and #3876. This overlaps with something I already had in progress: a native library (`sigil`) that reads the title ID and serial straight out of the ROM binary during a scan. We talked it over internally and decided to handle it there, since reading the Switch content metadata gets the real per-file ids and versions (base, update, and DLC) without depending on the filename, and it extends to serials and title ids on most other platforms too.

The filename embedding your PR is going for did make it in, as an opt-in config flag that renames Switch files to embed the id and version for external tools that read it out of the name. That said, I'd rather clients and users pull this from the API than transform or rename files on disk, so the rename is off by default.

It's all in #3925.

PR #3892, opened 2026-07-22
fix(nginx): cache only fingerprinted assets, revalidate the rest

### LINKED ISSUE #3890: [Bug] Static assets are served with no Cache-Control, and any fix must cache only the fingerprinted ones
**RomM version**
Observed on 5.1.0-alpha.1, but not version-specific: the responsible nginx config has no caching directives, so all current releases on the official Docker image are affected.

**Describe the bug**
RomM's nginx serves everything under `/assets` (and the root favicons/manifest) with no `Cache-Control` or `Expires` header. Without them, browsers re-fetch or revalidate these files far more than necessary on repeat visits, which slows page loads and wastes bandwidth. It is most visible in the library grid, where each cover image (~100 KB) is re-checked or re-downloaded; a 60-tile view is ~6 MB and a long scroll ~20 MB of covers alone.

**Important nuance (why a blanket rule is unsafe):**
`/assets` holds a mix of assets with different versioning, and only some are safe to cache long-term:
- Frontend build output (e.g. `index-<hash>.js`) is fingerprinted by content hash. Safe to cache forever.
- Resources under `/assets/romm/resources/**` (covers, screenshots, manuals) are versioned by a `?ts=<timestamp>` query. Safe, because the URL changes when the file does.
- `/assets/emulatorjs/**` and `/assets/ruffle/**` are third-party bundles extracted verbatim (see `docker/Dockerfile`). They have STABLE filenames and are re-versioned only when RomM bumps `EMULATORJS_VERSION` / `RUFFLE_VERSION`. NOT safe to cache long-term.
- Root `/favicon.ico`, `/site.webmanifest` have stable names too. NOT safe to cache long-term.

A naive `Cache-Control: immutable` on all of `/assets` would pin EmulatorJS/Ruffle for a year, so a RomM update would ship a new player that browsers never load until the cache expired.

**To Reproduce**
1. Run RomM from the official Docker image.
2. `curl -sSI https://<romm>/assets/index-<hash>.js | grep -iE "cache|expires|etag"` -> only `ETag`/`Last-Modified`, no `Cache-Control`.
3. Same for a cover (`/assets/romm/resources/.../small.png?ts=...`), an EmulatorJS file (`/assets/emulatorjs/data/loader.js`), and `/favicon.ico`.
4. In DevTools > Network, load the library grid then reload: assets are re-requested/revalidated instead of served from cache.

**Expected behavior**
Send long-lived caching only for fingerprinted assets, and keep the rest revalidating:
- Content-hashed build output and `?ts`-versioned resources -> `Cache-Control: public, max-age=31536000, immutable`.
- `/assets/emulatorjs/**`, `/assets/ruffle/**`, and root favicons/manifest -> a short cache or revalidation (e.g. `no-cache`/`must-revalidate`) so RomM updates propagate immediately.
- `index.html` (served from `location /`, not `/assets`) stays revalidated as today, so new deploys are picked up.

**Server:** OS: any host on the official Docker image. Version: 5.1.0-alpha.1
**Client:** Any browser.

**Additional context**
Relevant config is `docker/nginx/templates/default.conf.template`:
```
location /assets {
    try_files $uri $uri/ =404;
}
```
It sets no caching at all. A safe fix could either (a) add `immutable` on `/assets/` while overriding `/assets/emulatorjs/` and `/assets/ruffle/` with a revalidating policy, or (b) match only content-hashed filenames. Note EmulatorJS/Ruffle are added to `/assets` at build time (`docker/Dockerfile`, `COPY ... /assets/emulatorjs` and `/assets/ruffle`), which is why they must be excluded from any long-cache rule.

Spinnich:
**Description**

Fixes #3890.

## Summary (what and why)

RomM's bundled nginx served everything under `/assets` (the frontend JS/CSS bundle and every cover, screenshot, and manual), plus the root favicons and web manifest, with **no `Cache-Control` header**. Browsers therefore re-fetch or revalidate (a `304` round-trip) each of these on every repeat visit. It is most visible in the library grid, where each ~100 KB cover is re-checked; a long scroll is tens of MB of covers alone.

A blanket `immutable` rule is unsafe, because `/assets` mixes assets with different versioning. This change caches **by fingerprint**:

| Assets | `Cache-Control` | Why it is safe |
| --- | --- | --- |
| Content-hashed build output (`index-<hash>.js`) | `public, max-age=31536000, immutable` | The filename changes whenever the content changes |
| Library resources with a recognized version param: covers (`?ts=`) and manuals (`?v=`) | `public, max-age=31536000, immutable` | The URL changes whenever the file changes |
| Query-less library resources (metadata screenshots, RA badges), the bundled EmulatorJS/Ruffle runtimes, and other static files under `/assets` | `public, max-age=3600, must-revalidate` | Stable URLs across image upgrades or re-scrapes; revalidate to pick up a change promptly |
| `index.html`, the service worker, and the root favicons/manifest (served from `location /`) | `no-cache` | A new deploy's fingerprinted bundles and any updated runtime must load at once |

The library-resources split is driven by an `$args` map that recognizes the `ts` and `v` version params: a resource is frozen only when one of those is present, not on any arbitrary query (an unrelated query such as `?download=1` does not freeze a stable file). Covers carry `?ts=` and manuals carry `?v=`; metadata screenshots and RA badges reuse stable filenames like `0.jpg` and are overwritten in place, so they revalidate.

## A note on "screenshots" (avoiding an ambiguity)

RomM has two distinct "screenshot" concepts, and only one is affected here:

- **Metadata screenshots** (`Rom.merged_screenshots`): scraped game media, written to `/assets/romm/resources/roms/<platform>/<rom>/screenshots/<idx>.jpg` with a stable name and no query. These are covered by this change and now `must-revalidate`, so a re-scrape that overwrites `0.jpg` is picked up.
- **User-uploaded (gallery) screenshots** (the `Screenshot` asset): served from `/api/screenshots/<id>/content?timestamp=<updated_at>`, which nginx proxies to the backend. These are **not** under `/assets` and are **not** touched by this change.

So no user-uploaded content is affected by the `/assets` rules. Note that user-uploaded covers and manuals do live under `/assets/romm/resources/`, but they carry `?ts=` / `?v=` and so remain correctly `immutable` (the query changes when `updated_at` bumps on re-upload).

## Files modified

- **`docker/nginx/templates/default.conf.template`**
  - Added a `map $args $resources_cache_control` that yields `immutable` only when a recognized version param (`ts` or `v`) is present, and `must-revalidate` otherwise (so an unrelated query does not freeze a stable file).
  - Split the single `location /assets { try_files ... }` into three: a regex `location` that freezes content-hashed bundles as `immutable`, a `location /assets/romm/resources/` that applies the map above, and the existing `location /assets` as the `must-revalidate` catch-all for stable-named runtimes and icons.
  - Added `add_header Cache-Control "no-cache";` to `location /` so `index.html`, the service worker, and the root favicons/manifest revalidate. All pre-existing headers on that block (CORS and COEP/COOP) are unchanged.
- **`backend/tests/test_nginx_cache_headers.py`** (new)
  - Parses the template's `location` and `map` blocks and replays nginx's location selection (exact, longest-prefix, first-matching-regex) plus the `$args` version-param map, asserting the `Cache-Control` each representative URL resolves to. Covers hashed bundles, cover/manual (`?ts=` / `?v=`), metadata screenshots and RA badges (query-less and unrelated-query), the EmulatorJS/Ruffle runtimes, `index.html`, and the root favicons/manifest/service worker. Pure static parsing, so it is DB-agnostic.

## Testing

- **Unit test:** 29 cases, all pass. Runs in the existing pytest workflow (MariaDB and PostgreSQL matrices).
- **Real nginx validation:** rendered the template with `envsubst` inside the shipped image `rommapp/romm:5.1.0-alpha.1` (which carries the custom js/zip modules), ran `nginx -t` (syntax OK), then started nginx and curled each URL class, including the unrelated-query edge case:

```text
/assets/index-C-Cea5tF.js                            -> public, max-age=31536000, immutable
/assets/romm/resources/.../cover/small.png?ts=123    -> public, max-age=31536000, immutable
/assets/romm/resources/.../manual/1.pdf?v=123        -> public, max-age=31536000, immutable
/assets/romm/resources/.../screenshots/0.jpg         -> public, max-age=3600, must-revalidate
/assets/romm/resources/.../screenshots/0.jpg?download=1 -> public, max-age=3600, must-revalidate
/assets/romm/resources/.../ra/badges/12345.png       -> public, max-age=3600, must-revalidate
/assets/emulatorjs/data/loader.js                    -> public, max-age=3600, must-revalidate
/  /index.html  /favicon.ico  /site.webmanifest  /sw.js  -> no-cache
```

- **Full backend suite:** `uv run pytest -n 4` green (2304 passed, 0 failed).
- **Lint:** `trunk fmt && trunk check` clean (ruff, black, isort, mypy).

## Reviewer notes

- **No dropped headers.** `/assets` had no `add_header` on `master`, so nothing is removed; `location /` retains its CORS and COEP/COOP headers and only gains `Cache-Control`. The EmulatorJS player's cross-origin isolation is unaffected (same-origin runtime files need no CORP under `require-corp`).
- **Resources are a mix, matched by param name.** Only covers (`?ts=`) and manuals (`?v=`) carry a recognized version param and are frozen; metadata screenshots and RA badges are query-less and revalidate. The map matches the `ts`/`v` param names at a boundary (`(^|&)(ts|v)=`), so an unrelated query such as `?download=1` does not freeze a stable resource. See the "screenshots" note above for the metadata-vs-upload distinction.
- **Two revalidation policies by design.** `must-revalidate` with a 1h `max-age` for the `/assets` bulk avoids a `304` storm across many small icons, while `no-cache` on `location /` keeps `index.html` and `sw.js` always fresh (a `max-age` there could serve a stale `index.html` referencing purged bundles).
- **Vite hash assumption.** The immutable regex expects Vite's default 8+ char hash. A shorter hash would simply route a bundle to the `must-revalidate` bucket: safe and non-breaking, never stale.
- **Test placement.** There is no existing infra-config test harness, so the guard lives under `backend/tests/` (the available runner). It is a fast regression check; the authoritative verification is the `nginx -t` and curl run above.
- **Reverse proxies.** These headers reach the browser as long as a proxy passes them through (the default). Anyone who previously worked around the missing headers with a blanket cache rule at their proxy should drop it so these per-type headers govern.

## AI assistance disclosure

This change was implemented with AI assistance (Claude). The nginx caching policy, the test, and the commits were AI-authored under maintainer direction, reviewed locally, and verified empirically against the shipped nginx image as described above.

**Checklist**

N/A. This is a response-header change; the `curl` output above is the relevant evidence.

### DIFF
--- a/backend/tests/test_nginx_cache_headers.py
+++ b/backend/tests/test_nginx_cache_headers.py
@@ -1,0 +1,286 @@
+"""Guards the asset caching policy in the nginx template (issue #3890).
+
+The Docker image's nginx serves the whole frontend. Assets fall into caching
+classes and the template must sort each request into the right one:
+
+  * Frozen forever (fingerprinted): content-hashed Vite bundles
+    (``index-<hash>.js``) and library resources requested with a cache-busting
+    query (covers ``?ts=``, manuals ``?v=``). The URL changes when the file
+    does, so a cached copy is never stale.
+  * Revalidating (stable URLs across image upgrades / re-scrapes): the bundled
+    EmulatorJS/Ruffle runtimes, icons/fonts/logos, and the query-less resources
+    (screenshots, RA badges, which reuse names like ``0.jpg`` and are
+    overwritten in place).
+  * ``no-cache`` (must always be fresh): ``index.html``, the service worker,
+    and the root favicons/manifest.
+
+The resources bucket splits on whether a recognized version param (``ts``/``v``)
+is present, via the ``$resources_cache_control`` map (an unrelated query like
+``?download=1`` must not freeze a stable file). Rather than grepping the
+template for a string, this parses the ``location`` and ``map`` blocks and
+replays nginx's matching for representative URLs, asserting the
+``Cache-Control`` each resolves to.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+NGINX_TEMPLATE = (
+    Path(__file__).resolve().parents[2]
+    / "docker"
+    / "nginx"
+    / "templates"
+    / "default.conf.template"
+)
+
+IMMUTABLE_MAX_AGE = 31536000  # one year, in seconds
+# Any revalidating cache must stay well under this so upgrades/re-scrapes show.
+MAX_REVALIDATING_AGE = 86400  # one day, in seconds
+
+
+def _read_block(text: str, start: int) -> tuple[str, int]:
+    """Return (body, end) for a brace block whose opening ``{`` is at start-1,
+    scanning to the matching ``}``. Blocks in this template do not nest."""
+    depth = 1
+    i = start
+    while i < len(text) and depth:
+        if text[i] == "{":
+            depth += 1
+        elif text[i] == "}":
+            depth -= 1
+        i += 1
+    return text[start : i - 1], i
+
+
+class Location:
+    """A parsed nginx ``location``: modifier, match, and the Cache-Control it
+    sets (a quoted literal, or a ``$variable`` resolved through a map)."""
+
+    def __init__(self, modifier: str, match: str, body: str):
+        self.modifier = modifier
+        self.match = match
+        cache = re.search(
+            r'add_header\s+Cache-Control\s+(?:"([^"]*)"|(\$[A-Za-z0-9_]+))',
+            body,
+            re.IGNORECASE,
+        )
+        self.cache_control = (
+            (cache.group(1) if cache.group(1) is not None else cache.group(2))
+            if cache
+            else None
+        )
+
+    @property
+    def is_regex(self) -> bool:
+        return self.modifier in ("~", "~*")
+
+    @property
+    def is_prefix(self) -> bool:
+        return self.modifier in ("", "^~")
+
+
+def _parse_locations(text: str) -> list[Location]:
+    """Extract every ``location`` block, preserving file order (nginx evaluates
+    regex locations in the order they appear)."""
+    locations: list[Location] = []
+    header = re.compile(r'location\s+(=|\^~|~\*|~)?\s*(?:"([^"]+)"|(\S+))\s*\{')
+    for m in header.finditer(text):
+        modifier = m.group(1) or ""
+        match = m.group(2) if m.group(2) is not None else m.group(3)
+        body, _ = _read_block(text, m.end())
+        locations.append(Location(modifier, match, body))
+    return locations
+
+
+def _parse_maps(text: str) -> dict[str, tuple[str, dict[str, str]]]:
+    """Parse ``map $in $out { key value; ... }`` -> {out: ($in, {key: value})}."""
+    maps: dict[str, tuple[str, dict[str, str]]] = {}
+    entry = re.compile(r'(?:"([^"]*)"|(\S+))\s+(?:"([^"]*)"|(\S+))\s*;')
+    for m in re.finditer(r"map\s+(\$[A-Za-z0-9_]+)\s+\$([A-Za-z0-9_]+)\s*\{", text):
+        input_var, output_var = m.group(1), m.group(2)
+        body, _ = _read_block(text, m.end())
+        entries: dict[str, str] = {}
+        for e in entry.finditer(body):
+            key = e.group(1) if e.group(1) is not None else e.group(2)
+            value = e.group(3) if e.group(3) is not None else e.group(4)
+            entries[key] = value
+        maps[output_var] = (input_var, entries)
+    return maps
+
+
+def _input_value(input_var: str, url: str) -> str:
+    """The nginx source a map keys on. Only ``$args`` is used here."""
+    query = url.split("?", 1)[1] if "?" in url else ""
+    if input_var == "$args":
+        return query
+    if input_var == "$is_args":
+        return "?" if query else ""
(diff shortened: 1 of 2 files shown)


Spinnich:
Thanks @greptile-apps, good catch. Keying on `$is_args` was too broad: any query string would have frozen a stable resource.

Fixed in 61c3dd1e4 by keying on `$args` and only treating a recognized version parameter as a fingerprint:

```nginx
map $args $resources_cache_control {
    default         "public, max-age=3600, must-revalidate";
    "~(^|&)(ts|v)=" "public, max-age=31536000, immutable";
}
```

So `?ts=` (covers) and `?v=` (manuals) freeze, while everything else revalidates. Verified against real nginx in the shipped image, including the edge cases:

- `screenshots/0.jpg?download=1` -> `must-revalidate` (your example)
- `screenshots/0.jpg?foo=ts` (value contains `ts`) -> `must-revalidate`
- `screenshots/0.jpg?nav=1` (param name contains `v`) -> `must-revalidate`
- `cover/small.png?ts=1&extra=2` -> `immutable`

The `(^|&)(ts|v)=` boundary matches only the recognized parameter names, not arbitrary substrings. Also added the unrelated-query case to `test_nginx_cache_headers.py`, per your second point.

gantoine (approved):

PR #3896, opened 2026-07-23
Add music favorites and playlists API

tmgast:
## Description

The music query API (#3640) makes soundtracks searchable and browsable, but nothing a user marks is stored: no liked tracks, no playlists. This adds both, modeled on the collections feature. The key differs because track identity is unstable: scans purge and re-insert `rom_files`, so `rom_file_id` churns on every COMPLETE or HASHES scan, while `rom_id` and file content hashes are stable. Playlist and favorite entries therefore store `(rom_id, md5_hash)` and resolve to the current file at query time. Entries survive rescans and renames, self-heal when a file returns, and identical-content duplicates within a rom resolve to a single row.

For that key to be reliably present, soundtrack audio files are now always hashed individually during scans, including on `NON_HASHABLE_PLATFORMS` and with `SKIP_HASH_CALCULATION` enabled (the file-size rationale behind both exclusions does not apply to audio), and the soundtrack upload endpoint computes crc/md5/sha1 on write. Existing libraries backfill with one HASHES scan; adding an unhashed track returns a 400 that names the fix.

### Changes

- **Favorites.** `music_favorite_tracks` table (PK `user_id`, `rom_id`, `md5_hash`) with idempotent add/remove endpoints; `is_favorite` and `md5_hash` now included on every music track row.
- **Playlists.** `music_playlists` and `music_playlist_tracks` tables; CRUD with per-user name uniqueness, `is_public` read-only sharing, position-ordered membership with no duplicates, and full-order reorder. Row-level visibility (hidden platforms/roms) applies to public playlists through the viewer's own permissions.
- **Scopes.** New `playlists.read` / `playlists.write`, wired through the granular permission system (`PermEntity.PLAYLISTS`). Migration `0102` mirrors each group's existing collections grants onto the playlists entity, so current users keep working without admin intervention; the permission parity tests pass unchanged.
- **Scan/upload hashing.** Audio files under soundtrack folders always get individual hashes; uploads hash on write. The rom-level combined hash is untouched.
- **`rom_id` filter on `GET /api/music/tracks`.** Clients can fetch one game's tracks by rom identity instead of exact album-tag match (tags are user data and frequently wrong or shared across games).
- **Frontend.** Regenerated API types only; no hand-written frontend changes.

### Tests

- Endpoint matrices for favorites and playlists: CRUD, ownership x public/private, scope enforcement, hidden-platform/rom filtering, duplicate names, reorder including partial orders and foreign-track rejection.
- Rescan durability: entries survive a simulated purge/re-insert with new file ids; dangling entries (changed content) are omitted from listings while stored counts persist.
- Scan hashing on non-hashable platforms and with hashing disabled; upload hashing on both new-file and reupload paths.
- Migration `0102` upgrade and downgrade verified on MariaDB and PostgreSQL.
- Exercised end-to-end against a device client, including rescan durability of favorites and playlists.

## New API Endpoints

| Method | Route | Scope | Purpose |
| --- | --- | --- | --- |
| GET | `/api/music/favorites` | `playlists.read` | Paginated liked tracks, same row shape and filters as `/tracks` |
| POST / DELETE | `/api/music/favorites` | `playlists.write` | Idempotent add/remove by `rom_file_ids` |
| GET / POST | `/api/music/playlists` | `playlists.read` / `write` | List own + public playlists; create |
| GET / PUT / DELETE | `/api/music/playlists/{id}` | `playlists.read` / `write` | Fetch, update (owner only), delete (owner only) |
| GET | `/api/music/playlists/{id}/tracks` | `playlists.read` | Paginated tracks ordered by position |
| POST / DELETE | `/api/music/playlists/{id}/tracks` | `playlists.write` | Append (no duplicates) / remove by `rom_file_ids` |
| PUT | `/api/music/playlists/{id}/tracks/order` | `playlists.write` | Full desired order; unlisted entries keep relative order after |

REQUEST `POST /api/music/playlists/3/tracks`

```json
{ "rom_file_ids": [1101, 1106] }
```

RESPONSE

```json
{ "added": 2 }
```

RESPONSE `GET /api/music/playlists` (excerpt)

```json
[
  {
    "id": 3,
    "name": "Chill",
    "description": null,
    "is_public": false,
    "user_id": 1,
    "owner_username": "admin",
    "track_count": 2,
    "created_at": "2026-07-23T05:10:00+00:00",
    "updated_at": "2026-07-23T05:12:00+00:00"
  }
]
```

Writes resolve `rom_file_ids` to `(rom_id, md5_hash)` server-side; a 400 names ids that are missing, not music tracks, or not yet hashed.

### Known Gaps

- Reuploading a soundtrack file with different content changes its md5, so md5-keyed entries pointing at the old content dangle until re-added. Inherent to content addressing.
- Per-user permission overrides are not mirrored to the new playlists entity by the migration (group grants are); admins can add overrides where needed.

---

> AI Disclosure

**Checklist**


gantoine (commented):
Heads up: I merged `master` in and pushed a commit that changes the track reference design here, since #3900 landed and invalidated the premise this PR was built on.

`sync_rom_files` now reconciles `rom_files` in place, matching existing rows by path first and then by content hash, writing only the columns that actually changed. A file renamed or moved inside a ROM keeps its id, its `created_at` and its track metadata. So rom file ids are stable across rescans now, and the `(rom_id, md5_hash)` indirection isn't needed.

`a84ad22` switches `music_playlist_tracks` and `music_favorite_tracks` to a `rom_file_id` FK. That let a fair amount of supporting machinery go, and the scan and upload paths are back to byte-identical with `master`:

- soundtrack audio is no longer force-hashed during scans (it was overriding both `NON_HASHABLE_PLATFORMS` and the user's `SKIP_HASH_CALCULATION` setting)
- soundtrack uploads no longer hash on write
- adding an unhashed track no longer 400s, and existing libraries no longer need a HASHES scan to backfill before they can use playlists
- the `min(file id)` correlated subquery that collapsed identical copies within a ROM is gone
- `md5_hash` is off `MusicTrackSchema` (the `rom_file_ids` request shape is unchanged, so nothing else moved on the frontend)

Net about 190 lines removed. Full backend suite is green.

Two behavior changes I'd like a second opinion on, since they're judgment calls rather than mechanical consequences:

1. **Identical copies are now distinct tracks.** Two files with the same content in one ROM are separately addressable instead of both resolving to the lowest file id. I think this is the more honest behavior, but it does mean a user can add what looks like "the same song" twice.
2. **A file leaving the library now removes its entries** via the FK cascade. The old design left a dangling entry that would resurrect if matching content came back. Worth noting that resurrection only ever fired on re-add anyway, because `sync_rom_files` hard-deletes unmatched rows and there's no `missing_from_fs` marking at the file level. Practical effect: playlist `track_count` and the resolvable page total now diverge only for viewer-hidden platforms.

If you'd rather keep either of those, say so and I'll adjust.

Two things I left for you rather than rewriting on your branch:

- **The PR description and the first commit message still describe the hash-keyed design.** They explain `(rom_id, md5_hash)`, the always-hash-audio behavior, and the HASHES-scan backfill, none of which is true anymore. The history is fine as a sequence, but that's what a reviewer reads first.
- **AI-assistance disclosure**, per `CONTRIBUTING.md`. Flagging for completeness on your side, and for transparency on mine: I used Claude Code for the rework I pushed.

gantoine (approved):

PR #3925, opened 2026-07-23
feat: add binary title ID and save ID extraction during scan

### LINKED ISSUE #2526: [Feature] Additional Info on Switch TitleIDs
Seeing issues of DLCs and Updates not always combining under the BASE game.

the TitleID of a switch title is unique across BASE, UPDATES, and DLC. There is a pattern.

Scan and group by Base TitleID
Base ID always ends in `000`
Updates always end in `800` (group these as update with the BASE)
DLC is always the Base ID with the 4th to last nibble increased by `0x1`
    and the last three nibbles can be `000-FFF` (group these as DLC with BASE)
Tinfoil also requires the titleID in brackets [] to be scanned properly: `Title [titleID][vVersion].ext` as a minimum.

**Example with BotW**
BASE: 01007EF00011E000
UPD:  01007EF00011E800
DLC1: 01007EF00011F001
DLC2: 01007EF00011F002

This is true across all `digital` titles. Multi-content physical cart dump identification is still up for debate in the community.

Edit: if it helps, Base TitleID's 4th to last nibble is always even, DLC's 4th to last nibble is always odd.

Edit2: This can also help with user's organization. They can sort their `/switch/roms` or `/roms/switch` subfolders however they see fit: `/BASE|UPD|DLC/` or `/Game Name/Base (update/updatefiles | /dlc/dlcfiles)`

tmgast:
## Description

RomM derives a game's identity from its filename during a scan, which is fragile in the cases that matter most. Base, update, and DLC files that do not carry a title ID in their name go uncategorized, Switch retail titles cannot be distinguished without decryption, and save-sync clients have no reliable way to pair a save to the folder an emulator actually creates on disk. This reads the platform-native identifier directly out of the ROM binary during the scan's file pass, using [argosy-sigil](https://github.com/rommforge/argosy-sigil) (a native helper library, MPL-2.0), and stores it as typed columns surfaced in the v2 game details.

Extraction covers the platforms sigil supports, listed in its [README](https://github.com/rommforge/argosy-sigil#supported-platforms), and runs independently of the hash-database gate so it reaches non-hashable platforms like Switch. sigil is an optional dependency: the backend imports it behind a guard, so an image or dev environment built without it skips extraction rather than failing.

RomM does not handle a user's console keys, so formats that need them to be read, such as an encrypted retail Switch dump, are left unidentified. A decrypted dump still resolves, since NCA headers are tried as plaintext first, and a client that holds the keys can extract locally and submit what it found through the identity endpoint below.

### Changes

- **Binary identity on the rom.** A rom gains `title_id`, `save_target`, and `save_target_layout`. `title_id` is the game's identity, for matching and display. `save_target` is the literal on-disk name an emulator gives the game's saves, so a save-sync client pairs by folder instead of guessing from filenames. Save layout belongs to a game rather than to one of its files, so these live on the rom; `RomFile.category` still carries the per-file update and DLC distinction.
- **`save_target_layout` tells a client how to apply the target.** The shape differs per platform: one exact folder, several folders sharing a prefix (PS2 splits a game across `BASLUS-20642SYS` and `BASLUS-20642RD0`), files sharing a prefix, or a nested path (the 3DS nests a save under `title/<high 8>/<low 8>/`). Storing the classification alongside the target means consumers dispatch on it rather than re-deriving the rule per platform.
- **Targets match the case the emulator writes.** `save_target` addresses a real directory, and azahar, dolphin and cemu all create theirs with lowercase hex. On case-sensitive storage an uppercase value addressed a *second* directory beside the emulator's own, so a game's saves silently split across the two. 3DS, Wii and Wii U store the path in the emulator's case while `title_id` keeps its canonical uppercase form.
- **Clients can submit what they extracted.** `PUT /api/roms/{id}/identity` (`roms.write`) stores the three fields for a ROM the scan could not read. A Switch update or DLC id resolves to its base first, because reassociating a renamed non-hashable ROM matches on `title_id` and an update id there would strand the entry.
- **Switch categorization from the binary.** The NCA program ID alone reports the base title for every content type, so an update is indistinguishable from its base. Reading the content metadata (CNMT) instead yields each file's authoritative ID, content type, and version, so `RomFileCategory` is set from the content type (application/patch/addon to GAME/UPDATE/DLC), which takes precedence over the folder name a file sits under when extraction succeeds, and leaves it untouched when it does not.
- **Opt-in Switch filename embedding.** A new `config.yml` flag `filesystem.embed_switch_title_ids` (default off) renames Switch files during a scan to embed the binary id and version (`Name [TITLEID][vN].ext`) and reconciles the database record. It is idempotent (skips names that already carry a title-id bracket), collision-safe, and Switch-only. This is the binary-driven form of what filename-parsing tools consume.
- **Packaging.** sigil builds in a dedicated Docker stage (the RAHasher pattern), pinned by commit, and is copied into the venv. The guarded import keeps it optional everywhere else.
- **Frontend.** Regenerated API types plus the Title ID and Save File/Folder Target rows on the game details Metadata tab; no other hand-written frontend changes.

### Tests

- Adapter: content-type and version mapping, absent-binding degradation, per-platform gating, and that no key path is ever passed.
- Switch title-id knowledge: base-id recognition, base derivation from an update or DLC id, content-type mapping, and the embedded-id patterns.
- Filesystem handler: nested update and DLC extraction, content-type-over-folder category precedence, rom-level base selection and derivation, and the per-platform `save_target_layout` mapping.
- Identity endpoint: storage, base-id normalization for a submitted update or DLC id, non-Switch serials stored verbatim, partial updates leaving other fields alone, and the not-found and unauthenticated paths.
- Embedding: rename format, idempotent skip, collision skip, non-Switch never renamed, and single-file (`Rom.fs_name`) versus multi-part reconciliation.
- The migration upgrades and downgrades on MariaDB and PostgreSQL.
- Verified end-to-end against real Switch, PS2, and 3DS dumps, including per-file categorization, the 3DS nested target, and the embedding rename with database reconciliation.

### Known Gaps

- Encrypted retail Switch dumps need decryption keys to read, which RomM does not handle, so they are not identified from the binary. Decrypted dumps are, and a client that holds the keys can extract locally and submit the result to the identity endpoint.
- Archived-content members are out of scope; extraction runs on served files, not entries inside archives.
- sigil flags its newer extractors as experimental until they have been validated against a broad sample of real discs. RomM stores their output like any other, so a low-confidence id is not distinguishable in the UI today.

---

> AI Disclosure

Closes #2526

**Checklist**


matthew-pye:
Hi, I noticed that there isn't Xbox 360 titleID extraction, not sure if you wanted to add that, if you do here is the code I wrote for PlayniteAchievements plugin ([TitleIDExtractor.zip](https://github.com/user-attachments/files/31151630/TitleIDExtractor.zip)) it is in C# so will need adapting to Python(?) but I don't know Python so couldn't give it you in that language. 
I will note that the .iso extraction portion of the code was written with heavy AI assistance (Due to me not full knowing the ISO format and old xbox 360 iso packers using the iso format incorrectly) where as the .xex portion I wrote myself. Also it doesn't support xenia's .zar format yet as I haven't looked into how that format is laid out.

Hopefully I'm not adding more work to your plate :)

backend/models/rom.py
```diff
@@ -610,6 +624,9 @@ class Rom(BaseModel):
     md5_hash: Mapped[str | None] = mapped_column(String(length=100))
     sha1_hash: Mapped[str | None] = mapped_column(String(length=100))
     ra_hash: Mapped[str | None] = mapped_column(String(length=100))
+    title_id: Mapped[str | None] = mapped_column(String(length=100))
+    save_id: Mapped[str | None] = mapped_column(String(length=100))
+    save_usage: Mapped[SaveUsage | None] = mapped_column(Enum(SaveUsage), default=None)
```
gantoine: can you make these two field names more specific? `save` is an existing model and common symbol in the repo already, and i can't figure out what it means at a glance.

backend/handler/filesystem/roms_handler.py
```diff
@@ -169,6 +178,113 @@ class ParsedRomFiles:
...
+            return base.title_id, base.save_id, _parse_save_usage(base.usage)
+
+        derived = _derive_switch_base_title_id(extractions[0].title_id)
+        if derived is None:
+            return None, None, None
+        # Switch saves are keyed by the base title id itself.
+        return derived, derived, SaveUsage.FOLDER_EXACT
+
+    first = extractions[0]
+    return first.title_id, first.save_id, _parse_save_usage(first.usage)
```
gantoine: all these siwtch-specific functions should live in a new switch utility file, and same goes for the ones below like `_extract_title_id`

backend/endpoints/sockets/scan.py
```diff
@@ -373,6 +381,29 @@ def _should_hash_firmware(
...
 
 
+def _resolve_prod_keys_path(platform: Platform) -> str | None:
+    """Locate the console's prod.keys, which sigil needs to decrypt Switch
+    XCI/NSP headers, among the platform's scanned firmware."""
+    if platform.slug not in SWITCH_PLATFORM_SLUGS:
+        return None
+
+    prod_keys = db_firmware_handler.get_firmware_by_filename(platform.id, "prod.keys")
+    return f"{LIBRARY_BASE_PATH}/{prod_keys.full_path}" if prod_keys else None
```
gantoine: this is scary, let's discuss on github async lol

gantoine (changes requested):

backend/models/rom.py
tmgast: It's the serial that maps to the save file or folder name. Basically the string representation of the base save file/folder name.

Maybe `save_basename` is okay, but that becomes inaccurate for folders. 🤷‍♂️

gantoine: maybe something like switch_* so we know its related to switch titleid?

tmgast: It's not just switch... it's all serial-based save files/folders. It's paired with `save_usage` to tell clients how the value is used to identify save files.

gantoine (approved):

PR #3931, opened 2026-07-24
feat(cloud-sync): support config/thumbnails/system as opaque per-user blobs

fmustafayaman:
## Summary

Builds on #3904. RetroArch's Cloud Sync also offers a `config/`, `thumbnails/`, and `system/` category alongside saves/states (Settings → Saving → Cloud Sync → Sync Configuration/Thumbnails/System Files). None of these belong to a ROM, so the current PR rejects them with 409 rather than answering a fake success — correct, but it means those three toggles simply can't be used.

This adds them as opaque per-user blobs instead of going through the asset/ROM matching #3904 introduces:

- `CLOUD_SYNC_BLOB_BASE_PATH` (new config path, alongside the existing library/resources/assets paths under `ROMM_BASE_PATH`).
- `FSCloudSyncBlobHandler` (`backend/handler/filesystem/cloud_sync_blob_handler.py`) — plain-file storage built on the existing `FSHandler` base (path validation, atomic writes, locking all come from there for free).
- `cloud_sync_handler.py`: `parse_cloud_sync_blob_path`, `blob_md5` (same Redis-cache-by-path+size+mtime approach as `asset_md5`), `build_blob_manifest_entries`.
- `cloud_sync.py`: GET/PUT/DELETE/MOVE now branch to blob handling before falling through to the asset logic.

Blobs are namespaced per RomM user (`fs_asset_handler.user_folder_path`), so two RetroArch installs syncing to the same RomM instance under different accounts never see each other's config/thumbnails/system files — matching how saves/states are already scoped.

Unlike asset hashes, blob hashes are always real MD5s of the file on disk (no `content_hash`-is-null problem to work around), still cached the same way since these files are typically tiny but numerous (thumbnail packs especially).

`test_rejects_unsupported_sync_root` now asserts against a genuinely unsupported root (`deleted/saves/...`) instead of `config/`, since `config/` is no longer rejected. A new `TestCloudSyncBlobPathParsing` + `TestCloudSyncBlobs` class covers path parsing, upload/download/delete, nested thumbnail paths, and manifest inclusion.

## Test plan

- [x] `uv run pytest backend/tests/endpoints/test_cloud_sync.py` — 49 passed (34 existing + 15 new/updated)
- [x] `uv run pytest backend/tests/` filtered to asset/save/state/filesystem/handler suites — 1119 passed, no regressions
- [x] `ruff check` / `ruff format --check` clean on all changed files
- [ ] Not yet exercised against a real RetroArch client with Sync Configuration/Thumbnails/System Files enabled — same caveat as #3904 itself

## AI assistance disclosure

Written with Claude Code, building directly on the design and conventions established in #3904.

### DIFF
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -36,6 +36,10 @@ ROMM_TMP_PATH: Final[str | None] = _get_env("ROMM_TMP_PATH")
 LIBRARY_BASE_PATH: Final[str] = f"{ROMM_BASE_PATH}/library"
 RESOURCES_BASE_PATH: Final[str] = f"{ROMM_BASE_PATH}/resources"
 ASSETS_BASE_PATH: Final[str] = f"{ROMM_BASE_PATH}/assets"
+# Opaque storage for RetroArch Cloud Sync categories RomM has no concept of
+# (config/, thumbnails/, system/) — unrelated to any ROM, so it lives outside
+# the asset tree but still under the same persistent volume.
+CLOUD_SYNC_BLOB_BASE_PATH: Final[str] = f"{ROMM_BASE_PATH}/cloud_sync_blobs"
 ZIP_CACHE_PATH: Final[str] = f"{ROMM_BASE_PATH}/cache/zips"
 FRONTEND_RESOURCES_PATH: Final[str] = "/assets/romm/resources"
 
--- a/backend/endpoints/cloud_sync.py
+++ b/backend/endpoints/cloud_sync.py
@@ -10,6 +10,8 @@ from a fixed-size buffer, and a large body has been observed to corrupt its
 heap.
 """
 
+import os
+
 from fastapi import APIRouter, Request, Response, status
 from fastapi.responses import JSONResponse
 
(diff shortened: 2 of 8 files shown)


fmustafayaman:
Psp suppont not added yet. Waiting for review.

fmustafayaman:
Added a second commit: normalizes the emulator/core-name casing between RetroArch's cloud-sync path segment (its own directory casing, e.g. `Snes9x`) and RomM's `emulator` field convention (lowercase libretro core id, e.g. `snes9x` — the exact convention RomM's own web player matches saves against, confirmed against `EmulatorJS.vue`'s exact-match filter and `_EJS_CORES_MAP`).

Without this, a save synced in over cloud-sync would store `emulator="Snes9x"` verbatim, making it invisible to RomM's web player (which only ever looks for the lowercase form), and the manifest would hand RetroArch back whatever casing happened to be stored instead of its own local directory name.

`handler/cloud_sync_emulator_names.py` adds a small table (ported from the community `romm-retroarch-sync` project, same source the shim this PR builds on already credits) plus a plain, table-free lowercase/underscore normalization for the write direction. Cores outside the table round-trip unchanged rather than guessing at a casing that hasn't been verified against a real install.

10 new tests (table lookups both directions + a manifest round-trip test), 60/60 passing, no regressions in the broader asset/save/state suite.

PR #3933, opened 2026-07-24
fix(cloud-sync): remap web-player states into RetroArch's numbered load slots

fmustafayaman:
## Summary

Follow-up to #3931 (merged). Found while live-testing that PR's merged branch against a real RetroArch client: a state created through RomM's own web player (EmulatorJS) carries a display label + timestamp in its file name, not a RetroArch slot number. Since the `State` model has no `slot` column (unlike `Save`), such a state was surfaced in the cloud-sync manifest under its raw file name.

Reproduced live: RetroArch's Load State menu only ever offers numbered slots 0-999. A web-player state synced this way downloads successfully to disk but is never reachable from that menu -- it just silently sits in the sync folder.

## Fix

Mirrors the equivalent (already-shipped, already-live) fix in the `retroarch-webdav-romm` shim project's manifest builder: states are grouped by `(rom, emulator, slot suffix)`, where the slot suffix is derived from the file name's trailing `.state` / `.state<N>` / `.state.auto` pattern. The newest state in each bucket -- regardless of whether it came from RetroArch itself or the web player -- is advertised in the manifest under the canonical name RetroArch's own upload for that slot would carry (`<rom>.state`, `<rom>.state3`, `<rom>.state.auto`, ...).

GET/DELETE resolve that canonical name back to the real underlying row via a new `resolve_state_by_slot`, so serving and deleting always agree with what the manifest advertised, even when the underlying row's actual `file_name` differs from the canonical one.

Saves are untouched -- they already have a real `slot` column and are filtered correctly (`slot is None` excludes web-player saves from the manifest).

## Test plan
- [x] 65/65 cloud-sync tests passing (new: `TestCloudSyncStateSlotResolution`, `test_remaps_web_player_state_to_canonical_slot`, `test_newest_state_in_a_slot_wins_regardless_of_origin`)
- [x] Full backend suite: 2385 passed, 2 skipped, 0 failed -- no regressions
- [x] `ruff check` clean
- [x] Verified live against a real RetroArch 1.22.2 client synced to a deployed test instance running this branch: reproduced the bug (web-player state downloaded but unreachable from Load State), confirmed the fix resolves it

### DIFF
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -1,3 +1,4 @@
+import json
 import os
 from typing import Final, overload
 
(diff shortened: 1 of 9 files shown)


fmustafayaman:
Verified the PSP bundling live against a real PPSSPP capture (Patapon (Europe), serial UCES00995) synced from an actual RetroArch 1.22.2 client to a deployed instance of this branch:

- Both save slots (`UCES00995_DATA00`, `UCES00995_DATA01`) resolved to the correct rom automatically via PARAM.SFO title matching — no PSP_SERIAL_MAP override needed.
- Reproduced the exact out-of-order-upload scenario the retroarch-webdav-romm shim's `pspSave.ts` already documented: `DATA01/ICON0.PNG` reached the server before `DATA01/PARAM.SFO` and got a 409 (folder not yet resolved), same as an unbuffered implementation would. Once `PARAM.SFO` arrived and resolved the rom, the buffered `ICON0.PNG` was correctly folded into the bundle — confirmed via the server manifest afterward, which lists all three members (`ICON0.PNG`, `PARAM.SFO`, `SECURE.BIN`) for both save slots with correct hashes.
- A PPSSPP save-state and its screenshot (`Patapon (Europe) (En,Fr,De,Es,It).state[.png]`) synced correctly alongside the PSP saves.
- Full sync: 9 files uploaded, 0 permanent failures (the one transient 409 above self-resolved within the same sync pass).

No code changes from this — just confirming the buffering/merge logic (previously only unit- and mock-tested) holds up against real out-of-order PPSSPP upload traffic, not just the synthetic ordering my tests assumed.

PR #3937, opened 2026-07-24
feat(cloud-sync): read-only WebDAV browsing (PROPFIND) for the rom library

fmustafayaman:
## Summary

Adds read-only WebDAV directory browsing (`PROPFIND`, plus a fake `LOCK`/`UNLOCK` handshake some clients require) to the same `/api/cloud-sync` WebDAV surface RetroArch's Cloud Sync uses. This is unrelated to RetroArch's own sync traffic -- verified against its source, RetroArch's Cloud Sync client never issues PROPFIND -- it's purely so a real WebDAV client (iOS Files' "Connect to Server", Cyberduck, ...) can mount the same URL and browse the rom library (and current saves/states) as plain files.

Ports the `retroarch-webdav-romm` shim's `romBrowser.ts` + `webdavXml.ts` design:

- `roms/<platform_fs_slug>/<file>` -- lists platforms with roms, then roms within a platform, using the same display-name logic as the shim (`fs_name_no_ext.zip` for multi-file roms, the real per-file name for single-file roms). **Read-only**: no PUT/DELETE for this tree.
- `saves/` / `states/` -- browses the *current* cloud-sync manifest as a virtual directory tree. Unlike the shim, this only shows current entries, not full per-revision history (RomM's own web UI already covers save/state history; re-deriving that here felt out of scope for a browsing convenience).
- `LOCK`/`UNLOCK` -- fake, always-succeeds. Some WebDAV clients (iOS Files by report) won't complete "Connect to Server" without a server that at least answers these, even read-only.

For actually downloading a rom file, `GET`/`HEAD` redirect (307) to RomM's existing `/api/roms/{id}/content/{file_name}` endpoint rather than reimplementing Range support, the multi-file zip cache, and (in production) nginx's `X-Accel-Redirect` -- that endpoint already accepts HTTP Basic Auth alongside OAuth, so the same credentials carry over on the redirect.

## Found along the way

`Rom.has_multiple_files`/`.files` depend on a deferred `column_property` and a relationship that `get_roms_scalar` doesn't eager-load -- accessing them outside that query's own session raised `DetachedInstanceError`. Fixed by re-fetching the visibility-filtered rom ids via `get_roms_by_ids` (which does eager-load them via `@with_details`) instead of using the raw scalar results directly.

## Test plan
- [x] 71/71 cloud-sync tests passing (11 new: PROPFIND for root/roms/platform/rom-file/saves-states, LOCK/UNLOCK, GET redirect, 404s, visibility)
- [x] `ruff check` clean

### DIFF
--- a/backend/endpoints/cloud_sync.py
+++ b/backend/endpoints/cloud_sync.py
@@ -11,11 +11,13 @@ heap.
 """
 
 import os
+import uuid
+from urllib.parse import quote
 
 from fastapi import APIRouter, Request, Response, status
-from fastapi.responses import JSONResponse
+from fastapi.responses import JSONResponse, RedirectResponse
 
-from handler import cloud_sync_handler
+from handler import cloud_sync_handler, webdav_browser
 from handler.auth.constants import Scope
 from handler.auth.dependencies import get_permissions
 from handler.cloud_sync_handler import MANIFEST_FILE_NAME, AssetKind, CloudSyncPath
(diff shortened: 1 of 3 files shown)


fmustafayaman:
Closing in favor of merging this into #3933 instead, to keep everything on one branch.

PR #3957, opened 2026-07-26
Hide v2 actions the backend won't allow, with Playwright cover in CI

### LINKED ISSUE #3954: [Bug] Viewer role options visibility
**RomM version**
TrueNAS v1.2.31 RomM v5.0.0

**Describe the bug**
Several options are viewable to a user with the "viewer" role that are not intended to be, like changing role or editing/deleting games. They are not actually usable (i.e. trying to change role from user to administrator on the profile setting page will say "User [username] updated successfully" but will reset to user after refreshing the page; trying to delete a game will cause the page to reload itself and lead back to the login page, also with no actual effect to said game) but neither should be viewable as options to begin with. 

**To Reproduce**

1. Create a user with the Viewer role
2. Log in as the user
3. Try using any options that are only usable by an administrator

**Expected behavior**
None of these options should be viewable to begin with

**Screenshots**

https://github.com/user-attachments/assets/bb4a72b9-b31c-4eb6-bda6-31a922f0e6d3

https://github.com/user-attachments/assets/57220bf8-63b8-486a-ab39-c731b986d7ea

- OS: TrueNAS Scale
- Version: 25.10.5

- Device: MacBook Pro 
- OS: 26.5.1
- Browser: Firefox
- Version: 153

**Additional context**

gantoine:
**Description**

Fixes #3954

A user with a read-only role was shown ROM write actions and an editable role picker that the backend refuses. Both fail badly rather than harmlessly:

- **The role picker lies.** `update_user` ignores `role` on a self-edit (`backend/endpoints/user.py:421`, _"You can't change your own role"_), so the PUT succeeds, the snackbar says "updated successfully", and the value silently reverts on reload. This affected admins too, not just read-only users.
- **Everything else logs you out.** `POST /roms/delete` requires `ROMS_WRITE`. The axios interceptor treats _any_ 403 as an expired session (`frontend/src/services/api/index.ts`), so it clears `romm_session` and redirects to `/login` — the "page reloads back to the login page" from the issue video.

Four commits:

**1. Gate the actions on the grants that authorise them.**

- `useGameActions` exposes `canEdit` / `canDelete` / `canMatch` / `canRefresh`, so every ROM action surface shares one permission source.
- `GameActionsList` hides match / refresh / edit / delete. Their leading dividers are conditional too, or a read-only user gets a menu trailing stray separators.
- Same bug class, same 403-to-logout trap, in the Media and Files tabs: `ManualSubtab` (upload / delete / redownload), `MediaTab` + `SoundtrackPanel` (soundtrack upload / delete), `ScreenshotsSubtab` (the shared **ROM** section only) and `FilesTab` (upload button). Every one of those endpoints gates on `ROMS_WRITE` while the UI didn't. The per-user **"My screenshots"** section is deliberately untouched — it writes user assets, not the ROM, and hiding it would be the regression in the opposite direction.
- The profile page drops the role select entirely. Nobody can change their own role, so it could only ever revert; the role already renders as a chip in the identity row, and admins change other users' roles from Settings → Administration.

No new translation keys.

**2. Require the write grant alongside delete grants** (from review feedback).

A bare `(ROMS, DELETE)` grant projects to **no** OAuth scope — `grants_to_scopes` has no row for it — but the delete endpoints gate on the matching `*_WRITE` scope. So a custom group granted delete-without-write got the delete affordance, a 403, and the logout. Every delete gate now requires both: ROM delete (⋯ menu and bulk selection bar), ROM file delete, and platform delete.

**3 & 4. A Playwright suite covering this, running in CI.**

See Verification.

---

**Verification**

Static: `npm run typecheck`, `npm run test` (575 pass), `check_i18n_locales.py`, `check_i18n_sorted.py`, `trunk check` — all clean.

Visibility is what this PR changes, and that depends on grants the backend resolves and `useCan` projects — which unit tests can't reach. So it adds a Playwright suite (`frontend/e2e/`) driving a real assembled instance:

| Spec                      | Covers                                                       |
| ------------------------- | ------------------------------------------------------------ |
| `rom-actions.spec.ts`     | ⋯ menu items, divider collapse, light theme                  |
| `profile-role.spec.ts`    | no self-editable role control; role still shown as a chip    |
| `game-media-files.spec.ts` | Media + Files upload/delete affordances                     |

Every spec asserts **both** sides — absent for `e2e_viewer` (the seeded "Viewer (legacy)" group), present for `e2e_admin`. A viewer-only assertion would pass against a page that failed to render at all, and wouldn't notice a fix that over-hides and breaks admins.

I checked the specs actually catch the regression instead of passing vacuously: reverting the source fix gives **8 failures of 15**, with the 6 admin-side guards staying green (they were true before the fix too).

**Wired into CI** (`.github/workflows/e2e.yml`): MariaDB + Valkey services → migrations → an offline scan of `backend/romm_test/library` (empty `metadata_sources`, so no provider calls) → fixture users → backend → Playwright, whose `webServer` starts the Vite dev server. `DEV_PORT` (5000) is both what `main.py` binds and what Vite's `/api` proxy targets, so neither side needs configuring. Report and `test-results/` upload as an artifact on failure. **Green on the runner: 15/15 in 3.9m**, job ~5.7m.

Rehearsing the CI shape locally against a throwaway database — rather than pushing and hoping — caught three bugs in my own specs plus a broken workflow step:

- **Vite force-reloads the page** when it discovers a new dependency to pre-bundle (`optimized dependencies changed. reloading`), wiping the half-filled login form so the submit never lands. This, not slowness, caused every flake; my first two attempts fixed the wrong thing and one made it worse. Defended with a passive warm-up pass in `global-setup.ts` and a retry in `login()`.
- **`button[type="submit"]` matched two elements** — the collapsed reset-password form has its own. A strict-mode violation that only fired once that form had mounted, so it presented as intermittent.
- **Best-effort `.catch()` calls in the warm-up inherited a 120s timeout**, so each miss sat for two minutes and setup took 8. Now bounded at 15s.
- **`uv sync` with `working-directory: backend`** would have failed immediately: `pyproject.toml` is at the repo root.

`frontend/e2e/README.md` documents the setup and the flakiness signature to recognise (a couple fail, they pass in isolation, the set changes between runs).

---

**Not in scope**

1. **403 means two different things.** RomM returns 403 for both "not authenticated" and "not authorized" (Starlette's `requires` default), so the interceptor can't tell them apart and logs you out on any authorization failure. Hiding the controls removes the reachable path, not the trap. Fixing it properly means returning 401 for unauthenticated requests, which changes the API contract for every client — too broad to bundle here.

2. **Other scope mismatches found while auditing the v2 surface.** Most significant: the scan socket requires `TASKS_RUN` (`backend/endpoints/sockets/scan.py:951`) but the UI gates on `library.scan`, which `_ENTITY_ACTION_KEYS` derives from `(PLATFORMS, WRITE)`. The **Editor (legacy)** role has `platforms.write` but not `tasks.run` (verified against `grants_to_scopes`), so an editor sees the Scan button, passes the `/scan` route guard, and gets `scan:done_ko`. `MissingGamesSection`'s `cleanup_missing_roms` has the same mismatch, and firmware has no `ActionKey` at all so `FirmwareTab` gates on `platform.edit` while the endpoints need `FIRMWARE_WRITE`. Happy to open a follow-up PR — the cleanest fix for the scan case is moving `LIBRARY_SCAN` to the `(TASKS, WRITE)` row so `useCan("library.scan")` becomes honest everywhere at once.

---

**Checklist**

- [x] I've tested the changes locally <sup>(static checks, unit tests, and a Playwright run against a live instance — see Verification)</sup>
- [x] I've added unit tests that cover the changes <sup>(plus the e2e specs above)</sup>

Read-only user, Media → Manual: the plain empty state replaces the dropzone — no CTA, no drag-and-drop hint, no Upload button.

Read-only user, Media → Screenshots: the shared **ROM screenshots** section is gone, while **My screenshots** keeps its dropzone (it writes per-user assets, not the ROM).

<sup>Both captured during the Playwright run; the specs assert exactly these states.</sup>

---

**AI assistance disclosure**

This PR was written primarily by Claude Code (Opus 5), including the investigation, the code, the tests, the CI workflow and this description. Reviewed by me before opening.


gantoine:
@greptileai

PR #3964, opened 2026-07-27
feat: add managed SMB library sharing

Danilop95:
**Description**

This PR adds optional SMB support to RomM so users can access their ROM library from other devices on the local network without keeping a separate copy of every game on each device.

The main use case is connecting systems such as Windows PCs, Linux machines, Steam Deck, Batocera, RetroBat, homebrew consoles, and other emulation devices directly to the library managed by RomM.

Samba runs in a separate sidecar container instead of being installed inside the main RomM container. RomM communicates with the SMB controller through a restricted Unix socket, without exposing the Docker socket to the application.

The new SMB administration page is available from the `SYSTEM` section and provides:

- SMB service status and connection details.
- Server address, port, workgroup, and share information.
- Limited Samba logs for troubleshooting.
- Starting and restarting the SMB service when the sidecar is available.
- Creation and management of dedicated SMB users.
- Password generation and rotation.
- Separate SMB credentials that are not linked to RomM login passwords.
- Read-only or read-write permissions for each platform.
- Bulk actions to select all platforms as read-only or read-write.
- Platform search and clearer permission controls.
- Immediate permission updates for active SMB connections.
- Persistent SMB user management across container restarts.

The implementation also includes:

- Backend endpoints and response schemas.
- Database models and an Alembic migration.
- A dedicated Samba controller.
- Docker and Docker Compose configuration.
- Frontend v2 integration.
- Generated TypeScript API models.
- Translations for all currently supported locales.
- Backend tests for SMB users and platform permissions.

The Samba container is configured with restricted capabilities, guest access disabled, encrypted SMB connections, protected system accounts, limited logs, and persistent internal state.

SMB is intended for trusted local networks or VPN access. It should not be exposed directly to the Internet.

**Current limitations and follow-up work**

There are still some usability improvements that I would like to add in a follow-up:

- Allow administrators to enter their own SMB password when creating or updating a user, instead of only using an automatically generated password.
- Improve the way shares and platform folders are named inside Samba, including clearer and more configurable folder names.
- Add more control over how platform names are converted into SMB-safe share paths.
- Improve the installation flow when the SMB sidecar has not been deployed yet.
- Add more detailed documentation for connecting from different clients and operating systems.

These improvements are not required for the current implementation to work, but they would make the feature easier to configure for different libraries and devices.

**Checklist**

##### SMB access page and service status

SMB access page

SMB service and connection details

##### SMB user creation and permissions

Create SMB user

Platform permission selection

##### User credentials and management

SMB user credentials

SMB user management

**AI assistance disclosure**

AI-assisted development tools were used during repository analysis, implementation, testing, and code review. The final changes were manually reviewed and tested before submission.


bphenriques:
Hi, I am not a maintainer but as a avid user of Romm that uses SMB. Why does Romm need to concern about files over SMB when the operating system could do that?

I think this project should be filesystem agnostic and let the users decide how they expose the information to the container. In my case, I am using read only [readonly docker volumes](https://github.com/bphenriques/dotfiles/blob/fc40e66aac3044a5be63f0cb25987529213eb8ee/hosts/compute/selfhost/services/romm/romm.nix#L220) - can provide more references if you need further support.

It is up for the maintainer to decide, but I think that, to some extent, we need some level of restraint to avoid excessive complexity.

Danilop95:
Thanks for the feedback. I think you make a good point.

Perhaps this does not need to be part of RomM core. Another option could be to keep it as a separate companion service or community integration, and add documentation explaining how to use it with RomM.

I would be happy to maintain the SMB service myself. It has been very useful in my setup because I can access the same ROM library from my PS2 and Steam Deck without copying the games to each device or filling their internal storage.

Would a separate integration, plugin-style approach, or documentation page be more appropriate for the project?

gantoine:
@Danilop95 i think both of you have the right idea, leave this as-is and i'll riff on it after 5.1 is released.

Danilop95:
Sounds good. I’d be happy to help with it after 5.1, so just let me know. Thanks a lot!

gantoine:
tldr i think a good chunk of this can be extracted into a micro-app that interfaces with romm transparently, and we update the API where necessary to allow this to be seamless.

Danilop95:
Sounds good. I’ll keep looking into it and do my best to approach it properly. I’ll keep you posted!

gantoine:
gave this some thought and i agree with @bphenriques, this would work better as another app that interfaces with romm. going to close this as "won't do", and maybe someone will spin up something else.

PR #3970, opened 2026-07-27
feat(tasks): allow orphaned resource cleanup to run on a schedule

XenuIsWatching:
**Description**

Orphaned covers and screenshots left behind by deleted ROMs accumulate indefinitely, because `cleanup_orphaned_resources` lives in `tasks/manual/` and the only way to reclaim the space is to trigger it by hand. Every other cleanup task (`cleanup_netplay`, `cleanup_upload_tmp`, `cleanup_zip_cache`) runs on a schedule.

On a long-lived instance the resources directory grows without bound. That matters because the recommended `docker-compose` puts `romm_resources` and `mysql_data` in named volumes on the same filesystem, so unbounded artwork growth can fill the disk out from under the database.

This moves the task to `tasks/scheduled/`, converts it to a `PeriodicTask`, and adds an opt-in cron schedule:

- `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` (default `false`)
- `SCHEDULED_CLEANUP_ORPHANED_RESOURCES_CRON` (default `0 5 * * *`)

5:00 AM so it runs after the nightly scan (3 AM) and the metadata tasks (4 AM) have settled, rather than racing a scan that is actively writing artwork.

**One deliberate deviation from the other scheduled tasks**

The others set `enabled=ENABLE_SCHEDULED_<X>`. This task does not, and keeps `enabled=True`.

`run_task` in `endpoints/tasks.py` rejects a task unless *both* `enabled` and `manual_run` are set. Since this task is currently `enabled=True, manual_run=True`, gating `enabled` on a new env var that defaults to `false` would have silently removed the ability to run the cleanup by hand, which is how everyone uses it today.

So the cron string is what is opt-in, and `init()` is overridden to unschedule when it is unset. `schedule()` already returns `None` on a falsy `cron_string`, so nothing gets scheduled by default.

Happy to switch to the `enabled=ENABLE_...` form for consistency if you would rather accept that behavior change.

**Checklist**

Six tests in `tests/tasks/test_cleanup_orphaned_resources.py` covering the func path, that it stays manually runnable, cron off by default, cron set when enabled, and both `init()` branches. `tests/tasks` + `tests/endpoints/test_tasks.py`: 111 passed. Three failures in `TestPeriodicTask` are pre-existing and environmental (they need a live Redis); they fail identically on a clean checkout. Also verified by import that the task registers under `scheduled_tasks`, keeps `enabled`/`manual_run` true, and picks up the cron only when the env var is set.

`docs/BACKEND_ARCHITECTURE.md` and `env.template` updated.

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written with AI assistance (Claude Code). The AI performed the investigation, wrote the patch and the tests, and ran them locally. All changes were reviewed by me before submitting.


gantoine:
@greptileai

backend/tasks/scheduled/cleanup_orphaned_resources.py
```diff
@@ -40,17 +45,33 @@ def to_dict(self) -> dict[str, int]:
...
 
-class CleanupOrphanedResourcesTask(Task):
+class CleanupOrphanedResourcesTask(PeriodicTask):
     def __init__(self):
         super().__init__(
             title="Cleanup orphaned resources",
             description="Clean up orphaned resources in the ROMs directory",
             task_type=TaskType.CLEANUP,
+            # Always enabled so the task stays manually runnable; only the
+            # cron schedule is opt-in.
```
gantoine: remove comment

gantoine (changes requested):

backend/tasks/scheduled/cleanup_orphaned_resources.py
XenuIsWatching: Removed, thanks.

backend/config/__init__.py
```diff
@@ -237,6 +237,13 @@ def _get_env(var: str, fallback: str | None = None) -> str | None:
     "SCHEDULED_CONVERT_IMAGES_TO_WEBP_CRON",
     "0 4 * * *",  # At 4:00 AM every day
 )
+ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES: Final[bool] = safe_str_to_bool(
+    _get_env("ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES")
+)
```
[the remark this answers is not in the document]
XenuIsWatching: Good catch, this was real. Fixed in 9d09c82.

`docker/init_scripts/init` only starts `rqscheduler` when one of three flags is set, and mine was not among them, so enabling only the cleanup schedule would register the cron job and never run it. Added `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` to both the defaults block and the condition.

Worth flagging separately: that condition looks like it is missing other flags too. `ENABLE_SCHEDULED_CONVERT_IMAGES_TO_WEBP` and `ENABLE_SCHEDULED_RETROACHIEVEMENTS_PROGRESS_SYNC` are not in it either, and `cleanup_netplay` / `cleanup_upload_tmp` / `cleanup_zip_cache` are always-enabled `PeriodicTask`s that also need `rqscheduler` running. So on a default deployment those never fire unless the user happens to enable a rescan or a metadata update. Left alone here to keep this PR focused, but happy to open a separate PR if you want it fixed.

gantoine (approved):
nice

PR #3972, opened 2026-07-27
perf(tasks): walk resource directories with os.scandir

XenuIsWatching:
**Description**

Follow-up to #3970 and #3971. Now that `cleanup_orphaned_resources` runs on a schedule, its cost is worth looking at.

The task walks the resource tree through `anyio.Path`, which dispatches to a worker thread for every entry `iterdir()` yields and again for every `await entry.is_dir()`. On a library with ~30k ROM resource directories that is roughly 60k thread round trips, all to read information the directory listing already contains.

`os.scandir` returns `DirEntry` objects that carry the entry type from the directory read itself, so `is_dir()` needs no extra syscall. The walk moves into `_scan_resource_dirs()` and runs once via `asyncio.to_thread`, so the event loop stays free while the whole traversal happens in a single thread hop instead of tens of thousands.

**Measurements**

Real library, 28 platform directories / 30,262 ROM resource directories, Raspberry Pi 4, ext4 on a spinning disk:

| Implementation | Time |
| --- | --- |
| `anyio` `iterdir()` + `is_dir()` (current) | **50.3s** |
| `anyio` `iterdir()` alone, no `is_dir()` | 25.6s |
| `os.scandir` (this PR) | **0.15s** |

Cold and warm runs of the current code came out within 1% of each other (49.9s vs 50.3s), which is what identified this as thread-dispatch overhead rather than disk I/O. For reference the database half of the task, all 94 per-platform ROM queries, takes 0.83s, so the directory walk was ~98% of the runtime.

**Behavior notes**

- The ROM directories are now scanned before the empty-database guard from #3971 rather than after. The guard still short-circuits before any deletion; it just has the full map in hand when it does. At 0.15s for the whole tree this is strictly cheaper than the old code was for the platform level alone.
- `_numeric_subdirs()` logs and returns an empty set when a directory cannot be listed. Previously an unreadable platform directory raised and aborted the entire walk, so one bad directory stranded every other platform's orphans.

**Checklist**

Six new tests in `TestScanResourceDirs` cover the platform-to-ROM mapping, non-numeric names ignored, files ignored, an empty tree, a platform with no ROM directories, and an unreadable directory yielding an empty set instead of aborting. The four `TestCleanupOrphanedResourcesRun` tests from #3971 pass unchanged, which is the main assurance that behavior is identical. `tests/tasks/` + `tests/endpoints/test_tasks.py`: 121 passed. The three `TestPeriodicTask::test_get_existing_job_*` failures in my environment are pre-existing and need a live Redis; they fail identically on a clean checkout of `master`.

black, isort, ruff and mypy are clean on both changed files.

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written with AI assistance (Claude Code). The AI profiled the existing walk, wrote the patch and the tests, and ran the benchmarks and test suites locally. Reviewed by me before submitting.

### DIFF
--- a/backend/tasks/scheduled/cleanup_orphaned_resources.py
+++ b/backend/tasks/scheduled/cleanup_orphaned_resources.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import shutil
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


gantoine:
@greptileai

PR #3974, opened 2026-07-28
fix(scan): refuse concurrent scans and stop queued ones

XenuIsWatching:
**Description**

Found while debugging a real instance whose scan appeared to keep stopping. It never stopped: it had been running for five hours, with two more full scans queued behind it.

**Nothing prevents a second scan from being started**

`scan_handler` checks authorization and then enqueues, unconditionally. Whether a scan is already running is never considered.

That matters because the progress indicator is driven entirely by pushed socket events (`scan:scanning_rom` flips `scanning` on in `useScanLifecycle`). When the socket drops, which happens routinely on mobile as soon as the tab is backgrounded and the page is suspended, the client shows no scan in progress. Pressing scan again is the natural response, and it silently queues another full library scan. On the instance in question that produced three sequential scans of a 27k ROM library, all started from one browser.

`scan_handler` now refuses when a scan is running or queued, reporting it over the `scan:done_ko` channel that `reject_unauthorized_scan` already uses for a refused scan.

**Stopping a scan left its duplicates queued**

`stop_scan_handler` returned after cancelling the running job, so any queued scans stayed and the worker picked the next one up immediately. Stopping now cancels queued scans too.

**A dead lookup loop**

```python
existing_jobs = high_prio_queue.get_jobs()
for job in existing_jobs:
    if get_job_func_name(job) == "scan_platform" and job.is_started:
```

This could never match. RQ records the name as `endpoints.sockets.scan.scan_platforms`, not `scan_platform`, and a started job is not in the queue at all, so `is_started` is never true for anything `get_jobs()` returns. Verified against a live scan:

```
func_name: endpoints.sockets.scan.scan_platforms
is_started: True
in high_prio_queue.get_jobs()? False
```

The second loop worked, because it queried the workers and used the full path, so stopping a scan did function. But the two lookups had drifted apart, and the stale literal is what the first one drifted to. Both lookups now come from `_get_running_scan_job()` / `_get_queued_scan_jobs()`, which derive the name from `scan_platforms` itself so it cannot drift again, and the new guard uses the same helpers.

**A stale stop flag could abort the next scan**

`STOP_SCAN_FLAG` is only cleared inside `stop_scan()`, which runs when a scan observes it. If a scan ends between the flag being set and being checked, the flag survives and the next scan stops before doing any work. `scan_platforms` now clears it on entry, so a new scan never inherits an old stop request.

**Not addressed here**

The underlying reason a client cannot tell a scan is running: there is no way to query current scan state, only to receive pushed events. A reconnecting client stays blank until the next ROM happens to be identified, and a suspended mobile tab never reconnects at all. A status endpoint the client can call on reconnect would fix that properly, but it needs frontend work and is worth agreeing on separately. This PR makes the failure harmless rather than compounding.

**Checklist**

Nine new tests across `TestScanConcurrency` and `TestStopScan`: enqueues when idle, refuses when a scan is running, refuses when one is queued, ignores unrelated jobs so a cleanup or metadata task cannot block a scan, sets the stop flag for a running scan, cancels queued scans both with and without one running, and does not set a stop flag when there is nothing to stop.

`tests/endpoints/` + `tests/tasks/`: **836 passed, 0 failed**. black, isort, ruff and mypy clean on both changed files.

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written with AI assistance (Claude Code). The AI diagnosed the behavior against a live instance, wrote the patch and the tests, and ran them locally. Reviewed by me before submitting.

### DIFF
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -63,6 +63,38 @@ from utils.pegasus_exporter import PegasusExporter
 STOP_SCAN_FLAG: Final = "scan:stop"
 
 
+def _scan_job_func_name() -> str:
+    """Fully qualified name RQ records for an enqueued scan.
+
+    Derived from the function itself so it cannot drift out of sync with the
+    name RQ stores when the job is enqueued.
+    """
+    return f"{scan_platforms.__module__}.{scan_platforms.__name__}"
+
+
+def _get_running_scan_job() -> Job | None:
+    """The scan currently executing on a worker, if any.
+
+    A started job is no longer in the queue, so it can only be found by asking
+    the workers what they are holding.
+    """
+    for worker in Worker.all(connection=redis_client):
+        job = worker.get_current_job()
+        if job is not None and get_job_func_name(job) == _scan_job_func_name():
+            return job
+
+    return None
+
+
+def _get_queued_scan_jobs() -> list[Job]:
+    """Scans waiting in the queue, not yet picked up by a worker."""
+    return [
+        job
+        for job in high_prio_queue.get_jobs()
+        if get_job_func_name(job) == _scan_job_func_name()
+    ]
+
+
 @dataclass
 class ScanStats:
     total_platforms: int = 0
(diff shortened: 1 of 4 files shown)


gantoine (approved):

gantoine:
@greptileai

PR #3989, opened 2026-07-29
feat(screenscraper): pace and report scans against the account's reported limits

### LINKED ISSUE #3978: [Feature] Make scanning aware of ScreenScraper's reported thread and quota limits

ScreenScraper gives every account a scraping budget that depends on how much you have contributed to their database or donated to them. A brand new free account gets 1 simultaneous request ("thread") at 128 Kb/s. Contributors climb to 8 threads at 40 Mb/s, and a one-off 10 EUR donation adds 5 threads for life. On top of that there is a limit on how many requests you may make per minute (threads x 50) and per day (20,000 to 100,000, plus a separate, smaller allowance for ROMs that ScreenScraper does not recognise).

ScreenScraper reports all of these numbers back to RomM on every single response. RomM currently reads exactly one of them, `maxthreads`, and ignores the rest. That leads to two problems that pull in opposite directions:

1. **People who paid or contributed do not get what they paid for.** `SCAN_WORKERS` defaults to `1`, which means RomM only ever scans one ROM at a time, which means only one ScreenScraper request is ever in flight. A 5-thread donor account therefore scans at exactly the same speed as a free account. The extra threads sit unused unless you happen to know to raise `SCAN_WORKERS` yourself, and nothing in RomM tells you that.

2. **Raising `SCAN_WORKERS` can quietly cost you metadata.** RomM limits how many ScreenScraper requests run *at the same time*, but ScreenScraper also limits how many you make *per minute*, and RomM does not track that at all. Once you go over, ScreenScraper starts refusing requests. RomM waits 2 seconds and retries once, but if that retry is also refused the error is logged and the ROM is simply saved with no ScreenScraper data. The scan still reports success, so the only way to notice is spotting missing box art later.

There is a similar blind spot on the daily quotas. RomM has no idea it is at 19,000 of 20,000 requests until ScreenScraper cuts it off mid-scan. ScreenScraper actually keeps two counters, a total allowance and a smaller one for ROMs it does not recognise, and nothing in RomM surfaces either. Whether the second one matters depends entirely on your library: the two sit at exactly 10:1 in every tier, so the unrecognised-ROM counter only becomes the binding limit if more than about 10% of your lookups come back empty. ScreenScraper's database is broad enough that a tidy No-Intro or Redump collection will stay comfortably under that. A library heavy on ROM hacks, translations or homebrew can go well past it, and those are the users least likely to know the limit exists.

Use the rest of the account information ScreenScraper already sends us:

- **Respect the per-minute limit.** Pace ScreenScraper requests against the reported `maxrequestspermin` the same way we already pace IGDB, MobyGames and RetroAchievements. ScreenScraper is currently the only metadata provider with no rate limiting at all.
- **Show people their quota.** Surface remaining daily requests (and remaining unrecognised-ROM requests) in the scan progress output, so a scan that is about to hit the wall is visible before it does. The numbers are in every response already.
- **Warn when the configuration wastes the account.** If `SCAN_WORKERS` is lower than the account's thread allowance, log a one-line hint that scanning could be faster. If it is much higher, note that the extra workers will just queue.
- **Do not fail silently.** A repeated rate-limit refusal should be reported as a skipped ROM in the scan summary, not swallowed.
- **Apply the account's download speed to media too.** Cover art, screenshots and manuals are downloaded outside the ScreenScraper limiter entirely, so the reported `maxdownloadspeed` is neither used nor respected.

- **Just document it.** Adding a note to the docs explaining the relationship between `SCAN_WORKERS` and your ScreenScraper account would help, but it puts the burden on every user to look up their own thread count and do the maths, and it does nothing about the silent metadata loss.
- **Pick a safe hardcoded rate.** Simple, but it would either throttle contributors and donors unnecessarily or still overrun free accounts. The correct value is different for every account and ScreenScraper already tells us what it is.
- **Leave it reactive.** RomM does already handle the "quota exhausted" errors gracefully and stops hammering a dead quota, which is good. The gap is that everything is discovered by being refused rather than by reading the limits that were handed to us.

**Additional context**

Relevant code:

- `backend/adapters/services/screenscraper.py` - `_update_thread_allowance()` reads `maxthreads` and resizes the concurrency limiter. The retry block's error handling has no branch for a repeated rate-limit refusal, so it falls through to `log.error(...); return {}`.
- `backend/adapters/services/screenscraper_types.py` - `maxrequestspermin`, `maxrequestsperday` and `requeststoday` are already declared in the `SSUser` type but never read. `maxrequestskoperday`, `requestskotoday` and `maxdownloadspeed` are not declared at all.
- `backend/endpoints/sockets/scan.py` - `SCAN_WORKERS` is an `asyncio.Semaphore`, not worker processes. Worth noting that `env.template` describes it as "Number of worker processes for scanning tasks", which is misleading; the description in `docs/BACKEND_ARCHITECTURE.md` is accurate.
- `backend/handler/filesystem/resources_handler.py` - media downloads use their own session and bypass the ScreenScraper limiter.

On the per-minute maths: the limit is `threads x 50` per minute. If all slots are busy and the average response takes `t` seconds, we issue `60 x threads / t` requests per minute, which goes over budget whenever `t` is under 1.2 seconds, regardless of how many threads you have. ScreenScraper publishes its own average response times on their homepage: full game lookups sit around 1.5s (safe), but name searches are about 0.36s and lookups that come back empty about 0.7s. Name searches are used for platforms that cannot be matched by file hash, such as arcade and most computer systems, so the per-minute limit is easiest to trip on those platforms, and on any library with a high proportion of unmatched files, independently of how well the rest of the collection matches. This is worked out from ScreenScraper's published figures rather than measured against a live scan, so it is worth confirming with real timings before tuning anything.

The limits themselves are documented in ScreenScraper's FAQ (sections 02 and 03), which requires being logged in to read. The field list is public at https://api.screenscraper.fr/webapi2.php.

Analysis for this issue was AI-assisted.

Spinnich:
**Description**

Fixes #3978

ScreenScraper returns the account's allowances (`ssuser`) on **every** response, but RomM only ever read `maxthreads`. Everything else, the per-minute budget, the two daily quotas and the download speed, was discarded, so a scan could not tell how fast it was allowed to go or how close it was to the wall. Two concrete gaps followed:

1. **Fast lookups overran the per-minute budget without ever exceeding the thread cap.** The concurrency limiter bounds simultaneous requests, but a name search often returns in well under a second, so a single thread can comfortably issue far more than the allowed requests per minute. This adds a rate limiter alongside the existing thread cap.
2. **Media downloads bypassed the limiters entirely.** Covers, screenshots, manuals and mixed media are served from `screenscraper.fr` api2 endpoints and count against the same thread and per-minute allowances, but they went out through a plain `httpx.stream` with a fixed 120s timeout. On a multi-worker scan that alone could overrun the account.

The scan now also reads the limits *before* the first ROM (via `ssuserInfos.php`, which reports the same `ssuser` block and, measured, does not consume the daily quota) so pacing is correct from the first request instead of from the first response, and reports the remaining quota as it goes.

**On `maxrequestspermin`:** ScreenScraper's FAQ defines the budget, and that field, as `threads x 50`. The field does not follow it. Two accounts report `1024 x (threads + 1)`:

| threads | reported | documented (`threads x 50`) |
| ------- | -------- | --------------------------- |
| 9       | 10240    | 450                         |
| 6       | 7168     | 300                         |

At those rates the limiter would never engage. Every neighbouring field *does* reconcile exactly with the FAQ (both thread counts, both `maxdownloadspeed` values against the participation ladder plus the donation bonus), so the outlier is treated as unreliable and **the lower of the reported and documented budgets wins**. A stricter-than-documented account limit is still honoured.

**What a scan now does**

- Paces requests against the account's per-minute budget, on top of the existing thread cap.
- Logs the remaining daily and unrecognized-ROM quotas after each platform and at the end of the scan (repeats suppressed), and warns once when either drops below 10%.
- Hints once per scan when `SCAN_WORKERS` sits below the thread allowance (scanning slower than the account permits) or above it (the extra workers just queue).
- Raises a distinct `ScreenScraperRateLimitError` when a request is refused twice for exceeding the rate, so the ROM is listed as skipped in the scan summary rather than silently saved with no ScreenScraper data. Kept separate from the daily-quota 429 because this one clears on its own within the minute, so it must not trip the breaker that short-circuits the rest of the scan.
- Routes ScreenScraper media downloads through the same limiters and derives their timeout from the account's advertised download speed (120s floor, 600s cap), so a throttled free account gets room to finish a large manual while a stalled download still gives up.
- Says when the quotas actually reset (midnight French time) instead of "tomorrow".

**Files modified**

| File | What changed |
| ---- | ------------ |
| `backend/adapters/services/screenscraper.py` | The bulk of the work. New `SSAccountLimits` dataclass and `_update_account_limits()` replacing `_update_thread_allowance()`; module-level `RateLimiter` alongside the existing `ConcurrencyLimiter`; `prime_account_limits()` / `get_user_info()`; `reset_scan_state()`; `ScreenScraperRateLimitError`; `media_download_slot()` and `media_download_timeout()`; low-quota and `SCAN_WORKERS` advisories. `is_screenscraper_url()` moved here from `ss_handler` (unchanged logic) so the media path can share it. |
| `backend/adapters/services/screenscraper_types.py` | Three `SSUser` fields that were already on the wire but untyped: `maxrequestskoperday`, `requestskotoday`, `maxdownloadspeed`. |
| `backend/endpoints/sockets/scan.py` | Primes the account limits before the first ROM and resets the per-scan state (breaker, learned limits, skipped ROMs) at scan start; logs the quota per platform and the skipped-ROM summary at the end. All gated on `MetadataSource.SS` being in play. |
| `backend/handler/filesystem/resources_handler.py` | The four media download paths (cover, screenshot, manual, mixed media) now wrap their `httpx.stream` in `media_download_slot()` and take their timeout from it instead of a hardcoded `120`. Non-ScreenScraper URLs pass straight through at the same 120s as before. |
| `backend/handler/metadata/ss_handler.py` | Tracks ROMs skipped due to rate limiting (`note_rate_limited_rom` / `get_rate_limited_rom_names` / `reset_rate_limited_roms`) and catches `ScreenScraperRateLimitError` at the three lookup sites. `_is_screenscraper_host` removed in favour of the adapter's `is_screenscraper_url`. |
| `backend/utils/rate_limiter.py` | `RateLimiter` gains a `requests_per_second` property and `set_requests_per_second()`, so a limiter can be re-paced once the API reports its own per-account rate. Already-reserved slots keep their time. |
| `env.template` | `SCAN_WORKERS` comment now says what it actually bounds (ROMs in flight, and therefore concurrent metadata requests) rather than "worker processes". |
| `backend/tests/**` (5 files) | Coverage for all of the above, see below. |

**Testing notes**

- `uv run pytest` across the five affected test files — 336 passed. `trunk check` — clean across all 481 backend files.
- New tests cover: limit parsing (including junk/absent values), the `min(reported, documented)` pacing rule in both directions, priming (endpoint, applied limits, skipped without credentials, never fails the scan), the low-quota and `SCAN_WORKERS` advisories and their once-per-scan behaviour, the double-429 path raising `ScreenScraperRateLimitError` without tripping the daily breaker, per-scan state reset, the media slot (held, released on failure, paced, and left alone for other hosts), the speed-derived timeout at floor/middle/cap, each of the four `resources_handler` download sites, the skipped-ROM bookkeeping, and the scan-level quota/summary logging.
- Limits were read against live accounts to derive the `maxrequestspermin` finding; the `ssuserInfos.php` no-quota-cost claim was measured (two reads 90s apart moved `requeststoday` by 0).
- Backend only, no API schema change, so no `npm run generate` and no frontend impact.

**What a reviewer should look at**

1. **Scans will get slower on a free account, by design.** Media downloads now take a paced slot. At the default 1 thread / 50 requests per minute that is ~1.2s of spacing per download, and a ROM can pull a cover, screenshots and a manual. This is the account limit being respected rather than exceeded, but it is a visible behaviour change and worth agreeing on before merge.
2. **`resources_handler.py` now imports from `adapters/services/screenscraper`.** A filesystem handler taking a dependency on a provider adapter is a bit against the grain here. The alternative is passing the slot or timeout down from the caller; happy to restructure if you prefer that.
3. **The media slot is held through post-download processing.** `_discard_if_chroma_key()` and the synchronous `convert_to_webp()` run inside the `async with`, so the account's thread slot is held a little longer than the transfer needs. The sync call blocks the loop regardless so it costs nothing extra in practice, but narrowing the block to just the streaming would be tidier.
4. **`_is_daily_quota_error()` has a now-stale docstring** (`ss_handler.py`). It says "ScreenScraper only raises 429 when its daily quota is exhausted", which this PR makes untrue, since `ScreenScraperRateLimitError` is also a 429. Every current call site catches the subclass first so there is no live bug, but a future `except HTTPException` site would misread a transient rate-limit as permanent quota exhaustion. Left as-is deliberately rather than expanding scope; say the word and I will either fix the docstring or make the check exclude the subclass explicitly.
5. **`reset_scan_state()` clears `_account_limits` but not the limiters.** They keep the previous scan's learned pacing. Normally harmless because priming re-reads them at the next scan start, but if priming fails the scan paces at the previous account's rate while `media_download_timeout()` correctly falls back to the default: a small asymmetry, flagging it in case you want them reset together.
6. **Module-level mutable state.** `_account_limits`, the limiters and the skipped-ROM set are module globals, matching the existing `_daily_quota_exhausted` pattern. This is the right scope because `SCAN_WORKERS` is an `asyncio.Semaphore` inside a single worker process, not separate processes, but it is worth a second pair of eyes.

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written with substantial AI assistance (Claude). The AI wrote the implementation and tests; I directed the design, ran the live-account measurements that produced the `maxrequestspermin` and `ssuserInfos.php` findings, and reviewed and verified the result.

**Checklist**

N/A, backend only. Scan log output changes as described above.


Spinnich:
Thanks — all four were valid in some form. Addressed in c2786429f, and the branch is now merged up with `master`.

**1. Concurrent scans corrupt shared state (P1)**

Fixed. The reset block is now gated on `MetadataSource.SS` being in the scan's sources, matching how priming and the quota logging were already gated, so a scan that never touches ScreenScraper leaves the shared state alone.

Worth noting for anyone reading this later: #3974 landed concurrent-scan refusal while this PR was open, but it deliberately skips `DEV_MODE` (`if not DEV_MODE and (_get_running_scan_job() or _get_queued_scan_jobs())`), which is exactly the overlap window called out here. So the finding survives that fix and the gate is still needed.

Full isolation between overlapping scans would mean threading a per-scan context through `ss_handler` and the service. I've left that out of scope: the limiters *should* stay process-global (they model a per-account cap, so concurrent scans in one process must share them), and the pre-existing `_daily_quota_exhausted` breaker already had the same property before this PR.

**2. Media slots cover local processing (P2)**

Fixed for the case where it was real. Of the four download paths, only `_store_cover` did post-transfer work inside the slot; `_store_screenshot` and `_store_manual` have none, and `store_media_file` already ran its chroma-key check outside. `_store_cover` now hands the slot back before `_discard_if_chroma_key()` and `convert_to_webp()`.

One correction on the description: the write can't be moved out. It's interleaved with the network read (`async for chunk in response.aiter_raw()`), so releasing the slot before it would mean buffering the whole file in memory. Content-type validation is header-only and free. The inspect-and-re-encode step was the part actually worth moving, and that's what changed.

**3. Scan reset preserves stale limiters (P2)**

Fixed. `reset_scan_state()` now returns both limiters to their defaults alongside `_account_limits`.

Defaults rather than "keep what we learned" is the deliberate choice: priming re-reads the account moments later, and if that fails, the conservative default is safe for whatever account the credentials now name, whereas the previous scan's allowance may be more than this one is entitled to. It also makes pacing consistent with `media_download_timeout()`, which already fell back to the default on reset.

**4. Rate-limited ROM retries another lookup (P2)**

Fixed, and this was the sharpest of the four. Confirmed the mechanism in `scan_handler.py`: `lookup_rom` returned `(SSRom(ss_id=None), False)`, so `fetch_ss_rom` fell through to `get_rom()` and spent another request plus its 2s retry backoff against a budget that was already gone.

Rather than special-casing the fallthrough, `ScreenScraperRateLimitError` now propagates out of the three handler call sites and is caught once in `fetch_ss_rom`, which records the skip and returns an empty match. One refusal ends ScreenScraper work for that ROM.

This required making `_is_daily_quota_error()` exclude the subclass explicitly, since both are 429s. Previously that only worked because every call site happened to catch `ScreenScraperRateLimitError` first — correct, but fragile, and it would have silently misclassified a transient rate limit as permanent quota exhaustion at any future `except HTTPException` site.

**Tests**

The three tests that asserted the old swallow-in-handler behavior became propagation tests, plus new coverage for: the daily-quota vs rate-limit distinction, the pacing reset, the non-ScreenScraper gating, and a `scan_rom` test asserting the name-search fallback is not attempted after a refusal.

1499 passed / 2 skipped locally, `trunk check` clean.

Spinnich:
@gantoine Okay, I believe this is in a good spot now. Probably best to hold this until a v5.2 alpha (which might be your plan already)! That's assuming we merge it into master at all obviously.

I also have a performance PR based on this branch that I plan to submit after the fact. They're sort of two sides of the same coin but I broke them apart to (hopefully) make reviewing more manageable.

gantoine (approved):

PR #3995, opened 2026-07-30
fix(tasks): honor ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES

### LINKED ISSUE #3994: [Bug] 5.1.0 ignores ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES and runs orphaned resource cleanup nightly on every install
**RomM version**

5.1.0

**Describe the bug**

`ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` no longer does anything in 5.1.0. The
"Cleanup orphaned resources" task schedules itself and runs every night at 05:00 on every
install, including installs that left the setting at its default of `false`.

Some history, because this is a two step regression:

1. Through 5.0.x and up to 5.1.0-beta.2, "Cleanup orphaned resources" was a **manual only**
   task. It lived in `backend/tasks/manual/` and only ran when an admin pressed the button.
2. #3970 (in 5.1.0-beta.3) made it schedulable, gated behind the new
   `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` flag, which defaults to `false`. "Schedule
   off" was implemented by giving the task no cron expression.
3. That broke the Administration -> Tasks page: an empty cron string made the frontend's
   `cronstrue` call throw, which blanked the entire Tasks panel. Commit `06cafd4b1` (the
   commit 5.1.0 is tagged from) fixed the blank page by always passing the cron expression.

The blank page fix works, but it removed the only thing the flag was wired to. The flag is now
an unused import in `backend/tasks/scheduled/cleanup_orphaned_resources.py`, and since the task
is also hardcoded `enabled=True` (so it stays manually runnable), nothing is left to stop it
from being scheduled. The `if not self.cron_string: self.unschedule()` guard in `init()` is now
unreachable, because `SCHEDULED_CLEANUP_ORPHANED_RESOURCES_CRON` defaults to `0 5 * * *` and
`_get_env` falls back to that default even for an explicitly empty value.

The same commit also left the test suite red, see below.

**Why this matters**

The task deletes directories under `RESOURCES_BASE_PATH/roms/` with `shutil.rmtree` whenever the
matching platform or ROM is not in the database. On a healthy install that is harmless, it clears
genuinely orphaned artwork. The problem is the case where the database and the filesystem
temporarily disagree, for example a library share that did not mount, so a scan dropped a
platform. In that window the nightly run deletes that platform's whole resource tree, custom
uploaded covers and screenshots included, with no undo.

#3971 added a guard for the total wipe case (database reports zero platforms while artwork exists
on disk), which is good. A partial mismatch, one missing platform out of many, is not guarded.
That is presumably why the schedule was made opt in to begin with, and it is now on for everyone.

Separately from the blast radius: a documented setting that silently does nothing is worth fixing
on its own.

**To Reproduce**

1. Run 5.1.0 with `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` unset or `false` (the default in
   `env.template`).
2. Start RomM and log in as an admin.
3. Go to **Settings -> Administration -> Tasks**.
4. Under **Scheduled**, "Cleanup orphaned resources" is listed as enabled with a schedule of
   "at 05:00 AM, every day". With the flag off it should not appear as scheduled at all.
5. Leave it running past 05:00 and the task executes, visible in the task history and in the
   logs as `Starting Cleanup orphaned resources task...`.

**Expected behavior**

With `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES=false`, the task is not scheduled and never
runs on its own. It stays available as a manual run from the Tasks page, which is how it behaved
in every release before 5.1.0. Setting the flag to `true` is what opts you into the nightly run.

**Additional context**

**The test suite is red on the 5.1.0 release commit.** The CI run for `06cafd4b1` fails on both
`pytest (mariadb)` and `pytest (postgresql)`, with `2 failed, 2511 passed, 2 skipped`:

```
FAILED tests/tasks/test_cleanup_orphaned_resources.py::TestCleanupOrphanedResourcesTask::test_no_cron_string_by_default
    AssertionError: assert '0 5 * * *' is None

FAILED tests/tasks/test_cleanup_orphaned_resources.py::TestCleanupOrphanedResourcesTask::test_init_unschedules_when_no_cron
    AssertionError: assert Job('fb83a3bd-db9a-4e40-a6ae-148023f5370a', enqueued_at=None) is None
```

Those two tests are the ones #3970 added to pin exactly this behavior, so they are not stale
noise, they are correctly reporting the regression. The second failure is also the cleanest proof
of it: `init()` returned a real RQ `Job`, meaning the task scheduled itself inside a test run
where the flag is off.

**Suggested fix.** Restore the gate without reintroducing the empty cron string that broke the
Tasks page, for example keep `cron_string` populated at all times and gate the scheduling
decision instead:

```python
def init(self) -> Job | None:
    # `enabled` stays True so the task remains manually runnable; the env flag
    # is what decides whether it also runs on a schedule.
    if not ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES:
        self.unschedule()
        return None

    return super().init()
```

The two failing tests would then need updating, since `cron_string is None` is no longer the
signal for "schedule off".

**One loose end from the original blank page bug.** Neither side of the empty cron string path was
hardened, so the same failure is still reachable for any future scheduled task that ships without
a cron expression:

- `_build_task_info` (`backend/endpoints/tasks.py:148`) coerces a missing cron to `""` with
  `cron_string=task.cron_string or ""`, rather than leaving it null.
- `convertCronExperssion` (`frontend/src/utils/index.ts:102`) calls `cronstrue.toString()` with no
  guard, and cronstrue throws `cron expression is empty` on `""`.
- Both UIs call it from inside a Vue computed that the template reads, v1 at
  `frontend/src/components/Settings/Administration/Tasks.vue:28` and v2 at
  `frontend/src/v2/components/Settings/TasksSection.vue:41`. When the computed throws during
  render, Vue substitutes an empty comment node for the whole component, so one unschedulable task
  hides every task on the page. In a production build the error is only visible in the browser
  console.

Making `convertCronExperssion` return an empty string or a "not scheduled" label instead of
throwing would turn that class of bug into a cosmetic gap rather than a blank page.

**Minor.** `06cafd4b1`'s commit message reads "fix cron_string in update_switch_titledb", but the
change is in `cleanup_orphaned_resources.py`. Worth noting for anyone bisecting later.

---

*AI assistance disclosure: this report was investigated and written with Claude. The findings were
verified by running the 5.1.0 task code against the shipped config and by reading the upstream CI
logs for the release commit, not just by reading the diff.*

Spinnich:
**Description**

Fixes #3994

`ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES` had no effect in 5.1.0. The
"Cleanup orphaned resources" task was scheduled at 05:00 on every install
regardless of the setting. Because the task `shutil.rmtree`s resource
directories for platforms and ROMs the database does not report, a library
share that is temporarily unmounted or a database that is mid-migration could
cost users their custom covers and screenshots, with no undo.

The flag was wired up by expressing "schedule off" as "no cron string". A
later fix for the resulting blank Tasks page made the task always pass its
cron expression, which left the `if not self.cron_string` guard in `init()`
unreachable and silently removed the flag's only effect.

The underlying problem is that `enabled` carried two meanings: "the feature is
available and can be run by hand" (`can_run_manually = manual_run and
enabled`) and "the schedule is on" (what `PeriodicTask.init()` read). This
task needs those to differ, since it should stay manually runnable with its
schedule off.

This splits them. A new `Task.is_scheduled` property defaults to
`enabled and bool(cron_string)`, and `init()` now drives off it.
`CleanupOrphanedResourcesTask` overrides it to return the flag, so the cron
expression stays populated and the flag alone decides whether the task also
runs on its own. Reading the flag inside the property rather than at
construction keeps it re-evaluated on every `init()`.

The API side follows: `_build_task_info` reports a cron expression only when
the task is actually scheduled with it, so the Tasks page no longer shows a
schedule that will never fire.

**Files modified**

Backend

- `backend/tasks/tasks.py` — adds the `Task.is_scheduled` property and makes
  `PeriodicTask.init()` schedule/unschedule from it instead of from `enabled`.
  Side benefit: an enabled task with no cron expression now has a stale job
  cleaned up instead of left in the scheduler.
- `backend/tasks/scheduled/cleanup_orphaned_resources.py` — replaces the
  unreachable `init()` override with an `is_scheduled` override returning
  `ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES`. `enabled` and `manual_run`
  stay `True` so admins can still run it on demand.
- `backend/endpoints/tasks.py` — `_build_task_info` reports `cron_string` only
  when the task is scheduled.

Frontend

- `frontend/src/utils/index.ts` — `convertCronExperssion` returns `""` instead
  of throwing on an empty or malformed expression. Callers use it inside
  computed properties that templates read, and a throw there costs the whole
  surrounding component (Vue substitutes a comment node, blanking the panel).
- `frontend/src/v2/components/Settings/TasksSection.vue` — the clock icon is
  derived from the schedule rather than from `enabled`, so a task that is
  runnable but unscheduled shows `mdi-clock-remove-outline`.

Tests

- `backend/tests/tasks/test_tasks.py` — `is_scheduled` truth table and the new
  unschedule-on-missing-cron path.
- `backend/tests/tasks/test_cleanup_orphaned_resources.py` — replaces the two
  tests that asserted the "no cron string" contract with coverage of the flag:
  not scheduled by default, scheduled when the flag is on, `init()` neither
  schedules nor leaves a stale job when it is off, and the cron expression
  stays populated either way.
- `backend/tests/endpoints/test_tasks.py` — `/api/tasks` reports an empty
  `cron_string` for an unscheduled task while keeping `manual_run` true.
- `frontend/src/utils/index.test.ts` — empty and unparseable expressions
  describe as `""`.
- `frontend/src/v2/components/Settings/TasksSection.test.ts` (new) — renders
  the section at the layer the blank-panel bug lived at: an unscheduled task
  still renders with its run button and an off clock, and a scheduled one
  still shows its description and on clock.

No migration, no schema change, no new setting. `env.template` and
`docs/BACKEND_ARCHITECTURE.md` already document the flag as the gate, so they
needed no edit.

**Testing notes**

- Backend suite: `2519 passed, 2 skipped`. For reference, the 5.1.0 release
  commit was `2 failed, 2511 passed, 2 skipped`, the two failures being the
  tests referenced in the issue.
- Frontend: `594 passed` across 43 files, `npm run typecheck` clean for
  `src/`, `npm run build` succeeds, i18n parity check clean.
- `trunk fmt && trunk check` clean on every changed file.
- Every new test was confirmed to fail against the unfixed code before the fix
  was written.
- Verified in-process against a real RQ scheduler in both directions: with the
  flag off `init()` schedules no job and `/api/tasks` reports `cron_string:
  ""`; with it on a real cron job is created and the endpoint reports
  `"0 5 * * *"`; flipping it back off removes the existing job, which is the
  upgrade path for installs that already have the 5.1.0 job in Redis.
- `/api/tasks` checked live against a running backend, not only through the
  test client.

**Things worth a reviewer's attention**

1. *Disabled scheduled tasks also stop advertising a cron.* Because
   `_build_task_info` now gates on `is_scheduled`, a scheduled task that is
   disabled reports `cron_string: ""` where it previously reported its
   inactive expression. For example `scan_library` with
   `ENABLE_SCHEDULED_RESCAN=false` reported `"0 3 * * *"` before and reports
   `""` now. This is consistent with only reporting a schedule that fires, and
   it touches the API output of five other tasks, so it is worth a deliberate
   yes or no. Reverting just that line keeps the core fix intact.
2. *`schedule()` still checks `enabled`, not `is_scheduled`.* Left alone on
   purpose: `init()` is its only production caller, and switching it would
   turn the documented "return `None` when there is no cron expression"
   behaviour into a raise. Policy lives in `init()`, the primitive stays dumb.
3. *No runtime re-check inside `run()`.* Sibling tasks such as
   `cleanup_zip_cache` and `scan_library` re-read their flag in `run()` and
   self-unschedule. Copying that here would break manual runs, which is the
   entire point of this task, so it is intentionally absent. Stale jobs are
   handled at startup by `init()` instead.
4. *The cron helper fix lives in shared `src/utils/`.* That means v1's
   `Tasks.vue` is hardened too without editing a frozen v1 file, so
   `check-v1-frozen` stays green.

**Checklist**

**AI assistance disclosure**

This change was written with AI assistance (Claude Code), used extensively:
for investigating the regression, designing the `is_scheduled` split, writing
the implementation and all tests, and running the verification described
above. Every change was reviewed and tested by me before submitting, including
the empirical scheduler and endpoint checks in both flag states.

### DIFF
--- a/backend/endpoints/tasks.py
+++ b/backend/endpoints/tasks.py
@@ -138,6 +138,10 @@ manual_tasks: list[ManualTask] = [
 
 def _build_task_info(name: str, task: Task) -> TaskInfo:
     """Builds a TaskInfo object from task details."""
+    # A task holding a cron expression it never runs on reads as scheduled in
+    # the UI, so only report the expression the task is actually scheduled with.
+    cron_string = task.cron_string if task.is_scheduled else None
+
     return TaskInfo(
         name=name,
         type=task.task_type,
@@ -145,7 +149,7 @@ def _build_task_info(name: str, task: Task) -> TaskInfo:
         description=task.description,
         enabled=task.enabled,
         manual_run=task.can_run_manually,
-        cron_string=task.cron_string or "",
+        cron_string=cron_string or "",
     )
 
 
--- a/backend/tasks/scheduled/cleanup_orphaned_resources.py
+++ b/backend/tasks/scheduled/cleanup_orphaned_resources.py
@@ -4,7 +4,6 @@ import shutil
 from dataclasses import dataclass
 
 from anyio import Path as AnyioPath
-from rq.job import Job
 
 from config import (
     ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES,
@@ -86,14 +85,12 @@ class CleanupOrphanedResourcesTask(PeriodicTask):
             func="tasks.scheduled.cleanup_orphaned_resources.cleanup_orphaned_resources_task.run",
         )
 
-    def init(self) -> Job | None:
-        # `enabled` stays True for manual runs, so the absence of a cron string
-        # is what drives unscheduling when the schedule is turned off.
-        if not self.cron_string:
-            self.unschedule()
-            return None
-
-        return super().init()
+    @property
+    def is_scheduled(self) -> bool:
+        # `enabled` stays True so an admin can always run this by hand, and the
+        # cron string stays populated either way. This flag alone decides
+        # whether the task also runs on its own.
+        return ENABLE_SCHEDULED_CLEANUP_ORPHANED_RESOURCES
 
     @initialize_context()
     async def run(self, force: bool = False) -> dict[str, int]:
--- a/backend/tasks/tasks.py
+++ b/backend/tasks/tasks.py
@@ -79,6 +79,15 @@ class Task(ABC):
         """Whether an admin can trigger this task on demand."""
         return self.manual_run and self.enabled
 
+    @property
+    def is_scheduled(self) -> bool:
+        """Whether the task runs itself on a schedule.
+
+        Separate from `enabled` so a task can stay manually runnable while its
+        schedule is off. Subclasses override this to gate on their own setting.
+        """
+        return self.enabled and bool(self.cron_string)
+
     @abstractmethod
     async def run(self, *args: Any, **kwargs: Any) -> Any: ...
 
@@ -106,9 +115,9 @@ class PeriodicTask(Task, ABC):
         """
         job = self._get_existing_job()
 
-        if self.enabled and not job:
+        if self.is_scheduled and not job:
             return self.schedule()
-        elif job and not self.enabled:
+        elif job and not self.is_scheduled:
             self.unschedule()
             return None
         return None
--- a/backend/tests/endpoints/test_tasks.py
+++ b/backend/tests/endpoints/test_tasks.py
@@ -165,6 +165,45 @@ class TestListTasks:
         assert watcher_task["cron_string"] == ""
 
     @patch("endpoints.tasks.ENABLE_RESCAN_ON_FILESYSTEM_CHANGE", False)
+    @patch("endpoints.tasks.manual_tasks", [])
+    @patch(
+        "endpoints.tasks.scheduled_tasks",
+        [
+            {
+                "name": "test_unscheduled",
+                "type": TaskType.CLEANUP,
+                "task": Mock(
+                    spec=Task,
+                    task_type=TaskType.CLEANUP,
+                    title="Unscheduled Task",
+                    description="Manually runnable, but not on a schedule",
+                    enabled=True,
+                    manual_run=True,
+                    can_run_manually=True,
+                    is_scheduled=False,
+                    timeout=300,
+                    cron_string="0 5 * * *",
+                ),
+            }
+        ],
+    )
+    def test_list_tasks_omits_cron_for_unscheduled_task(self, client, access_token):
+        """A task that carries a cron expression it never runs on reports none.
+
+        Reporting the expression anyway makes the Tasks page show a schedule
+        that will not fire.
+        """
+        response = client.get(
+            "/api/tasks", headers={"Authorization": f"Bearer {access_token}"}
+        )
+
+        assert response.status_code == status.HTTP_200_OK
+        task = response.json()["scheduled"][0]
+
+        assert task["cron_string"] == ""
+        assert task["manual_run"] is True
+
+    @patch("endpoints.tasks.ENABLE_RESCAN_ON_FILESYSTEM_CHANGE", False)
     @patch("endpoints.tasks.RESCAN_ON_FILESYSTEM_CHANGE_DELAY", 10)
     @patch("endpoints.tasks.manual_tasks", [])
     @patch("endpoints.tasks.scheduled_tasks", [])
--- a/backend/tests/tasks/test_cleanup_orphaned_resources.py
+++ b/backend/tests/tasks/test_cleanup_orphaned_resources.py
@@ -1,8 +1,9 @@
 import os
 from types import SimpleNamespace
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
+from rq.job import Job
 
 import tasks.scheduled.cleanup_orphaned_resources as mod
 from tasks.scheduled.cleanup_orphaned_resources import CleanupOrphanedResourcesTask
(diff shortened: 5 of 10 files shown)


gantoine:
fixed this in another PR

PR #4017, opened 2026-07-30
perf(scan): stop hashing single-file ROMs twice

XenuIsWatching:
**Description**

`_calculate_rom_hashes` streams every chunk through **two** independent sets of hashers:

```python
def update_hashes(chunk: bytes | bytearray):
    md5_h.update(chunk)
    rom_md5_h.update(chunk)

    sha1_h.update(chunk)
    rom_sha1_h.update(chunk)

    nonlocal crc_c
    crc_c = binascii.crc32(chunk, crc_c)
    nonlocal rom_crc_c
    rom_crc_c = binascii.crc32(chunk, rom_crc_c)
```

The second set accumulates the ROM-level hash. That is genuinely required for a multi-file ROM, whose hash spans the concatenation of its top-level files and therefore cannot be derived from the per-file digests. But two callers pay for it without needing it:

- **Single-file ROMs** (the `elif hashable_platform:` branch). The ROM spans one file, so both sets see identical bytes and produce identical digests. The ROM-level pass is pure duplicate work.
- **Nested files inside a multi-file ROM.** These were handed throwaway accumulators whose results were then discarded at the call site with `_`:

```python
crc_c, _, md5_h, _, sha1_h, _ = await asyncio.to_thread(
    self._calculate_rom_hashes,
    Path(f_path, file_name),
    0,
    hashlib.md5(usedforsecurity=False),   # hashed into...
    hashlib.sha1(usedforsecurity=False),  # ...then thrown away
)
```

Since hashing is CPU-bound, a second pass over a chunk costs about as much as the first, so this roughly doubles hashing cost for the common case. Most libraries are predominantly single-file, so this is the default path rather than an edge case.

**Changes**

- The ROM-level accumulators on `_calculate_rom_hashes` become optional. When they are not supplied, only the per-file hashers run.
- Single-file ROMs reuse the file's own hashers as the ROM-level values instead of recomputing them.
- Nested files in multi-file ROMs stop requesting accumulators they discard.
- Top-level files in multi-file ROMs are unchanged: they still stream through the second set, because theirs is the case that needs it.

The signature stays backward compatible (the accumulators just gained defaults), so existing callers and tests are unaffected.

**Measurements**

Raspberry Pi 4, a 134 MB Nintendo 3DS ROM, page cache warmed first, best of three. Compared on **CPU time** rather than wall clock, because the machine was concurrently running a library scan:

| | CPU time | throughput |
| --- | --- | --- |
| before | 3.68s | 36.5 MB/s |
| after | 2.23s | **60.3 MB/s** |

**1.65x** on a real ROM (1.42x wall clock under that load). An isolated in-memory benchmark of just the hasher calls shows 2.4x; the difference is the `read()` syscalls and page-cache copies that the optimization cannot remove.

Digests are byte-identical before and after, at both the file and ROM level:

```
crc      7603056f       rom_crc  7603056f (before) == 7603056f (after)
md5      083c286178ed0ab28c941596429b86f3
sha1     51d121e366abe00ff3f90d8a1e7ad43e9f721d6a
```

**Testing**

Two tests added to `tests/handler/filesystem/test_roms_handler.py`, pinning the invariants on each side of the change:

- `test_get_rom_files_single_rom_hashes_match_its_only_file` — a single-file ROM's ROM-level hashes equal its one file's hashes. This is the identity that makes the reuse valid.
- `test_get_rom_files_multi_rom_hash_spans_every_part` — a multi-file ROM's hash still equals the hash of its top-level files concatenated, and is not merely one part's hash.

The second was mutation-tested: dropping the accumulators from the multi-file top-level branch (the change this PR deliberately does *not* make) makes it fail, so it is not vacuous.

Also verified:

- `tests/handler/filesystem/test_roms_handler.py`: 76 passed, including the pre-existing `test_get_rom_files_single_rom`, which already asserted exact ROM-level CRC/MD5/SHA1 for a single-file ROM and continues to pass unchanged.
- `ruff 0.15.22` and `black 26.5.1` (the versions pinned in `.trunk/trunk.yaml`) clean on both changed files.
- `mypy` reports zero errors in `handler/filesystem/roms_handler.py`. It surfaces pre-existing errors in transitively imported modules such as `handler/scan_handler.py`, none of them touched here.

<sub>Full disclosure on what I could not complete: a full `tests/handler tests/endpoints` run was killed by my own timeout while the host was saturated (load ~25 on 4 cores) by an unrelated library scan running concurrently. An earlier partial run of that same selection gave 355 passed / 3 failed, and those 3 (`TestScanTotals`) fail identically on unmodified `master` because they require a Redis on `127.0.0.1:6379`. I have no reason to think anything outside the hashing path is affected, but I would rather flag the gap than imply coverage I did not get.</sub>

**Checklist**

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this change was written with AI assistance (Claude Code). The bottleneck was found by profiling a real scan on my own hardware; the AI wrote the patch, the tests and the benchmarks, and I reviewed the result.

### DIFF
--- a/backend/handler/filesystem/roms_handler.py
+++ b/backend/handler/filesystem/roms_handler.py
@@ -414,9 +414,6 @@ class FSRomsHandler(FSHandler):
                             crc_c, _, md5_h, _, sha1_h, _ = await asyncio.to_thread(
                                 self._calculate_rom_hashes,
                                 Path(f_path, file_name),
-                                0,
-                                hashlib.md5(usedforsecurity=False),
-                                hashlib.sha1(usedforsecurity=False),
                             )
                     except zlib.error:
                         crc_c = 0
@@ -545,19 +542,18 @@ class FSRomsHandler(FSHandler):
                 )
         elif hashable_platform:
             try:
-                crc_c, rom_crc_c, md5_h, rom_md5_h, sha1_h, rom_sha1_h = (
-                    await asyncio.to_thread(
-                        self._calculate_rom_hashes,
-                        Path(abs_fs_path, rom.fs_name),
-                        rom_crc_c,
-                        rom_md5_h,
-                        rom_sha1_h,
-                    )
+                crc_c, _, md5_h, _, sha1_h, _ = await asyncio.to_thread(
+                    self._calculate_rom_hashes,
+                    Path(abs_fs_path, rom.fs_name),
                 )
             except zlib.error:
                 crc_c = 0
                 md5_h = hashlib.md5(usedforsecurity=False)
                 sha1_h = hashlib.sha1(usedforsecurity=False)
+
+            # A single-file ROM spans exactly one file, so its ROM-level hashes
+            # are that file's hashes.
+            rom_crc_c, rom_md5_h, rom_sha1_h = crc_c, md5_h, sha1_h
 
             # Calculate the RA hash if the platform has a slug that matches a known RA slug
             if calculate_hashes:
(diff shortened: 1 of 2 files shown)


Spinnich:
Nice find! I bet this will help speed up scanning a fair bit.

gantoine (approved):
👍🏼

gantoine:
@greptileai

gantoine:
@XenuIsWatching thanks for this! for future PRs can you setup commit signing? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

XenuIsWatching:
> @XenuIsWatching thanks for this! for future PRs can you setup commit signing? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

uhh... i can, but i'm rather a surprised for a project like this (more of a hobby kind of deal) it's requested. Unless if I've been under a rock for a while and now more people (or rather agents) can spoof who I am a lot more nowadays. I don't wanna sound like I'm pushing back here, as I can do that for sure, just rather new to using GPG keys (as i've been never been asked/forced to use them) despite all the rather critical software i've worked on in the past

gantoine:
yeah that's reasonable, it's just becoming good practice to sign your commits with your SSH (or a GPG) key so we can attribute commits to the same user/system.

Issue #4002, opened 2026-07-30
[Bug] ScreenScraper "Update metadata" scan only fills in missing data, it never refreshes stale data


Spinnich:
**RomM version**

5.1.0 (also present on `master`)

**Describe the bug**

The **Update metadata** scan says it "Updates the metadata for games that have been matched with selected metadata sources using the external ID". With ScreenScraper selected, it does not do that. It only fills in blanks. Anything already stored on a game is kept, even when ScreenScraper now has corrected or better data for it.

What actually gets refreshed is the hidden ScreenScraper data blob (genres, release date, player count, rating, media links). Everything you actually look at does not:

- **Game title and description**: kept as-is if already set.
- **Box art, manual, screenshots**: only re-downloaded when the stored image URL is different from the new one. On an Update scan the old URL is deliberately kept, so it can never be different. On top of that, ScreenScraper's media links are fixed endpoints (`mediaJeu.php?...&media=box-2D`), so the link stays byte-identical even after somebody uploads brand new art. So there are two independent reasons that check can never fire for ScreenScraper.
- **All the extra ScreenScraper media** (bezel, logo, marquee, fanart, mix image, 3D box, physical media, title screen, video): the download step exits early the moment it sees a file already on disk. There is no comparison whatsoever.
- **The manual re-match dialog** has the same limitation: it only re-pulls ScreenScraper media when the ScreenScraper ID changes, so re-matching a game to the same entry refreshes nothing.

A side effect of this split: the database can end up storing a freshly fetched media URL while the file on disk is still the old image, because the URL is updated but the download is skipped.

The result is that the only way to pick up corrected ScreenScraper data is a **Complete rescan**, which is a far bigger hammer. It throws away every provider ID and re-matches every game from scratch, so hand-corrected matches are lost, and it re-downloads everything for the whole platform, which is expensive against the ScreenScraper request quota. "My box art is out of date" should not require that.

This matters more for ScreenScraper than for most providers because ScreenScraper is community-curated: art and game data are corrected and re-uploaded there constantly, and RomM's per-game media feature (logos, marquees, mix images, bezels) is built almost entirely on it. A user who fixes an entry upstream currently has no reasonable way to pull the fix in.

[...]

PR #4028, opened 2026-07-31
fix(scan): refresh stale ScreenScraper data on update metadata scans

Spinnich:
**Description**

Part of #4002.

Rescoped at review request: this now carries only the ScreenScraper media hash
comparison. The `user_edited_fields` tracking is gone, along with its column and
migration, and will be superseded by explicit metadata field locking.

ScreenScraper serves every media from a fixed endpoint, so its URLs stay
byte-identical when the artwork behind them is replaced. The cover and screenshot
downloads are gated on a URL diff, which for ScreenScraper can therefore never
fire, and the extra media step exits early whenever a file is already on disk. So
an update scan never refreshes ScreenScraper artwork, no matter how stale it is.

`jeuInfos` returns a `crc`/`md5`/`sha1` alongside every media entry precisely so a
client can tell whether the online file still matches its local copy. Those hashes
are now recorded in `ss_metadata`, and a resource is refetched when the reported
hash differs from the one the previous scan recorded, falling back to hashing the
file on disk so libraries scanned before hashes were recorded don't redownload
what is already current. The cover is hashed at `cover/big.png`, the copy that
still holds the bytes as downloaded once WebP conversion has run.

ScreenScraper also reports a byte size per media entry. That is checked first,
because the recorded hash is written to the database before the download runs, so
a scan killed mid-stream leaves a short file described by a hash that still
matches. A `stat()` catches that without reading the file.

Re-matching a ROM to the same ScreenScraper game now also fills in media that
never made it to disk, rather than only running when `ss_id` changes.

**What this does not do**

- `name` and `summary` are still pinned to their stored values on an update scan.
  Reversing that needs a way to tell a provider-written value from a hand-edited
  one, which is what field locking will provide. #4002 stays open for it.
- Manuals are deliberately left out of the hash check. `POST /roms/{id}/manuals`
  writes `path_manual` but leaves any scraped `url_manual` in place, and both
  uploaded and downloaded manuals land at the same path, so nothing distinguishes
  them. On master that is harmless, since the URL-diff gate never fires for
  ScreenScraper. A hash check would make it live and overwrite an uploaded manual
  with the provider's copy. That ambiguity is a pre-existing gap and gets its own
  PR; manuals pick up the hash check once a scan can tell the two apart.
- `manual_md5` and `manual_size` are still recorded in `ss_metadata` alongside the
  other media, and are simply unread until that follow-up lands.

**Files modified**

| File | Change |
| --- | --- |
| `backend/handler/metadata/ss_handler.py` | Carries an `*_md5` and `*_size` through for each of the 18 media slots, keyed on the same stripped URL that lands in `*_url`. Adds `ss_media_is_stale()` and `ss_resource_needs_refresh()`. |
| `backend/handler/filesystem/resources_handler.py` | `file_md5()`, `file_matches_md5()` and `file_has_size()` (all path-validated), `get_downloaded_cover_path()`, and an `expected_md5` argument on `store_media_file()`. |
| `backend/endpoints/sockets/scan.py` | The cover and screenshot overwrite gates now also consult the ScreenScraper hash; extra media downloads carry their expected md5. |
| `backend/endpoints/roms/__init__.py` | The ScreenScraper media block also runs on a re-match to the same game, to fill in media that never made it to disk. |
| `backend/tests/**` | New coverage for md5 and size extraction, the staleness rule, the `store_media_file` gate, and the scan wiring. |

No schema change, no migration, and no response schema touched, so no OpenAPI
regeneration.

**Testing notes**

- `uv run pytest` — 2696 passed, 2 skipped.
- `trunk check` clean on every changed file.

Three update scans against a live ScreenScraper account, on a 16 ROM library
across 8 platforms whose resources had been scanned by master's code, so no
hashes were recorded to begin with.

1. **Upgrade path.** 16 `jeuInfos` calls, zero media requests, zero bytes
   downloaded. 76 extra media were confirmed current by comparing ScreenScraper's
   md5 against the bytes on disk, and the hashes were recorded for next time.
   Nothing redownloaded, which is the thing to check when a hash column arrives
   empty on an existing library.
2. **Steady state.** Identical counts, and the 1517 file resource tree came out
   byte-identical.
3. **Forced staleness.** Truncated a cover to 5000 bytes (from 431121) and an
   extra media file to 5000 bytes (from 936485), leaving both recorded hashes
   correct. The scan refetched exactly those two and restored both byte for byte,
   with the other 1515 files untouched. The truncated cover is detectable only by
   size, since its recorded md5 still matches what ScreenScraper reports, so this
   also exercises the partial-download check.

That first scan doubles as a check on ScreenScraper's own data: the reported size
matched the stored bytes for all 16 covers and the reported md5 matched for all 76
extra media, since any mismatch would have forced a redownload.

**Things to look at**

- In the scan path, `store_media_file` hashes the file on disk rather than first
  comparing the freshly reported hash against the one the previous scan recorded,
  the way the cover check does. Correct, but it means a scan reads every extra
  ScreenScraper media file in full.
- The first update scan after upgrading re-hashes covers from disk, since no
  recorded md5 exists yet. One-time cost per library.
- A refreshed cover costs two requests, not one, because `get_cover` calls
  `_store_cover` once per size and each fetches the URL. That is existing
  behaviour on any `overwrite=True`, unchanged here, but the hash check gives it a
  new way to trigger. Worth folding into whichever PR takes on the atomic-write
  change, since both live in the same download path.
- Screenshot reordering still goes undetected, because `pydash.xor` compares URL
  sets while files are named by index. That is unchanged from master and this
  branch leaves it no worse; flagged in review and worth its own issue.

**Checklist**

**AI assistance**

This PR was written primarily by Claude Code (Opus 5), including the code, tests
and this description. I reviewed and tested the result locally, and verified the
ScreenScraper media-hash behaviour against a live account.

### DIFF
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -1812,29 +1812,38 @@ async def update_rom(
             if badge_url and badge_path:
                 await fs_resource_handler.store_ra_badge(badge_url, badge_path)
 
-    # Handle special media files from Screenscraper when the ID has changed
-    if cleaned_data["ss_id"] and int(cleaned_data["ss_id"]) != rom.ss_id:
+    # Handle special media files from Screenscraper. A re-match to the same game
+    # still runs, to fill in media that never made it to disk.
+    if cleaned_data["ss_id"]:
+        ss_id_changed = int(cleaned_data["ss_id"]) != rom.ss_id
         preferred_media_types = get_preferred_media_types()
+        # Only a changed ID refetches the game, so only then is there a hash that
+        # describes something other than the copy already on disk. Checking a
+        # stored hash against the file it was recorded from can only ever report
+        # "current", at the cost of reading every media file.
+        fresh_metadata = cleaned_data.get("ss_metadata") if ss_id_changed else None
+        ss_metadata = fresh_metadata or rom.ss_metadata or {}
 
         for media_type in preferred_media_types:
-            # Remove old media files if the ss_id is changing
-            if rom.ss_metadata and rom.ss_metadata.get(f"{media_type.value}_path"):
+            # Media from the previous match is unrelated to the new game
+            if (
+                ss_id_changed
+                and rom.ss_metadata
+                and rom.ss_metadata.get(f"{media_type.value}_path")
+            ):
                 await fs_resource_handler.remove_media_resources_path(
                     rom.platform_id,
                     rom.id,
                     media_type,
                 )
 
-            media_path = cleaned_data.get("ss_metadata", {}).get(
-                f"{media_type.value}_path"
-            )
-            media_url = cleaned_data.get("ss_metadata", {}).get(
-                f"{media_type.value}_url"
-            )
+            media_path = ss_metadata.get(f"{media_type.value}_path")
+            media_url = ss_metadata.get(f"{media_type.value}_url")
             if media_path and media_url:
                 await fs_resource_handler.store_media_file(
                     add_ss_auth_to_url(media_url),
                     media_path,
+                    expected_md5=(fresh_metadata or {}).get(f"{media_type.value}_md5"),
                 )
 
     # Handle local media files from LaunchBox when the ID has changed
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -40,6 +40,7 @@ from handler.metadata.ss_handler import begin_scan as begin_ss_scan
 from handler.metadata.ss_handler import get_preferred_media_types
 from handler.metadata.ss_handler import log_quota as log_ss_quota
 from handler.metadata.ss_handler import log_scan_summary as log_ss_scan_summary
+from handler.metadata.ss_handler import ss_resource_needs_refresh
 from handler.redis_handler import (
     get_job_func_name,
     high_prio_queue,
@@ -512,12 +513,37 @@ async def _identify_rom(
     if scan_type == ScanType.HASHES:
         return
 
+    # ScreenScraper serves every media from a fixed endpoint, so its URLs stay
+    # byte-identical when the artwork behind them is replaced. The md5 it reports
+    # per media entry is the only signal that a stored file went stale.
+    async def _ss_media_changed(
+        media_key: str, resolved_url: str | None, stored_path: str | None
+    ) -> bool:
+        if MetadataSource.SS not in metadata_sources:
+            return False
+
+        return await ss_resource_needs_refresh(
+            previous_metadata=rom.ss_metadata,
+            fresh_metadata=_added_rom.ss_metadata,
+            media_key=media_key,
+            resolved_url=resolved_url,
+            stored_path=stored_path,
+        )
+
     path_cover_s, path_cover_l = await fs_resource_handler.get_cover(
         entity=_added_rom,
-        overwrite=_added_rom.url_cover != rom.url_cover,
+        overwrite=_added_rom.url_cover != rom.url_cover
+        or await _ss_media_changed(
+            "box2d",
+            _added_rom.url_cover,
+            fs_resource_handler.get_downloaded_cover_path(_added_rom),
+        ),
         url_cover=add_ss_auth_to_url(_added_rom.url_cover),
     )
 
+    # Manuals are left out of the hash check: an upload writes `path_manual` but
+    # leaves any scraped `url_manual` in place, so a refresh here would overwrite
+    # it with the provider's copy.
     path_manual = await fs_resource_handler.get_manual(
         rom=_added_rom,
         overwrite=_added_rom.url_manual != rom.url_manual,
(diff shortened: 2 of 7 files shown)


Spinnich:
Thanks @greptileai, this was a useful pass. Two of the three are fixed, details below.

---

### P1: Uploaded manuals lose protection — valid, fixed

Confirmed, and it is a regression this PR introduces rather than a pre-existing gap.

`POST /roms/{id}/manuals` writes only `path_manual` and leaves any scraped `url_manual` in
place, so the "stored without a source URL" test that infers an upload never fires for a
ROM that was previously scraped. Before this PR the upload survived anyway, because
`overwrite` was just `_added_rom.url_manual != rom.url_manual`, which is false when the
provider URL has not changed. The new media hash can set `overwrite=True` on that same
unchanged URL, and `get_manual` then writes the provider's PDF over the uploaded one.

An artwork upload already marks its field explicitly, so the manual endpoints now do the
same: upload marks `url_manual`, and redownload and delete both release it again, since
asking for the provider's manual back is an explicit handover.

---

### P2: Equal hashes skip local validation — valid and reachable, fixed a different way

You are right that the local bytes go unchecked, and it is more reachable than it looks.
The ordering in `_identify_rom` is what makes it bite:

```
485:  _added_rom = db_rom_handler.add_rom(scanned_rom)   # commits ss_metadata, incl. *_md5
533:  ... = await fs_resource_handler.get_cover(...)      # only now downloads the file
```

The hash is committed to the database before the download runs, and downloads stream
straight to their final path rather than to a temp file. So a scan killed mid-stream (OOM,
host reboot, `docker stop` escalating to SIGKILL) leaves a short file described by a hash
that still matches what ScreenScraper reports. `_discard_partial_file` does not help, since
nothing raised. The truncated file then satisfies `cover_exists()` forever.

Reproduced on a real library: truncate `cover/big.png`, leave the recorded md5 correct, run
an update scan. The file stays truncated.

I did not take the suggested edit, because removing the early return means every update
scan hashes every cover, manual and screenshot from disk. Measured on my mock library that
is about 4.4 MB per ROM behind that guard, so roughly 20 GB of reads per update scan at
5,000 ROMs, typically against a NAS.

Instead the fix targets what actually goes wrong. A partial download is always the wrong
*size*, and ScreenScraper reports a byte size next to each hash. `ss_media_is_stale` now
compares that first, which costs a `stat()` and leaves the hash shortcut intact for files
of the right length.

Validated before relying on it, since a size field that did not match the stored bytes
would cause constant redownloads, which is worse than the bug:

- 65 of 65 selected media entries had the reported size exactly equal to the bytes on
  disk, zero mismatches. ScreenScraper lists one entry per region and only the
  region-selected one describes what gets stored, which is worth knowing if you check this
  yourself.
- Same truncation test with the check in place: the cover is refetched and matches the
  original md5 byte for byte.
- A following update scan touches nothing (66 skips, byte-identical resource tree), so the
  fast path is unaffected.

Bit-rot that preserves the file length still slips through. That seemed like the right
place to stop.

Worth noting separately: the underlying cause is the non-atomic write. Streaming to a
temporary name and renaming on completion would make partial files impossible rather than
merely detectable. That is a broader change and would not repair libraries already
carrying truncated files, so I would rather it be its own PR than widen this one.

---

### P3: Screenshot reordering goes undetected — real, but pre-existing

`pydash.xor` is unchanged by this PR. It sits at line 527 on master and this branch only
appends `or await _ss_media_changed(...)` to the `overwrite=` argument below it. The
mismatch between reordered URLs and index-named files is on master today, independent of
these changes.

This PR makes that path strictly better rather than worse, since the ScreenScraper hash
check can now catch changes the set comparison misses. It does not fix ordering. Happy to
open a separate issue, but I would rather not fold an unrelated pre-existing bug into a PR
that is already fairly large.

---

Backend suite is green locally on MariaDB (2739 passed, 2 skipped) and CI covers both
MariaDB and PostgreSQL.

AI assistance: this reply, like the PR itself, was written with Claude Code. The
measurements above are from real runs against a live ScreenScraper account, not estimates.

gantoine (changes requested):
I see what you're trying to do with `user_edited_fields` but I think it's the wrong approach. I'd rather have users lock metadata fields explicitly. Can you update this PR to just be the screenscraper md5 hash comparison stuff?

Spinnich:
> I see what you're trying to do with `user_edited_fields` but I think it's the wrong approach. I'd rather have users lock metadata fields explicitly. Can you update this PR to just be the screenscraper md5 hash comparison stuff?

Scoping down works and for the most part should be easy enough. One piece doesn't
come along cleanly, so I'm curious on your thoughts on it before I start.

The hash check is what creates the problem, so dropping the rest of the PR
doesn't remove it. `POST /roms/{id}/manuals` writes `path_manual` but leaves any
scraped `url_manual` in place, and uploaded and downloaded manuals land at the
same path, so nothing distinguishes them. On master that's harmless: `overwrite`
is `_added_rom.url_manual != rom.url_manual`, which is never true for
ScreenScraper's fixed media endpoints, so an upload survives by accident. The
hash check can set `overwrite=True` on that same unchanged URL, and `get_manual`
then writes the provider's PDF over the uploaded one.

So my plan is to leave manuals out of the hash check in this PR. Covers,
screenshots and extra media get it; manuals keep the current URL-diff behaviour
and nothing regresses. Then the upload/download ambiguity gets its own PR, and
once a scan can tell the two apart, manuals get the hash check like everything
else.

Spinnich:
@greptileai Please re-review.

gantoine:
I'll be honest, this introduces a lot of complexity for something that IMO doesn't necessitate it. If you want to update "stale" assets you can delete the resources folder and run an update metadata scan, which should write them to disk. And I use "stale" in quotes because (personally) I don't really care of some assets are a bit out of date. 🤷🏼

Spinnich:
> I'll be honest, this introduces a lot of complexity for something that IMO doesn't necessitate it. If you want to update "stale" assets you can delete the resources folder and run an update metadata scan, which should write them to disk. And I use "stale" in quotes because (personally) I don't really care of some assets are a bit out of date. 🤷🏼

It's definitely not required for RomM, but the big win I was going for is the ability to easily keep media assets up-to-date with ss.fr without needing to redownload everything for the game(s) being scanned. Ss.fr offers a hash with each media asset which allows us to cheaply compare and only spend ss.fr scrapes on that which actually "needs" to be updated. For larger libraries, this is a boon if you want to efficiently stay synced with whatever is upstream on ss.fr. It also has the bonus of making it easier to repair games where media queries were truncated for whatever reason.

Issue #4029, opened 2026-07-31
[Bug] Smart collections load much slower than the platform they filter


Spinnich:
**RomM version**
5.1.0 (also reproducible on current `master`, 80e58d59b)

**Describe the bug**

Opening a smart collection takes noticeably longer to show games than opening a
platform does — even when the smart collection is just "everything on platform
X" and therefore contains exactly the same games.

The reason is that a smart collection doesn't ask the database "give me page 1
of the games matching these filters". Instead, on **every** request it:

1. re-runs the collection's filters across the entire library,
2. loads every single matching game fully into memory (all metadata included,
   not just the ~70 rows about to appear on screen),
3. saves that list back into the database, and
4. only then asks for the handful of games actually being displayed.

A platform page skips all four steps: it's one indexed database lookup that
returns just the rows needed. So a smart collection is guaranteed to be slower
than the equivalent platform page, and the gap grows with library size rather
than with collection size — a 20-game smart collection costs the same as a
20,000-game one.

It also isn't a one-off cost per visit:

- The **first** load does the expensive step **twice** — once for the game grid
  and once to populate the filter sidebar.
- **Every scroll** that loads more games repeats it again in full.

**Steps to reproduce**

1. Have a reasonably large library (this was noticed on ~92,000 ROMs).
2. Create a smart collection whose only filter is a single platform.
3. Open that platform's page and note how quickly games appear.
4. Open the smart collection and compare — it is visibly slower.
5. Scroll down to load more games; each batch pays the cost again.

**Expected behavior**

A smart collection should be about as fast as the equivalent platform page.
Its filters should be translated into a database query so the database can
return just the page being viewed, instead of the server assembling the whole
membership list in memory first.

**Additional context**

*Where it happens*

`_filter_by_smart_collection_id` in `backend/handler/database/roms_handler.py`
(around line 557) calls `SmartCollection.update_properties()` and then filters
with `Rom.id.in_(...)` over the full member list:

```python
if smart_collection:
    # Ensure the latest ROMs are loaded
    smart_collection = smart_collection.update_properties(user_id)
    return query.filter(Rom.id.in_(smart_collection.rom_ids))
```

`update_properties()` (`backend/models/collection.py`, ~line 231) calls
`get_smart_collection_roms()`, which uses `get_roms_scalar()` with no
`only_fields`. Because `Rom.platform` and `Rom.metadatum` are `lazy="joined"`,
that loads complete ROM objects with their joined metadata — on a large
[...]

PR #4032, opened 2026-07-31
fix(roms): search the gallery by CRC32, MD5, SHA-1 and RA hash

### LINKED ISSUE #4008: [Bug] Search box says you can search by hash, but hash search never returns anything
**RomM version**

5.1.0. Also confirmed on current `master` (`2c8009dc5`). Not a regression — this has never worked.

**Describe the bug**

The search box on the `/search` page reads **"Search by name, filename, hash…"**. Name and filename both work fine. Pasting a hash never returns anything — not for any game, any platform, or any hash type.

It isn't a matching or formatting quirk. The backend search genuinely never looks at the hash columns:

- The UI sends whatever you type to `GET /api/roms?search_term=…`
- That lands in `_filter_by_search_term` in `backend/handler/database/roms_handler.py`, which searches exactly two columns:
  - MariaDB/MySQL: `MATCH(roms.name, roms.fs_name) AGAINST(… IN BOOLEAN MODE)`
  - PostgreSQL fallback: `ILIKE` against the same two columns
- Hashes are stored elsewhere: `roms.crc_hash / md5_hash / sha1_hash` and `rom_files.crc_hash / md5_hash / sha1_hash`. Neither table's hash columns are part of the search.

So the search is doing what it was written to do; the placeholder is just promising something that was never built. Affects every database backend.

Worth noting: RomM *does* have a working hash lookup, it's just a separate API endpoint that nothing in the UI uses — `GET /api/roms/by-hash?crc_hash=…&md5_hash=…&sha1_hash=…&ra_hash=…` (added in #2764). Today, hand-crafting that URL is the only way to find a game by its hash.

**To Reproduce**

1. Open any game's detail page and copy one of its hashes (CRC, MD5, or SHA-1) from the file info.
2. Go to **Search**.
3. Paste the hash into the search box.
4. You get "No games match your search." Try the other two hash types — same result.
5. Sanity check: search that same game by its name, then by its exact filename. Both return it immediately.

Measured directly against the database (`Aero Fighters` / `aerofgt.zip`):

```
"Aero Fighters"                             ->  1 result
"aerofgt.zip"                               ->  1 result
"95b07885"                          (CRC)   ->  0 results
"c5347ceb9cfdabb188d3a1bf5f3a7f94"  (MD5)   ->  0 results
"ddacbb35c87232d701240717aa1ed43ab7f4c253"  ->  0 results  (SHA-1)
```

Uppercase and partial-prefix hashes also return 0.

**Expected behavior**

When the search term looks like a hash — 8 hex characters (CRC32), 32 (MD5), or 40 (SHA-1) — also match it case-insensitively against the hash columns on both `roms` and `rom_files`, and return the owning game. Matching on `rom_files` matters, because for multi-file games (multi-disc, multi-track) the hash a user has in hand is usually a *file's* hash, not the game's. Supporting RA hashes too would be a bonus. Note there is currently no database index on any of these hash columns, so a fix should add one to keep large libraries fast.

**Additional context**

- Placeholder string: `rom.search-placeholder` in `frontend/src/locales/en_US/rom.json`, used by `frontend/src/v2/views/Gallery/Search.vue`. It was hardcoded when the v2 gallery was written and later extracted for translation; the backend was never extended to match.
- **Smart collections are affected too.** The smart-collection "search term" criterion runs through the same filter (`collections_handler.py` → `filter_roms`), so a smart collection built on a hash will always be empty.
- The old (v1) UI just labels its box "Search", so it makes no such promise — but it's equally unable to find a game by hash, since both UIs share this backend filter.
- I searched open and closed issues and PRs and found nothing covering this.

Spinnich:
**Description**

Fixes #4008.

The search box promises "Search by name, filename, hash…", but `_filter_by_search_term`
only ever queried `roms.name` and `roms.fs_name`, so pasting a hash always returned
nothing. This extends that filter to the hash columns.

When a term is hex of exactly a digest length (8 for CRC32, 32 for MD5 and RA, 40 for
SHA-1) it is also matched, lowercased, against the hash columns on `roms` **and**
`rom_files`. Matching `rom_files` is what makes multi-disc and multi-track games work,
since the hash a user has in hand is usually a file's rather than the game's.

Two things shaped the implementation:

**1. Name search must not get worse.** A term only reaches the hash columns if it is hex
of exactly a digest length, so an ordinary name search compiles to byte-identical SQL to
before. When a hash-shaped term *is* present, both sides still run: a game named
`deadbeef` and a different game whose CRC is `deadbeef` both come back. Detecting a hash
never costs you the name match.

**2. The two sides are unioned, not OR'd.** This is the important part. No engine merges
a full-text index with a B-tree index, so `MATCH(...) OR hash = ...` discards the
fulltext index and scans `roms` end to end. Measured on a 300k-row MariaDB table:

| Query shape | Time | Plan |
| --- | --- | --- |
| Name only (baseline) | 0.294s | `fulltext` |
| Hash only, indexed | 0.038s | `index_merge` |
| `MATCH(...) OR hash = ...` | 3.306s | `ALL` (full scan) |
| `id IN (name UNION hash)` | 14.271s | `DEPENDENT SUBQUERY` |
| `id IN (SELECT ... FROM (name UNION hash) t)` | **0.005s** | every branch indexed |

Pre-resolving hash ids in Python and OR-ing them in was also measured, and
`MATCH(...) OR id IN (const)` is a full scan for the same reason. So each side resolves
through its own index and the id sets are unioned via a derived table, which forces
MariaDB to materialize instead of correlating.

Neither table had an index on any hash column, so the migration adds them.

No frontend changes: the v2 placeholder already advertised this, and v1 is frozen.

**Files modified**

| File | What changed |
| --- | --- |
| `backend/handler/database/roms_handler.py` | Split `_filter_by_search_term` into `_build_name_conditions` (the existing logic, unchanged) and a new `_build_hash_selects`. When a digest-shaped term is present, the two id sets are unioned through a derived table; otherwise the query is exactly what it was before. |
| `backend/models/rom.py` | Declared the nine hash indexes on `Rom` and `RomFile` so `alembic revision --autogenerate` does not propose dropping them. |
| `backend/alembic/versions/0106_hash_search_indexes.py` | New migration creating those indexes. Mirrors `0104_roms_missing_from_fs_index.py`, including `if_not_exists` / `if_exists`. |
| `backend/tests/handler/database/test_roms_hash_search.py` | New, 22 tests. |

**Testing notes**

- Full backend suite on MariaDB: 2653 passed, 2 skipped, 0 failed.
- New file: 22/22 on MariaDB, and re-run against a scratch PostgreSQL 16 alongside the
  existing `test_db_handler.py` (59 passed) to confirm the ILIKE fallback path.
- Migration verified `upgrade` + `downgrade` on MariaDB, and `upgrade` on PostgreSQL.
- The tests were written before the implementation. To confirm they are meaningful I
  stashed only the handler change and re-ran: the hash and smart-collection tests fail
  without it and pass with it.

Coverage spans each digest type, case-insensitivity, file-level hashes on a multi-disc
game, deduplication when several files share a hash, and a set of tests asserting name
search is not degraded.

**Anything a reviewer should pay attention to**

1. **The union-instead-of-OR shape is load-bearing**, not a style choice. Three tests in
   `TestSearchQueryShape` assert on the compiled SQL specifically to stop it silently
   regressing to an `OR`, which is a ~600x cliff that no functional test would catch.
   They are white-box, and that is deliberate.
2. **`rom_files.chd_sha1_hash` is included**, slightly beyond the column list in the
   issue. It is the SHA-1 datfiles publish for a CHD, so it is the digest a user is most
   likely to paste for a CHD. Easy to drop if you would rather keep to the issue's scope.
3. **Matching is exact-length only.** A partial prefix still returns nothing, because
   prefix matching would forfeit the index and put us back on a full scan. Worth
   confirming that is the trade-off you want.
4. **Nine new indexes.** Creating them is online in InnoDB but not instant on a large
   library, and they add a small write cost to every scan that populates hashes.
5. **Relevance ordering.** `_build_fulltext_relevance` only emits a clause for
   multi-word terms, so a pure hash search is unaffected. On a mixed
   `Super Mario|<hash>` search the hash-matched game scores zero relevance and sorts
   last. That seemed right, but it is a judgment call.
6. **Smart collections are fixed as a side effect.** Their "search term" criterion runs
   through this same filter, so a hash-based smart collection was always empty. Two
   end-to-end tests cover it. This is unrelated to #4029, which is about smart-collection
   *load performance* and is untouched here. Note that #4029 touches
   `_filter_by_smart_collection_id`, roughly 50 lines from the function this PR changes,
   so expect a light conflict if both land.

**AI assistance disclosure**

This PR was written primarily by Claude Code (Opus 5), covering the implementation,
tests, migration, benchmarking and this description, under my direction and review. The
performance investigation that led to the union shape was run against a real 300k-row
MariaDB table rather than reasoned about. Any follow-up review responses I post with AI
help will say so.

**Checklist**

**Screenshots**
image

### DIFF
--- a/backend/alembic/versions/0106_hash_search_indexes.py
+++ b/backend/alembic/versions/0106_hash_search_indexes.py
@@ -1,0 +1,43 @@
+"""Index the hash columns searched by the gallery search box
+
+Searching by a CRC32/MD5/SHA-1/RA digest now matches the hash columns on
+``roms`` and ``rom_files``. Neither table had an index on any of them, so every
+such search scanned both tables in full.
+
+Revision ID: 0106_hash_search_indexes
+Revises: 0105_fix_gamelist_epoch_ms
+Create Date: 2026-07-31 00:00:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0106_hash_search_indexes"
+down_revision = "0105_fix_gamelist_epoch_ms"
+branch_labels = None
+depends_on = None
+
+HASH_INDEXES: dict[str, tuple[str, ...]] = {
+    "roms": ("crc_hash", "md5_hash", "sha1_hash", "ra_hash"),
+    "rom_files": ("crc_hash", "md5_hash", "sha1_hash", "ra_hash", "chd_sha1_hash"),
+}
+
+
+def upgrade() -> None:
+    for table, columns in HASH_INDEXES.items():
+        with op.batch_alter_table(table, schema=None) as batch_op:
+            for column in columns:
+                batch_op.create_index(
+                    f"idx_{table}_{column}",
+                    [column],
+                    unique=False,
+                    if_not_exists=True,
+                )
+
+
+def downgrade() -> None:
+    for table, columns in HASH_INDEXES.items():
+        with op.batch_alter_table(table, schema=None) as batch_op:
+            for column in columns:
+                batch_op.drop_index(f"idx_{table}_{column}", if_exists=True)
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -25,6 +25,7 @@ from sqlalchemy import (
     select,
     text,
     true,
+    union,
     update,
 )
 from sqlalchemy.orm import (
@@ -131,6 +132,27 @@ FULLTEXT_BOOLEAN_OPERATORS_REGEX = re.compile(r'[+\-~<>()"@*]')
 
 # 3 is the default minimum size in InnoDB
 FULLTEXT_MIN_TOKEN_SIZE = 3
+
+# A term reaches the hash columns only when it is hex of exactly a digest
+# length, so an ordinary name search builds no hash SQL at all. Hashes are
+# stored lowercase, which keeps the lookup an indexed equality.
+HEX_DIGEST_REGEX = re.compile(r"[0-9a-fA-F]+")
+
+# CRC32 (8), MD5 and RetroAchievements (32), SHA-1 (40).
+ROM_HASH_COLUMNS_BY_DIGEST_LENGTH: dict[int, tuple[QueryableAttribute, ...]] = {
+    8: (Rom.crc_hash,),
+    32: (Rom.md5_hash, Rom.ra_hash),
+    40: (Rom.sha1_hash,),
+}
+
+# Multi-file games (multi-disc, multi-track) keep their hashes per file, which
+# is the hash a user has in hand. `chd_sha1_hash` is the uncompressed disc's
+# digest, the one datfiles publish for a CHD.
+ROM_FILE_HASH_COLUMNS_BY_DIGEST_LENGTH: dict[int, tuple[QueryableAttribute, ...]] = {
+    8: (RomFile.crc_hash,),
+    32: (RomFile.md5_hash, RomFile.ra_hash),
+    40: (RomFile.sha1_hash, RomFile.chd_sha1_hash),
+}
 
 # Filter dropdowns read the narrow `roms_facets` mirror instead of `roms`,
 # whose rows carry the raw metadata blobs. Column order matches the unpacking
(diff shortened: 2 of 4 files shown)


gantoine (approved):
👍🏼

PR #4039, opened 2026-07-31
perf(collections): serve smart and standard collections from a composed query

### LINKED ISSUE #4029: [Bug] Smart collections load much slower than the platform they filter

Spinnich:
**Description**

Fixes #4029.

A smart collection re-ran its filters across the entire library on every
request, loaded every match fully into memory, wrote the membership back to
the database, and only then queried the page. Cost tracked library size rather
than collection size, and it was paid twice on first load (grid + filter
sidebar) and again on every scroll.

This translates the stored `filter_criteria` into predicates applied directly
to the ROM query, so the database returns just the page being viewed. The
criteria are `filter_roms`'s own vocabulary already, so membership composes
into SQL as an indexed subquery. `EXPLAIN` confirms MariaDB flattens it into a
semi-join rather than running it as a dependent subquery.

Regular collections get the other half of #3886: `collections_roms` is keyed on
`(collection_id, rom_id)`, so membership is now a subquery instead of a list of
ids fetched into Python. The favorites filter uses the same shape.

Two side effects from the issue are gone with it:

- Serving a page no longer bumps `updated_at`, so cover thumbnails stop
  re-downloading on every visit and `updated_after` sync clients stop seeing
  constant modification.
- A shared or public smart collection no longer stores a count computed for
  whoever viewed it last. The cached columns are refreshed on write and always
  describe the **owner's** view, while the gallery itself resolves per viewer.

Cached membership is maintained where the library actually changes: smart
collection create/update, post-scan, `delete_roms`, `update_rom`, and the
metadata reset. The per-ROM paths only recount the collections that hold a
changed id (that check resolves as a primary-key lookup), so editing one ROM
does not scan the library once per collection. Every call site is guarded: the
ROM write is already committed by then, so a refresh failure must not surface
as a failed edit or cost a bulk delete its per-ROM report.

**Behavior change worth a reviewer's attention:** an unknown `collection_id` or
`smart_collection_id` now returns no ROMs instead of silently falling through
to the whole library. A stale bookmark used to render as "everything".

**Known limitation:** cover-only changes (artwork upload) do not refresh the
cached cover mosaic, since the per-ROM path compares membership rather than
cover paths. It corrects itself on the next scan.

Backend only. No API schema change and no migration.

**AI assistance disclosure**

This PR was written primarily by Claude Code (Opus 5), tests first, then
reviewed and verified by me. This description was also drafted by Claude Code.
PR responses will be written by me unless stated otherwise.

**Checklist**


Spinnich:
Good catch @greptileai , this one is valid and is now fixed in 7399c1947.

Removing the read-time recompute did leave the cached membership blind to
anything that isn't a ROM edit. Scope note for anyone reading later: only the
cached display metadata went stale (collection-list counts, cover mosaics,
ROM-detail chips). The gallery itself composes the criteria into SQL on every
request, so browsing a smart collection always showed correct contents.

Wired the refresh into the paths you named, plus two more in the same class:

- Save and state upload, delete, **and the visibility toggle**. That last one
  matters because `has_saves` matches "their own plus other users' public
  saves", so sharing an asset moves collections belonging to people other than
  the uploader. That detail also ruled out my first approach of scoping the
  refresh to the acting user's own collections.
- `update_rom_user` when a status is set.
- Adding and removing regular-collection members, which covers both the
  `collection_id` and `favorite` criteria.

One thing worth flagging, since the naive version of this fix would have been
worse than the bug: autosaves land constantly during play, and the existing
refresh recounts any collection the ROM already belongs to (deliberately, since
a rename can change sort order or the cover mosaic). So every autosave would
have forced a full recount of every `has_saves` collection.

`refresh_smart_collections_for_roms` therefore takes a `membership_only` flag.
All of the paths above leave the ROM row itself untouched, so nothing can move
except whether it matches, and a member that stayed a member needs no recount.
There's a test pinning exactly that: a second save on an already-member ROM
performs zero refreshes.

Added six tests, including the cache-invalidation coverage for saves and
statuses that you flagged as missing. Full backend suite is green at 2659
passed / 2 skipped, trunk clean.

Known residual, left deliberately: deleting a regular collection that a smart
collection filters on leaves that count stale until the next scan. It is rare
and self-healing, and covering it would mean triggering a full refresh from a
non-ROM event. Happy to close it too if you'd rather.

<sub>This reply was drafted with Claude Code (Opus 5) and reviewed by me before
posting.</sub>

gantoine (approved):

gantoine:
@greptileai

PR #4059, opened 2026-08-01
fix(metadata): identify archives by their largest member again

### LINKED ISSUE #3742: [Bug] Hasheous and Playmatch hash matching broken for multi-file archives since 4.9.0 (composite hash sent, but they index by largest file)
**RomM version**
4.9.0 and later (current master, 66596c8). 4.8.x is not affected.

**Describe the bug**
Some ROMs are zip files containing several files instead of one. Arcade (MAME/FBNeo) sets are the most common example, but this applies to any platform.

Since 4.9.0 (#3412), RomM hashes these archives by decompressing every file inside and computing one combined "composite" hash over all of them. That was Phase 1 of #3411, whose goal was populating the ROM-level hash fields for archives. Phase 2, deciding which hash each metadata provider should receive during lookup, was explicitly deferred ("this needs separate consideration" in the PR description) and never happened.

The unintended side effect: the metadata handlers for Hasheous and Playmatch read their hashes from the archive's file record, and that value silently changed from "hash of the largest file inside the archive" (pre-4.9.0) to the new composite hash. Both services index a multi-file archive by its largest internal file (the Playmatch developer confirmed the largest-file rule). The composite value matches nothing in their databases, so hash-based identification for multi-file archives silently fails on 4.9.0+. Those ROMs go unmatched or fall back to weaker filename matching.

Verified against the live APIs on 2026-08-07:

- **Hasheous**: the two MAME sets I tested return HTTP 404 on the composite and resolve on the largest member, to "Aero Fighters" and "Golden Axe: The Revenge of Death Adder" (signatures `MAMEArcade`, `FBNeo`).
- **Playmatch**: same rule for the datasets it carries, but it does not appear to index MAME sets at all. Both arcade sets return `NoMatch` on the composite *and* on the largest member, so **arcade is a Hasheous-only regression**. The Playmatch half reproduces on any non-arcade multi-file archive: a No-Intro ROM zipped with a readme returns `NoMatch` on the composite and `gameMatchType='SHA1'` on the largest member.

ScreenScraper still works in practice, but not because it matches the composite hash: its lookup endpoint falls back to the filename (`romnom`) when the hashes miss, and for arcade zips the filename is the set name, which ScreenScraper resolves directly.

Where this happens in the code (permalinks to master at 66596c8):

- Composite hash computed for archives: [`roms_handler.py#L450-L511`](https://github.com/rommapp/romm/blob/66596c85615967a4779e46023e37dab2cdc78053/backend/handler/filesystem/roms_handler.py#L450-L511)
- Sent to Hasheous: [`hasheous_handler.py#L258-L274`](https://github.com/rommapp/romm/blob/66596c85615967a4779e46023e37dab2cdc78053/backend/handler/metadata/hasheous_handler.py#L258-L274)
- Sent to Playmatch: [`playmatch_handler.py#L222-L232`](https://github.com/rommapp/romm/blob/66596c85615967a4779e46023e37dab2cdc78053/backend/handler/metadata/playmatch_handler.py#L222-L232)

**To Reproduce**

1. Run RomM 4.9.0+ with Hasheous and/or Playmatch enabled as metadata sources.
2. Add a multi-file zip to the library. Any ROM zipped together with a readme reproduces it on both services; a MAME arcade set reproduces it on Hasheous only, for the reason above.
3. Scan the platform.
4. The ROM gets no hash match, even though the same file matched on 4.8.x.

**Expected behavior**
Multi-file archives should be identified by Hasheous and Playmatch like they were before 4.9.0.

The fix should not require changing how ROMs are hashed. Since 4.9.0 the scanner already stores per-file hashes for every file inside an archive (`archive_members` on the file record), so the Hasheous and Playmatch handlers can select the largest member's hashes from there when the file is an archive. That completes Phase 2 of #3411 as originally planned, and works on existing libraries without re-hashing anything. The composite hash should stay as the stored ROM-level identity, which is what #3411 set out to add.

**Additional context**
Single-file zips are unaffected: the composite hash of one file equals that file's own hash, which is also what the pre-4.9.0 code produced. That is why most non-arcade libraries don't notice, and why this shows up primarily on arcade sets, where multi-file archives are the norm.

RetroAchievements matching is a separate pipeline and is unaffected by this change: for arcade, the RA hash is derived from the zip's filename, and for other platforms RAHasher cannot process multi-file zips at all regardless of what RomM computes.

Related: #3411 (Phase 1), #3412 (implementation).

gantoine:
**Description**

Completes phase 2 of #3411, which #3412 deferred.

Since 4.9.0 an archive's `RomFile` carries a **composite** hash computed across every entry inside it. The Hasheous and Playmatch lookups read their digests straight off that file record, so they silently switched from "hash of the largest file inside the archive" (what 4.8.x sent) to the composite. Both services index a multi-file archive by its largest internal file, so the composite matches nothing and those ROMs stopped resolving by hash — most visibly on arcade sets, where multi-file zips are the norm.

`RomFile.lookup_hashes` now owns the question of *which* digests a ROM database should be queried with:

| file | hashes sent |
| --- | --- |
| CHD | the disc-data SHA1 from its header (unchanged, just moved here) |
| multi-file archive | the largest member's crc / md5 / sha1, read from `archive_members` |
| anything else | the file's own crc / md5 / sha1 |

Both `hasheous_handler.lookup_rom` and `playmatch_handler.lookup_rom` use it, as does the Playmatch *suggestion* payload — contributing a composite hash back to their index would have been worse than not contributing at all.

Notes:

- Nothing is re-hashed. `archive_members` has been populated since 4.9.0, so existing libraries are fixed by a rescan without touching the files.
- The composite stays the stored ROM-level identity, which is what #3411 set out to add.
- Single-file archives are unaffected: a composite over one entry equals that entry's hash.
- ScreenScraper is deliberately left alone. It reads the same file-record digests, but its lookup falls back to `romnom` (the filename), which for arcade zips is the set name, so it still resolves. Changing it would shift match results for a lot of libraries and belongs in its own change.

**Checklist**

Tests added: `RomFile.lookup_hashes` (own digests / CHD / largest member) in `tests/models/test_rom.py`, plus a payload assertion for each of the two handlers.

n/a — backend only.

Fixes #3742

---

🤖 AI assistance disclosure: this change was written with Claude Code (Claude Opus 5), working from the analysis in the issue. Reviewed by me before opening.

### DIFF
--- a/backend/handler/metadata/hasheous_handler.py
+++ b/backend/handler/metadata/hasheous_handler.py
@@ -269,17 +269,12 @@ class HasheousHandler(MetadataHandler):
         # against any of them.
         data: list[dict] = []
         for file in filtered_files:
-            file_hashes: dict[str, str | None]
-            if file.chd_sha1_hash:
-                # CHD files are indexed by disc-data SHA1 only
-                # Raw file MD5/CRC are hashes of the container and won't match
-                file_hashes = {"shA1": file.chd_sha1_hash}
-            else:
-                file_hashes = {
-                    "mD5": file.md5_hash,
-                    "shA1": file.sha1_hash,
-                    "crc": file.crc_hash,
-                }
+            hashes = file.lookup_hashes
+            file_hashes: dict[str, str | None] = {
+                "mD5": hashes.md5,
+                "shA1": hashes.sha1,
+                "crc": hashes.crc,
+            }
 
             # Drop empty hashes and skip files that have none.
             file_hashes = {key: value for key, value in file_hashes.items() if value}
--- a/backend/handler/metadata/playmatch_handler.py
+++ b/backend/handler/metadata/playmatch_handler.py
@@ -228,15 +228,17 @@ class PlaymatchHandler(MetadataHandler):
         if first_file is None:
             return fallback_rom
 
+        hashes = first_file.lookup_hashes
+
         try:
             response = await self._request(
                 self.identify_url,
                 {
                     "fileName": first_file.file_name,
                     "fileSize": first_file.file_size_bytes,
-                    "md5": first_file.md5_hash,
-                    "sha1": first_file.sha1_hash,
-                    "crc": first_file.crc_hash,
+                    "md5": hashes.md5,
+                    "sha1": hashes.sha1,
+                    "crc": hashes.crc,
                 },
             )
         except Exception as exc:
@@ -307,8 +309,8 @@ class PlaymatchHandler(MetadataHandler):
                 None,
             )
             if first_file is not None:
-                md5 = first_file.md5_hash
-                sha1 = first_file.sha1_hash
+                md5 = first_file.lookup_hashes.md5
+                sha1 = first_file.lookup_hashes.sha1
                 file_name = first_file.file_name
                 file_size: int | None = first_file.file_size_bytes
             else:
--- a/backend/models/rom.py
+++ b/backend/models/rom.py
@@ -5,7 +5,7 @@ import enum
 import re
 from datetime import datetime
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, TypedDict
+from typing import TYPE_CHECKING, Any, NamedTuple, TypedDict
 
 from sqlalchemy import (
     TIMESTAMP,
@@ -103,6 +103,14 @@ class RomArchiveMember(TypedDict):
     sha1_hash: str
 
 
+class LookupHashes(NamedTuple):
+    """The hashes a ROM-database provider should be queried with."""
+
+    crc: str | None
+    md5: str | None
+    sha1: str | None
+
+
 class RomFile(BaseModel):
     __tablename__ = "rom_files"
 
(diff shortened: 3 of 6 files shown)


Spinnich:
Maybe I'm misunderstanding, but for multi-file folder .chd games submitted to Playmatch, PlayMatch doesn't get the largest top-level file's CHD SHA-1 hash, it just uses first file with a size > 0? This would cause .m3u files to get submitted I'd think...

Spinnich:
Picking this up. I'll re-land this branch with the file-selection problem from my
comment above fixed in the same PR.

To be clear about scope, there are two independent causes and this only fixed one:

1. the wrong hash of the right file (the composite, a 4.9.0 regression), which is what
   this branch fixes
2. the wrong file entirely, which is `first_file` picking the first file with bytes
   rather than the largest, unfiltered by `is_top_level`. That one is pre-existing, not
   a 4.9.0 regression: 4.8.1 has the same lines, from ca58028eb. It's also worse than I
   framed it originally. `iter_files` is a bare `os.walk` with no sort and `files` has
   no `order_by`, so the file Playmatch is asked about is arbitrary filesystem order and
   can differ between two machines scanning the same library.

Fixing only 1 leaves multi-disc folder ROMs still failing on Playmatch, so I'd rather
they went together. Shout if you'd prefer them split or want to take it back yourself.

PR #4069, opened 2026-08-02
fix(home): stop scanning the library filesystem on every home page load

### LINKED ISSUE #4063: [Bug] Home page runs the first-time-setup filesystem scan on every load, even with a full library
**RomM version**

5.1.0. Affected code unchanged on `master` (79a832247).

**Describe the bug**

The v2 home page calls `GET /api/setup/library` on every load. That endpoint walks
the ROM library on disk, listing the contents of every platform folder to count
files.

It's only supposed to run for brand-new users: it powers the "here's what RomM can
already see on disk" hint shown when your library looks empty, so the "run a scan"
prompt isn't a leap of faith. But it fires for everyone, on every home page visit,
including users with a fully populated library.

The cause looks like a timing problem rather than a logic problem. The check is
wired to an `isEmpty` flag, but that flag is evaluated immediately when the page
sets up — before any of the home page's data requests have started. At that moment
nothing has loaded and no fetch is marked in-flight, so the page looks empty and
the filesystem scan kicks off. The data arrives a moment later and `isEmpty`
becomes false, but the request is already gone.

**To Reproduce**

1. Use a populated library (the empty-state hint should never be shown).
2. Log in and land on the home page.
3. Watch the browser network tab: `GET /api/setup/library` fires anyway.
4. On a large library, note how long it takes.

**Expected behavior**

`/api/setup/library` should only be requested when the library is genuinely empty
— after the home page's data has loaded and confirmed there's nothing there. Users
with a populated library should never trigger it.

**Screenshots**

Reproduced on two independent 5.1.0 instances:

| Instance | Time | Second run |
| --- | --- | --- |
| 83,000 games, DB on SSD | **17.0 s** | 16.8 s |
| 23,698 games, DB on 5200 RPM HDD | **38.3 s** | — |

Worth stressing: the 17 second result is from a healthy, well-provisioned instance
whose database is on SSD. This endpoint doesn't touch the database at all, so fast
database storage doesn't help — what matters is the library storage it walks. The
second run showed no improvement, so OS page cache doesn't rescue it either.

Every home page load on that instance spends 17 seconds on a result that is then
thrown away, because the library isn't empty.

**Additional context**

Affected code:

- `frontend/src/v2/views/Home.vue` — `loadLibraryInfo()` is triggered by a
  `watch(isEmpty, ...)` with `{ immediate: true }`. The store `fetching*` flags all
  start `false`, so on the immediate run `isEmpty` evaluates true before any fetch
  has begun. `loadLibraryInfo()` has no other call site.
- `backend/endpoints/heartbeat.py` — `get_setup_library_info()` does an
  `iterdir()` per platform directory under the library path.

Two complementary fixes:

- **Frontend:** drop `immediate: true`, or gate on "loading has finished and the
  result is empty" rather than on the transient initial state.
- **Backend:** bail out early when the database already has platforms and ROMs,
  since the filesystem hint is meaningless then.

Supporting detail: in a captured page load this request fired in the same
millisecond as the page's other initial requests, consistent with running during
setup rather than after data had loaded.

Spinnich:
**Description**

Fixes #4063

The v2 home page called `GET /api/setup/library` on **every** page load, even on a
fully populated library. That endpoint walks the ROM library on disk (`iterdir()`
per platform directory) to build a first-run "here's what we see on disk" hint,
and the home page threw the result away because the library isn't empty.

Measured cost of the wasted walk:

- 17.0s on an 83,000-game instance with the DB on SSD (16.8s on a second run, so
  the OS page cache doesn't rescue it)
- 38.3s on a 23,698-game instance with a 5200 RPM HDD

**Root cause.** `loadLibraryInfo()` was driven by `watch(isEmpty, ..., { immediate: true })`.
An immediate watcher runs during `setup()`, which is strictly earlier than
`onMounted`, so at that moment no data request has even started: every store
`fetching*` flag is still `false` and every store array is still empty. `isEmpty`
therefore read `true` for any library, and the walk fired unconditionally.

Fixed on both sides of the stack, since either one alone leaves a sharp edge:

1. **Frontend.** `onMounted` now collects its conditional fetches and awaits
   `Promise.allSettled` before flipping a new `initialLoadDone` flag. The
   watcher and the empty-state `v-if` both gate on
   `showEmptyState = initialLoadDone && isEmpty`, so the filesystem hint is
   requested only once the library is *confirmed* empty. `allSettled` rather
   than `all` so a rejected fetch still releases the gate.
2. **Backend.** `/setup/library` returns early with `existing_platforms: []`
   when the database already holds at least one ROM, so any other caller (or an
   older frontend) can't trigger the walk either.

**Note on scope of the user-visible symptom.** This is a wasted-request fix, not
a visible-flash fix. I verified with a `requestAnimationFrame` frame sampler that
the empty-library hero rendered in **0 frames** on a populated library both before
and after the change: `onMounted` sets the fetching flags before the first paint,
so only the setup-time watcher was ever wrong. Please don't expect a visual
difference on a populated library; the difference is one fewer multi-second
request per page load.

**Design note for reviewers.** The backend bail-out is gated on **ROMs only**,
deliberately not on "platforms and ROMs". A ROM row implies a platform row via
FK, and a database with platform rows but zero ROMs is precisely the case the
filesystem hint exists to serve, so that case must still walk. Both behaviours
are pinned by tests.

**Files modified**

| File | Change |
| --- | --- |
| `frontend/src/v2/views/Home.vue` | `onMounted` awaits `Promise.allSettled` of the initial fetches, then sets `initialLoadDone`. New `showEmptyState` computed gates both the watcher and the empty-state `v-if`. `immediate: true` removed from the watcher. |
| `frontend/src/v2/views/Home.test.ts` | New. Covers the populated-library case (no request, at setup time or after settle), the empty-library case (exactly one request), and the template gate. |
| `backend/endpoints/heartbeat.py` | `get_setup_library_info` returns early with an empty `existing_platforms` when `db_stats_handler.get_roms_count() > 0`. Placed after the existing scope check, and still returns `detected_structure` + `supported_platforms`. |
| `backend/handler/database/roms_handler.py` | New `has_any_rom()` existence probe (`SELECT id LIMIT 1`, not `COUNT(*)`). |
| `backend/tests/endpoints/test_heartbeat.py` | Two tests: the walk is skipped when ROMs exist (asserts `fs_platform_handler.get_platforms` is never called), and the walk still happens for platform rows with no ROMs. |

**Testing notes**

- Backend: `uv run pytest` → 2704 passed, 2 skipped.
- The ROM existence check reuses `db_stats_handler.get_roms_count()` per review.
  On a 92,800-ROM replica (`roms` 963 MB, 256 MB buffer pool) it is an
  index-only scan of a 2.5 MB secondary index (`type=index`, `Using index`),
  measured at 25 ms cold and warm, against the 17-38 s walk it prevents.
- Frontend: `npm run test` → 615 passed across 48 files. `vue-tsc --noEmit` clean.
- Browser-verified against a real dev stack in both light and dark themes:
  - Populated library: **0** `/api/setup/library` requests (pre-fix: 1, firing in
    the same millisecond as the other initial requests).
  - Empty library: **1** correctly-deferred request, with the "9 platforms
    detected" / "16 games detected" chips still rendering as before.
  - The template gate removes an intermediate paint for empty-library users
    rather than adding one (`rows:5 → rows:4 → hero @377ms` became
    `rows:5 → hero @271ms`).
- No API contract change: the endpoint has no `response_model` and the response
  keys are unchanged, so no `npm run generate` is needed.
- No new user-visible strings, so no locale changes.
- No v1 files touched.

**Reviewer attention**

- **The ROMs-only gating** described above is the one judgement call in this PR
  that the issue text did not specify. If you'd rather it also considered
  platform rows, that flips the behaviour of
  `test_get_setup_library_info_walks_when_platforms_have_no_roms`.
- **Setup wizard is unaffected**, but it's the natural regression risk and worth
  a second pair of eyes: `SHOW_SETUP_WIZARD` is only true when there are zero
  admin users, at which point there are no ROMs, so the bail-out cannot trigger
  mid-wizard. The early return also preserves `detected_structure` and all 460
  `supported_platforms` entries the wizard reads.
- **`Promise.allSettled` semantics**: a failed initial fetch still flips
  `initialLoadDone`, which is intentional. A user whose requests all failed will
  see the empty state and one filesystem-hint request, which is the same thing
  they'd have seen before.

**Checklist**

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this change was written with AI assistance (Claude Code).
The AI wrote the tests and implementation, ran the test suites, and performed the
browser verification described above. I reviewed the diff, and the measurements in
the linked issue are my own from my instances.

### DIFF
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -1643,6 +1643,19 @@ class DBRomsHandler(DBBaseHandler):
         )
 
     @begin_session
+    def has_any_rom(
+        self,
+        *,
+        session: Session = None,  # type: ignore
+    ) -> bool:
+        """Whether the library holds at least one ROM.
+
+        An existence probe rather than a count, for callers that only
+        need to know whether the library has ever been scanned.
+        """
+        return session.scalar(select(Rom.id).limit(1)) is not None
+
+    @begin_session
     def get_roms_by_fs_name(
         self,
         platform_id: int,
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -1643,19 +1643,6 @@ class DBRomsHandler(DBBaseHandler):
         )
 
     @begin_session
-    def has_any_rom(
-        self,
-        *,
-        session: Session = None,  # type: ignore
-    ) -> bool:
-        """Whether the library holds at least one ROM.
-
-        An existence probe rather than a count, for callers that only
-        need to know whether the library has ever been scanned.
-        """
-        return session.scalar(select(Rom.id).limit(1)) is not None
-
-    @begin_session
     def get_roms_by_fs_name(
         self,
         platform_id: int,
--- a/backend/endpoints/heartbeat.py
+++ b/backend/endpoints/heartbeat.py
@@ -29,7 +29,7 @@ from decorators.auth import protected_route
 from endpoints.responses.heartbeat import HeartbeatResponse
 from exceptions.fs_exceptions import PlatformAlreadyExistsException
 from handler.auth.constants import Scope
-from handler.database import db_user_handler
+from handler.database import db_rom_handler, db_user_handler
 from handler.filesystem import fs_platform_handler
 from handler.filesystem.base_handler import LibraryStructure
 from handler.metadata import (
@@ -213,6 +213,17 @@ async def get_setup_library_info(request: Request):
         )
 
     detected_structure = fs_platform_handler.detect_library_structure()
+
+    # The per-platform rom counts below are a first-run hint, so a fresh
+    # instance can show what RomM already sees on disk. Once the database
+    # holds ROMs that hint is dead weight, and building it walks every
+    # platform directory: tens of seconds on a large library.
+    if db_rom_handler.has_any_rom():
+        return {
+            "detected_structure": detected_structure,
+            "existing_platforms": [],
+            "supported_platforms": get_supported_platforms(),
+        }
 
     # Get existing platforms from filesystem
     try:
--- a/backend/endpoints/heartbeat.py
+++ b/backend/endpoints/heartbeat.py
@@ -29,7 +29,7 @@ from decorators.auth import protected_route
 from endpoints.responses.heartbeat import HeartbeatResponse
 from exceptions.fs_exceptions import PlatformAlreadyExistsException
 from handler.auth.constants import Scope
-from handler.database import db_rom_handler, db_user_handler
+from handler.database import db_stats_handler, db_user_handler
 from handler.filesystem import fs_platform_handler
 from handler.filesystem.base_handler import LibraryStructure
 from handler.metadata import (
@@ -218,7 +218,7 @@ async def get_setup_library_info(request: Request):
     # instance can show what RomM already sees on disk. Once the database
     # holds ROMs that hint is dead weight, and building it walks every
     # platform directory: tens of seconds on a large library.
-    if db_rom_handler.has_any_rom():
+    if db_stats_handler.get_roms_count() > 0:
         return {
             "detected_structure": detected_structure,
             "existing_platforms": [],
--- a/backend/tests/endpoints/test_heartbeat.py
+++ b/backend/tests/endpoints/test_heartbeat.py
@@ -239,6 +239,72 @@ def test_get_setup_library_info_handles_errors(client, admin_user, access_token)
             assert data["existing_platforms"] == []
 
 
+def test_get_setup_library_info_skips_filesystem_walk_when_roms_exist(
+    client, rom, access_token
+):
+    """A library with scanned ROMs never needs the on-disk hint, so skip the walk."""
+    with (
+        patch(
+            "endpoints.heartbeat.fs_platform_handler.detect_library_structure"
+        ) as mock_detect,
+        patch(
+            "endpoints.heartbeat.fs_platform_handler.get_platforms"
+        ) as mock_get_platforms,
+    ):
+        mock_detect.return_value = "struct_a"
+        mock_get_platforms.return_value = ["n64"]
+
+        response = client.get(
+            "/api/setup/library",
+            headers={"Authorization": f"Bearer {access_token}"},
+        )
+
+    assert response.status_code == status.HTTP_200_OK
+    data = response.json()
+
+    assert data["detected_structure"] == "struct_a"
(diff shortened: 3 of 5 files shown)


backend/handler/database/roms_handler.py
```diff
@@ -1642,6 +1642,19 @@ def get_rom_count(
...
+        self,
+        *,
+        session: Session = None,  # type: ignore
+    ) -> bool:
+        """Whether the library holds at least one ROM.
+
+        An existence probe rather than a count, for callers that only
+        need to know whether the library has ever been scanned.
+        """
+        return session.scalar(select(Rom.id).limit(1)) is not None
```
gantoine: you can use `get_roms_count` in `stats_handler.py` instead, its performance is good

gantoine (changes requested):
good approach in `heartbeat.py`

Spinnich:
Done in d095f87 — swapped to `db_stats_handler.get_roms_count() > 0` and removed
`has_any_rom()`, so `roms_handler.py` is no longer touched by this branch at all.

I did want to confirm your read on the performance before relying on it, since
`roms` is the one table where a whole-library query usually hurts. It holds up:
the count never touches the fat rows.

```
EXPLAIN SELECT count(*) FROM roms;
-> type=index   key=ix_roms_updated_at   Using index
```

That is an index-only scan of a 2.5 MB secondary index rather than the 963 MB
clustered index. Measured on a replica of my instance (92,800 ROMs, `roms` at
963 MB, `innodb_buffer_pool_size` 256 MB, MariaDB 11.4.12):

| | |
| --- | --- |
| Raw SQL, cold buffer pool (post-restart) | 25.0 ms |
| Raw SQL, warm (n=8) | 23.4 ms avg |
| `get_roms_count()` through the ORM (n=10) | 25.4 ms median |

Cold and warm match because a 2.5 MB index stays resident even in a 256 MB pool,
so none of the I/O-bound behaviour of the other whole-library queries applies
here. Against the 17-38 s filesystem walk this guard prevents, that is roughly
0.15% of the cheapest case, and post-PR it is only paid by callers that still
hit `/setup/library` on a populated library, since the v2 home page no longer
does.

`uv run pytest` remains green at 2704 passed, 2 skipped. The two new backend
tests drive real DB fixtures rather than mocking the handler, so they cover the
new call path without modification.

gantoine (approved):

Issue #4068, opened 2026-08-02
[Bug] "Random ROM" button in platform and collection galleries scans the entire library on every click


Spinnich:
**RomM version**

5.1.0. Affected code unchanged on `master` (79a832247).

**Describe the bug**

Clicking the "Random ROM" button (the shuffle icon in the header action row of a
platform or collection page, next to Download/Upload/Scan) can hang for many
seconds before it jumps to a game.

The button only needs one game, but every call to `/api/roms` also builds three
extra things alongside the results:

- the A–Z jump index for the alphabet strip,
- the list of available filter values (genres, regions, companies, ...),
- the complete ordered list of ROM ids that powers infinite scroll.

All three cover the **whole** result set regardless of how many games you asked
for, so a request for a single game does roughly the same work as loading the
entire gallery. The endpoint can turn them off (`with_char_index`,
`with_filter_values`, `with_rom_id_index`), but the gallery random buttons don't
pass those flags.

The gallery case is worse than it first looks. Results for the whole library get
cached, so one slow call at least makes the next fast. But once a request is
narrowed to a platform or a collection, caching is skipped by design and
everything is recomputed from scratch — so the gallery button pays full price on
**every single click**, with no warm-up.

The comment above the code describes it as "a cheap count-only fetch", which is
what it was meant to be but isn't.

**To Reproduce**

1. Use a large library (most visible in the tens of thousands of games).
2. Switch to the v2 UI.
3. Open a platform page (or a collection page).
4. In the header band, next to the Download / Upload / Scan buttons, click the
   circular shuffle button (tooltip: "Random ROM").
5. Wait. The delay repeats on every click rather than getting faster.

**Expected behavior**

Picking a random game should be near-instant. It needs one row and a count, so it
should skip the alphabet index, the filter list and the full ROM id index.

**Screenshots**

Not a visual bug, so measurements are more useful. Taken on an 83,000-game
instance with the database on SSD, comparing the same single-game request with and
without the three extras:

| Same `limit=1` request | Time | Response size |
| --- | --- | --- |
| extras **off** | **0.27 s** | 16.9 KB |
| extras **on** (cold) | **7.19 s** | 878 KB |

That's about 6.9 seconds of pure overhead and a 52x larger response, to return one
game.

On a second instance (23,698 games, database on a 5200 RPM laptop HDD) the same
comparison was 0.79 s vs 40.8 s, so this scales badly with slower storage.

**Additional context**

Affected code:

[...]

PR #4071, opened 2026-08-02
perf(roms): pick a random rom without paging to a random offset

### LINKED ISSUE #4066: [Bug] Random Pick widget takes ~6s on large libraries — random offset triggers deep pagination (follow-up to #4056)
**RomM version**

5.1.1-beta.1, and current `master`. Introduced by #4056 (merged 2026-08-01), which
was itself a fix for a worse version of this widget's behaviour.

**Describe the bug**

On a large library the Home page "Random Pick" widget takes several seconds to
show a game, on every home page visit.

The widget picks a game in two steps: ask the API for one game to learn how many
games exist, then ask for one game at a random position in that list. The second
request is the slow one. Asking a database for "the one row at position 41,500"
means it has to walk through and throw away the 41,499 rows before it — there's no
shortcut. The bigger the library, the deeper the average pick, and the slower it
gets.

This wasn't visible before #4056 because of an accident. The endpoint used to
return a full ordered list of every ROM id alongside the results, and when that
list was present the code picked the row straight out of it by id instead of
asking the database to skip ahead. That list was enormously expensive to build the
first time (it's what #4056 correctly removed), but once cached it was
inadvertently acting as a fast index for exactly this kind of deep lookup.

Removing it was the right call — it was doing whole-library work and returning
a ~878 KB response to deliver one game. But the underlying approach of "fetch the
Nth row at a random N" was only ever fast because that index was papering over it.
Random selection needs a different mechanism than pagination.

**To Reproduce**

1. Use a large library (measured on 83,000 games).
2. Load the Home page with the Random Pick widget enabled.
3. Note the delay before a game appears. Reroll and note it recurs every time.

Directly, using the same parameters the widget now sends:

```
GET /api/roms?limit=1&offset=N&order_by=&order_dir=asc&group_by_meta_id=false \
    &with_char_index=false&with_filter_values=false&with_rom_id_index=false
```

varying `N`.

**Expected behavior**

Picking a random game should take roughly the same time regardless of library
size. It shouldn't get slower the more games you own.

**Screenshots**

83,000-game library, database on SSD, single-row request, varying the offset:

| offset | time |
| --- | --- |
| 0 | 0.27 s |
| 20,000 | 3.00 s |
| 40,000 | 5.34 s |
| 60,000 | 12.47 s |
| 80,000 | 15.75 s |

A random pick across 83,000 games averages an offset around 41,500, which lands
near 5.5 s — matching the ~6 s observed in the browser.

Net effect of #4056 on this widget:

| Scenario | Before #4056 | After #4056 |
| --- | --- | --- |
| First home load (cold cache) | ~7.4 s | ~6.3 s |
| Every later home load | **~0.4 s** | **~6.3 s** |
| Response size | 878 KB | 17 KB |
| Whole-library scan per load | yes | no |

So the change fixed the cold load, the payload and the whole-library scan, but made
the common case (a repeat visit, warm cache) substantially slower. This will be
worse on instances with slower storage.

**Additional context**

Affected code:

- `frontend/src/v2/components/Home/Widgets/RandomPickWidget.vue` — `reroll()` makes
  two `getRoms` calls, the second at `offset: randomOffset`
- `backend/endpoints/roms/__init__.py` — with `with_rom_id_index=false` the request
  takes the `query.offset(...).limit(...)` branch. The other branch slices the
  cached id list and fetches by primary key, which is why this was previously fast:

  ```python
  page_ids = rom_id_index[params.offset : params.offset + params.limit]
  page_rows = session.scalars(query.where(Rom.id.in_(page_ids))).all()
  ```

This is a follow-up rather than a revert — the payload and whole-library-scan wins
from #4056 are worth keeping. The remaining problem is that random selection is
implemented as pagination.

Possible fixes, roughly in order of how well they'd scale:

- A dedicated `/api/roms/random` endpoint that samples on the primary key: choose a
  random id within the min/max range, take the first row at or above it, retry on a
  miss. Constant time regardless of library size, and it removes the two-request
  dance entirely.
- Have the client pick from the existing cheap `/roms/identifiers` id list and then
  request that single id directly.
- `ORDER BY RAND() LIMIT 1` is the obvious one-liner but sorts the whole table, so
  it would likely be worse than the current behaviour. Not recommended.

The same two-call random-offset pattern is used by the platform and collection
gallery "Random ROM" buttons, so whatever replaces it should cover those too. Those
have a separate, compounding problem — they're scoped, so they never hit the cache
at all — filed separately as issue #4068.

Spinnich:
**Description**

Fixes #4066.

The v2 Home "Random Pick" widget resolved a pick by paging to it: one request to
learn the library total, a second at `offset: <random>`. An offset makes the
database walk and discard every preceding row, so the cost grew with the library.
On an 83k-rom instance a pick took **1.7s median and up to 6.9s**, and the widget
runs one on every Home mount.

This adds `GET /api/roms/random`, which samples on the primary key instead.

Two mechanisms, both uniform:

1. **Batch sampling (the fast path).** Sixteen random primary keys are offered to
   the filtered query in a single `IN (...)`. Every id in the table is equally
   likely to be offered, so any hit is an unbiased pick, and the batch costs one
   index lookup per candidate no matter how large the library is. Bounds come
   from `MIN`/`MAX` over the untouched primary key (two index seeks) rather than
   over the filtered set, which would be a scan of it.
2. **Uniform offset fallback.** A batch misses when the query matches too little
   of the id space (a narrow scope, or an id range left sparse by deletions). It
   then counts the set and takes the row at a random position. The statement
   selects nothing but the id and carries **no `ORDER BY`**, so it reads index
   entries rather than rows.

Measured on a real 83,132-rom library (read-only, production data):

| | before | after |
| --- | --- | --- |
| unscoped pick, median | 1.70 s | **2.9 ms** |
| unscoped pick, worst of 30 | 6.90 s | 5.7 ms |
| fallback path (0.58% of unscoped picks) | - | 236 ms |
| scoped to a large platform | 9.20 s | 19-25 ms |

Distribution over 200k draws against that same id space: 75,495 of 83,132 roms
drawn (exactly what uniform predicts for that many draws), stdev 1.55 against a
Poisson expectation of 1.55.

The endpoint accepts the platform/collection/virtual/smart scope params, so the
gallery "Random ROM" buttons tracked in #4068 can move onto it without a
signature change. Rewiring those call sites is deliberately left out of this PR.

**Files modified**

- `backend/handler/database/roms_handler.py` - new `get_random_rom_id(query)`
  implementing both mechanisms, plus the `RANDOM_ID_SAMPLE_SIZE` constant. Uses
  `secrets` rather than `random`.
- `backend/endpoints/roms/__init__.py` - new `GET /random` route returning
  `SimpleRomSchema | None`. Declared before `/{id}` so the int path param doesn't
  swallow it. Scope params and permission filtering mirror `GET /api/roms`, and
  the picked row is re-checked against the caller's permissions after the fetch
  (see reviewer notes).
- `backend/tests/endpoints/roms/test_random.py` - new, 11 tests covering both
  mechanisms (see testing notes).
- `frontend/src/services/api/rom.ts` - new `getRandomRom()` wrapper.
- `frontend/src/v2/components/Home/Widgets/RandomPickWidget.vue` - `reroll()` now
  makes one request. Drops `pickOnce()`, the `PICK_QUERY` opt-outs and the
  retry-on-drift branch, all of which existed only to work around the two-call
  approach.
- `frontend/src/v2/components/Home/Widgets/RandomPickWidget.test.ts` - new, 4
  tests.
- `docs/BACKEND_ARCHITECTURE.md` - endpoint table entry.

No migration, no schema change, and no regenerated types: the route reuses
`SimpleRomSchema`.

**Testing notes**

- Full backend suite: **2715 passed, 2 skipped**. The new file is 11 tests,
  covering auth, an empty library, platform/collection scoping, admin-hidden roms
  staying out of the pick, post-fetch visibility, and both code paths reaching
  every rom in the set. Two mechanism guards: one asserts `get_rom_count` and
  `get_rom_id_index` are never called and that no statement carries an `OFFSET`;
  the other patches the sample size to 0 to force the fallback.
- Frontend: 4 new tests (single request per pick, empty-library copy, error copy
  on first load, reroll keeping the previous pick on failure). Full vitest run
  green.
- Verified in the browser in both themes: pick renders, reroll reshuffles in
  place without navigating, keyboard focus survives the reroll.
- Timings and the distribution check above were run against a production
  database over a read-only account, `SELECT`/`EXPLAIN` only.

**What reviewers should look at**

- **Uniformity of the batch path.** The argument is exchangeability: the sampling
  procedure treats every existing id identically, so conditional on a non-empty
  hit set each is equally likely to be returned. Worth a second opinion.
- **Scoped picks mostly take the fallback.** Sampled keys rarely land inside a
  narrow scope, so #4068's call sites will pay the count-plus-offset path. That
  measured 19-25 ms on large platforms, but it is not the constant-time path and
  shouldn't be described as one.
- **The fallback deliberately has no `ORDER BY`.** Adding `ORDER BY roms.id` back
  looks harmless and is not: a PK-ordered scan *is* the clustered index, so it
  drags whole rows off disk. That measured 4.6 s against 0.067 s for the same
  query without it. Uniformity doesn't depend on the ordering, only on each row
  appearing exactly once.
- **`RANDOM_ID_SAMPLE_SIZE = 16`** is tuned for an id space roughly a quarter
  occupied, which is what deletions leave behind on a long-lived instance
  (mine is 27% dense). Happy to change it if that assumption seems wrong.
- **Post-fetch visibility re-check.** The id is chosen by a permission-filtered
  query but the row is fetched by raw id, so the loaded row is re-tested with
  `perms.can_see_rom(...)`, matching what every other raw-id ROM lookup does via
  `assert_rom_visible`. It returns `null` rather than raising 404, to keep a
  hidden ROM indistinguishable from an empty scope and to keep an error path out
  of a contract that otherwise has none.
- **Race on delete.** If a ROM is deleted between the id pick and the fetch, the
  endpoint returns `null` and the widget briefly shows its empty-library copy.
  I judged a third round trip not worth avoiding that.

**Checklist**

**AI assistance**

This PR was written primarily by Claude Code, including this description and the
replies on the review thread. I directed the work, required that the approach be
measured against my own production database before it was accepted, and reviewed
the result.

Two things it got wrong that the process caught, noted so reviewers can calibrate
how much scrutiny to apply. The first implementation sampled a random point in
the id range and took the nearest row at or above it, which is only uniform on a
dense id space; on my 27%-dense library it skewed 10,840x toward one ROM and
could reach only 2,773 of 83,132 ROMs. It was replaced with the batch-sampling
mechanism above and re-validated the same way. Separately, the first version of
the visibility regression test passed with the fix reverted, because the query
returned empty before control flow ever reached the guard; it was rewritten to
actually exercise it.

### DIFF
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -898,6 +898,62 @@ def get_rom_identifiers(
     return [r.id for r in db_roms]
 
 
+@protected_route(router.get, "/random", [Scope.ROMS_READ])
+def get_random_rom(
+    request: Request,
+    platform_ids: Annotated[
+        list[int] | None,
+        Query(
+            description=(
+                "Platform internal ids. Multiple values are allowed by repeating the"
+                " parameter, and the pick will match any of the values."
+            ),
+        ),
+    ] = None,
+    collection_id: Annotated[
+        int | None,
+        Query(description="Collection internal id.", ge=1),
+    ] = None,
+    virtual_collection_id: Annotated[
+        str | None,
+        Query(description="Virtual collection internal id."),
+    ] = None,
+    smart_collection_id: Annotated[
+        int | None,
+        Query(description="Smart collection internal id.", ge=1),
+    ] = None,
+) -> SimpleRomSchema | None:
+    """Retrieve one rom picked at random, or null when the scope holds none.
+
+    Sampled on the primary key instead of paged to, so the pick doesn't get
+    slower as the library grows.
+    """
+    perms = get_permissions(request)
+
+    base_query, _ = db_rom_handler.get_roms_query(user_id=request.user.id)
+    query = db_rom_handler.filter_roms(
+        query=base_query,
+        user_id=request.user.id,
+        hidden_platform_ids=perms.hidden_platform_ids,  # type: ignore
+        hidden_rom_ids=perms.hidden_rom_ids,  # type: ignore
+        platform_ids=platform_ids,
+        collection_id=collection_id,
+        virtual_collection_id=virtual_collection_id,
+        smart_collection_id=smart_collection_id,
+        include_related=False,
+    )
+
+    rom_id = db_rom_handler.get_random_rom_id(query=query)
+    if rom_id is None:
+        return None
+
+    rom = db_rom_handler.get_rom_simple(rom_id)
+    if not rom:
+        return None
+
+    return SimpleRomSchema.from_orm_with_request(rom, request)
+
+
 @protected_route(
     router.get,
     "/download",
--- a/backend/endpoints/roms/__init__.py
+++ b/backend/endpoints/roms/__init__.py
@@ -951,6 +951,13 @@ def get_random_rom(
     if not rom:
         return None
 
+    # The fetch is by raw id, so it re-checks the row it actually loaded rather
+    # than trusting the filter that chose the id: a rom that moved to a hidden
+    # platform in between was picked under its old one. Reads no database, and
+    # null keeps a hidden rom indistinguishable from an empty scope.
+    if not perms.can_see_rom(rom.id, rom.platform_id):
+        return None
+
     return SimpleRomSchema.from_orm_with_request(rom, request)
 
 
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -2,6 +2,7 @@ import functools
 import hashlib
 import json
 import re
+import secrets
 from collections.abc import Iterable, Sequence
 from datetime import datetime
 from typing import Any, NamedTuple
@@ -139,6 +140,11 @@ FULLTEXT_MIN_TOKEN_SIZE = 3
 # length, so an ordinary name search builds no hash SQL at all. Hashes are
 # stored lowercase, which keeps the lookup an indexed equality.
 HEX_DIGEST_REGEX = re.compile(r"[0-9a-fA-F]+")
+
+# Primary keys offered per round trip when picking a random rom. Sixteen
+# lands a hit ~99% of the time on a library occupying a quarter of its id
+# range, which is what deletions leave behind on a long-lived instance.
+RANDOM_ID_SAMPLE_SIZE = 16
 
 # CRC32 (8), MD5 and RetroAchievements (32), SHA-1 (40).
 ROM_HASH_COLUMNS_BY_DIGEST_LENGTH: dict[int, tuple[QueryableAttribute, ...]] = {
@@ -1641,6 +1647,51 @@ class DBRomsHandler(DBBaseHandler):
             )
             or 0
         )
+
+    @begin_session
+    def get_random_rom_id(
+        self,
+        query: Query,
+        *,
+        session: Session = None,  # type: ignore
+    ) -> int | None:
+        """Pick one rom id at random, uniformly, from a filtered query.
+
+        Two mechanisms, both uniform. First a batch of random primary keys is
+        offered to the query: every id in the table is equally likely to be
+        offered, so any hit is an unbiased pick, and the whole batch costs one
+        index lookup per candidate no matter how large the library is.
+
+        A batch misses when the query matches too little of the id space (a
+        scoped gallery, an id range left full of gaps by deletions). It then
+        falls back to counting the set and taking the row at a random position
+        in it. That reads no rows, only index entries, since the statement
+        selects nothing but the id.
+        """
+        id_query = query.order_by(None).with_only_columns(Rom.id)  # type: ignore
+
+        # Bounds come from the table rather than the filtered set: they only
+        # need to cover it, and MIN/MAX over an untouched primary key are two
+        # index seeks, where the same pair over a joined and filtered set is a
+        # scan of it.
+        lowest, highest = session.execute(
+            select(func.min(Rom.id), func.max(Rom.id))
+        ).one()
+        if lowest is None or highest is None:
+            return None
+
+        span = highest - lowest + 1
+        candidates = {
+            lowest + secrets.randbelow(span) for _ in range(RANDOM_ID_SAMPLE_SIZE)
+        }
+        hits = session.scalars(id_query.where(Rom.id.in_(candidates))).all()
+        if hits:
+            return secrets.choice(hits)
+
+        total = self.get_rom_count(query=query, session=session)
+        if total == 0:
+            return None
+        return session.scalar(id_query.limit(1).offset(secrets.randbelow(total)))
 
     @begin_session
     def get_roms_by_fs_name(
--- a/frontend/src/services/api/rom.ts
+++ b/frontend/src/services/api/rom.ts
@@ -415,6 +415,31 @@ async function getRomSimple({
   return api.get<SimpleRom>(`/roms/${romId}/simple`, { signal });
 }
 
+async function getRandomRom({
+  platformIds = null,
+  collectionId = null,
+  virtualCollectionId = null,
+  smartCollectionId = null,
+}: {
+  platformIds?: number[] | null;
+  collectionId?: number | null;
+  virtualCollectionId?: string | null;
+  smartCollectionId?: number | null;
+} = {}) {
+  // `/roms/random` samples the primary key server-side, so one request
+  // resolves a pick regardless of library size. `null` means the scope
+  // holds no roms.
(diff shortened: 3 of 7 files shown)


Spinnich:
Thanks @greptileai, the second half of this is a real gap. Splitting it because only one of the two mechanisms holds:

**The permission-change race isn't reachable.** `get_permissions()` caches the resolved permissions on `request.state` (`handler/auth/dependencies.py:35-42`), so the filtered ID selection and any post-fetch check read the same snapshot for the life of the request. A hide applied concurrently is invisible to both, and re-checking after the fetch closes nothing there.

**The platform-move case is real.** `can_see_rom(rom.id, rom.platform_id)` tests `platform_id` from the row that was actually loaded, whereas `filter_roms` matched on its value at SELECT time. A ROM moved onto a hidden platform between the two statements would be selected under its old platform and serialized under its new one.

The stronger argument for fixing it is consistency rather than the race: every other endpoint that resolves a ROM by raw ID calls `assert_rom_visible`, and this one was the exception, so its safety was only provable by tracing back to the filter several lines up. The check costs one in-memory set lookup and no query, so there's no reason not to have it.

Fixed in 7bbf94c08. I used `perms.can_see_rom(...)` returning `null` rather than `assert_rom_visible`, deliberately: the endpoint's contract is `SimpleRomSchema | None` with no error path, and returning `null` keeps a hidden ROM indistinguishable from an empty scope, which is the same masking `assert_rom_visible`'s 404-instead-of-403 exists to provide.

Added `test_get_random_rom_rechecks_visibility_after_fetching`, confirmed to fail with the guard reverted. Worth noting the first version of that test passed either way: the viewer's only ROM was the hidden one, so the filtered query came back empty and the endpoint returned `null` before ever reaching the fetch. It needed a second, visible ROM so control flow actually reaches the guard.

*This reply was drafted by Claude Code, as was the PR itself.*

frontend/src/services/api/rom.ts
```diff
@@ -415,6 +415,31 @@ async function getRomSimple({
...
+  smartCollectionId = null,
+}: {
+  platformIds?: number[] | null;
+  collectionId?: number | null;
+  virtualCollectionId?: string | null;
+  smartCollectionId?: number | null;
+} = {}) {
+  // `/roms/random` samples the primary key server-side, so one request
+  // resolves a pick regardless of library size. `null` means the scope
+  // holds no roms.
```
gantoine: not useful

backend/handler/database/roms_handler.py
```diff
@@ -1642,6 +1648,51 @@ def get_rom_count(
...
+            lowest + secrets.randbelow(span) for _ in range(RANDOM_ID_SAMPLE_SIZE)
+        }
+        hits = session.scalars(id_query.where(Rom.id.in_(candidates))).all()
+        if hits:
+            return secrets.choice(hits)
+
+        total = self.get_rom_count(query=query, session=session)
+        if total == 0:
+            return None
+        return session.scalar(id_query.limit(1).offset(secrets.randbelow(total)))
```
gantoine, line 1694: this is neat, go claude!

backend/endpoints/roms/__init__.py
```diff
@@ -898,6 +898,62 @@ def get_rom_identifiers(
...
+        collection_id=collection_id,
+        virtual_collection_id=virtual_collection_id,
+        smart_collection_id=smart_collection_id,
+        include_related=False,
+    )
+
+    rom_id = db_rom_handler.get_random_rom_id(query=query)
+    if rom_id is None:
+        return None
+
```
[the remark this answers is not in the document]
gantoine: if/when you address this pls resolve the comment

gantoine (changes requested):

frontend/src/services/api/rom.ts
Spinnich: Agreed, dropped in b6d43ce73. It restated the endpoint, and the backend docstring already carries the sampling rationale.

*This reply was drafted by Claude Code, as was the PR itself.*

backend/endpoints/roms/__init__.py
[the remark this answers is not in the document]
Spinnich: Addressed in 7bbf94c08 (the fetched row is re-tested with `perms.can_see_rom`, plus a regression test confirmed to fail with the guard reverted). Resolving.

*This reply was drafted by Claude Code, as was the PR itself.*

gantoine (approved):

PR #4077, opened 2026-08-03
feat(firmware): let missing firmware be found, cleaned up, and kept out of the player

### LINKED ISSUE #4075: [Bug] Firmware/BIOS entries whose files are gone can't be cleaned up, and are still offered to the emulator
**RomM version**
5.1.0 (also present on current `master`, e7c236252)

**Describe the bug**
When a ROM file disappears from disk, RomM flags it and gives you a place to deal with
it: **Settings → Library Management → Missing**, which lists every missing ROM across the
whole library, lets you filter by platform, and has a "Cleanup all" button that deletes
the stale database rows in one go.

Firmware (BIOS) files get flagged the same way by a scan, but there is no equivalent
place to clean them up. The only way to find a stale BIOS entry is to open each
platform's **Firmware** tab one at a time and look for rows drawn with a strikethrough,
then delete them by hand. On a library with dozens of platforms, that's not a realistic
way to keep things tidy, and there's nothing that tells you a cleanup is even needed.

Two things go wrong as a result of those stale entries sticking around:

1. **The emulator still offers them.** The BIOS dropdown in the browser player lists
   every firmware entry for the platform, including ones whose file no longer exists.
   Worse, when a platform has exactly one firmware entry, RomM auto-selects it (added in
   #3150). If that one entry is stale, the game just fails to load with no explanation
   pointing at the BIOS.

2. **Firmware counts are wrong.** The firmware count shown for a platform includes
   entries whose files are gone, so a platform with no usable BIOS at all can still
   claim to have several.

**To Reproduce**

1. Put a BIOS file in the firmware folder for a platform (e.g. `bios/ps/scph5501.bin`,
   or `ps/bios/scph5501.bin` depending on your library layout) and run a scan. It shows
   up under the platform's **Firmware** tab.
2. Delete or move that file on disk, then run a scan again. The entry is now flagged as
   missing — the row is struck through in the Firmware tab.
3. Go to **Settings → Library Management**. There are only three tabs — Folder mapping,
   Excluded, and Missing — and the Missing tab covers ROMs only. There is no missing-
   firmware view and no cleanup action anywhere.
4. Open a game for that platform in the browser player and look at the BIOS selector in
   the Setup panel. The deleted file is still listed and selectable. If it was the only
   firmware entry for the platform, it is pre-selected, and the game fails to start.

**Expected behavior**

Stale firmware entries should be as easy to clean up as stale ROM entries. Concretely:

- A way to see all missing firmware across the library in one place, and remove it in
  bulk — either a firmware section in the existing Missing tab, or a "Cleanup missing
  firmware" task alongside the existing "Cleanup missing ROMs" one.
- Firmware whose file is missing should not be offered as a BIOS choice in the player,
  and definitely should not be auto-selected. If it is shown, it should be clearly
  marked and unselectable.
- Platform firmware counts should reflect firmware that actually exists on disk.

**Additional context**

The database flag already exists and is set correctly — this is only about there being
no way to act on it. For anyone digging in:

- `mark_missing_firmware` in `backend/handler/database/firmware_handler.py` sets the
  flag during a scan, mirroring `mark_missing_roms`.
- `cleanup_missing_roms` in `backend/tasks/manual/` is the ROM-side cleanup task; there
  is no firmware counterpart registered in `backend/endpoints/tasks.py`.
- `GET /api/firmware` returns every entry with no way to filter out or identify missing
  ones server-side, which is why both the player's BIOS list and the platform firmware
  count include them.

This affects both the classic and the new UI. Neither has a bulk cleanup path; the new
UI at least marks missing entries in the platform Firmware tab.

Spinnich:
**Description**

Fixes #4075

A scan flags firmware whose file has vanished with `missing_from_fs`, but nothing acted on that flag. The result was three separate problems:

- Stale firmware could only be found by opening each platform's Firmware tab one at a time, and there was no bulk removal — unlike ROMs, which have had a "Cleanup missing ROMs" task for a long time.
- The player offered missing firmware as a BIOS choice and, when a platform had exactly one entry, **auto-selected** it (the behaviour added in #3150). Games then failed to boot with nothing on screen pointing at the BIOS as the cause.
- `firmware_count` on `PlatformSchema` counted rows, so a platform whose only BIOS file had been deleted still advertised firmware it could not serve.

This PR makes the flag actionable end to end:

1. **`GET /api/firmware` gains a `?missing=` filter** (`true` / `false` / omitted for everything), so callers can select on the flag instead of taking the whole set and filtering client-side.
2. **New `cleanup_missing_firmware` manual task**, modelled on `cleanup_missing_roms` and registered next to it. Optionally scoped to one `platform_id`. It only deletes database rows — the files are already gone, so nothing is removed from disk.
3. **New "Missing firmware" tab** under Settings → Library Management, listing every flagged entry library-wide with a platform multi-select and a "Clean up all" action wired to the task.
4. **The v2 player no longer offers missing firmware.** It fetches with `missing: false`, and BIOS selection moved into a pure util so the "sole entry auto-select" rule can only ever fire on an entry that actually exists.
5. **`firmware_count` counts only firmware present on disk**, so the platform stat matches what is usable.

**Files changed**

_Backend_

| File | Change |
| --- | --- |
| `backend/handler/database/firmware_handler.py` | `list_firmware()` gains a `missing: bool \| None` filter; stacks with the existing `platform_id` and hidden-platform filters |
| `backend/endpoints/firmware.py` | `?missing=` query param on `GET /api/firmware`, forwarded to the handler |
| `backend/endpoints/responses/firmware.py` | `FirmwareSchema` gains `platform_id` so a library-wide list can group by platform without a round trip per row |
| `backend/endpoints/responses/platform.py` | `firmware_count` excludes entries flagged `missing_from_fs` |
| `backend/tasks/manual/cleanup_missing_firmware.py` | **New.** `CleanupMissingFirmwareTask` + stats dataclass, mirroring the ROM cleanup |
| `backend/endpoints/tasks.py` | Registers `cleanup_missing_firmware` in `manual_tasks` |

_Frontend_

| File | Change |
| --- | --- |
| `frontend/src/v2/components/Settings/MissingFirmwareSection.vue` | **New.** The Missing firmware tab |
| `frontend/src/v2/views/Settings/LibraryManagement.vue` | Adds the `missing-firmware` tab to the `Tab` union, nav, and body |
| `frontend/src/v2/utils/playerFirmware.ts` | **New.** `resolveInitialFirmware()` — storage → core config → sole option, all restricted to firmware that exists |
| `frontend/src/v2/views/Player/EmulatorJS.vue` | Fetches firmware with `missing: false` and delegates initial selection to the util |
| `frontend/src/v2/components/Gallery/FirmwareTab.vue` | The locally patched `firmware_count` now matches what the server derives |
| `frontend/src/services/api/firmware.ts` | `getFirmware()` forwards an optional `missing` param |
| `frontend/src/__generated__/models/FirmwareSchema.ts` | Regenerated (`platform_id`) |
| `frontend/src/locales/*/settings.json` | 6 new keys, translated across all 18 locales |

**Testing notes**

Automated:

- Backend: **2723 passed, 2 skipped.** New coverage — 5 handler tests for the `missing` filter (including that it stacks with `platform_id` and with hidden platforms), 6 endpoint tests for `GET /api/firmware`, 5 task tests (deletes only flagged rows, scopes to one platform, counts delete failures, no-op stats), plus `firmware_count` on the platform endpoint and a registry test asserting the task is actually reachable by name.
- Frontend: **630 passed / 49 files.** 8 unit tests for `resolveInitialFirmware` (the key one: a sole entry whose file is gone resolves to `null` rather than being auto-selected) and 7 component tests for the new tab.
- `npm run typecheck`, `trunk fmt && trunk check`, `check_i18n_locales.py`, `check_i18n_sorted.py` all clean.

Manual, against a dev library with firmware rows flagged missing:

- Settings → Library Management → Missing firmware lists the flagged entries with platform, path and size; the platform select only offers platforms that actually have something to clean up.
- "Clean up all" with no platform selected runs unscoped; with exactly one platform selected it forwards `platform_id`; the list refreshes and empties.
- A platform whose only BIOS was deleted from disk: the player's BIOS dropdown is now empty and nothing is preselected, and the platform's firmware stat no longer counts it.
- Verified in both light and dark themes.

**Things worth a reviewer's attention**

1. **`firmware_count` is a behaviour change to an existing field.** Platforms holding stale firmware rows will report a lower count than before. That is what the issue asks for, but it is user-visible.
2. **`FirmwareSchema.platform_id` is an additive API field** — the regenerated types are included in this PR.
3. **The cleanup confirmation has no type-to-confirm gate**, unlike the ROM cleanup's `requireTyped: "DELETE"`. Per the v2 patterns guide, type-to-confirm is required when an action affects the filesystem; this one only drops rows pointing at files that are already gone, whereas the ROM cleanup additionally removes resource directories from disk. Happy to add the gate if you'd rather the two cleanups look identical.
4. **v1's player still offers missing firmware.** Its fetch lives in `frontend/src/views/Player/EmulatorJS/Base.vue`, a frozen v1 path that `check-v1-frozen` blocks, so it is untouched here. It would need the `allow-v1-changes` label as a follow-up if you want it fixed before v1 is retired.
5. **The task does not filter hidden platforms**, matching `cleanup_missing_roms`; both are gated on `Scope.TASKS_RUN`.
6. **The tab loads the whole missing set unpaginated** and filters by platform in memory. Firmware tables are dozens of rows, so this stays well inside a single request, and it lets the multi-select work against an API that only accepts one `platform_id`. There is no index on `missing_from_fs` — deliberately, at this table size.

**Checklist**

**AI assistance disclosure**

This PR was written primarily by Claude Code (Opus 5), under my direction and review: issue analysis, tests, implementation, and translations. I reviewed the diff, and this PR description was also AI-generated and edited by me.

image

image


Spinnich:
Valid catch @greptileai, fixed in 872afb243.

The refresh was on a fixed 1.5s delay, but `POST /tasks/run/{name}` returns as
soon as the job is enqueued, so any worker slower than that left the tab showing
rows that were already deleted, behind a success toast.

Replaced the delay with a new `useTaskCompletion` composable that polls the job's
own status via `GET /api/tasks/{task_id}` until it reaches a terminal state
(`finished` / `failed` / `stopped` / `canceled`), and refreshes off that:

- backs off 400ms → 5s between polls, caps the total wait at 5 minutes, and
  refreshes anyway on timeout
- treats an unfetchable job as done, since a 404 means it ran and aged out of
  Redis past `TASK_RESULT_TTL`
- cancels on scope dispose, so leaving the tab mid-wait can't refresh or raise an
  error over an unrelated page

Polling rather than sockets is deliberate: there's no task-completion socket
event, and `TasksSection` already polls task status on an interval, so this is
the same mechanism scoped to a single job.

Worth noting the missing **ROMs** cleanup in `MissingGamesSection` had the
identical race on `master` — I'd modelled this tab on it — so the fix is applied
there too.

Added 7 unit tests for the composable and 2 component tests (refreshes once the
task reports finished; leaves the list alone while it's still `started`).
Frontend suite 641 passed / 50 files, typecheck and trunk clean.

Spinnich:
Good catch on point 1 @greptileai — fixed in c59422280.

It also surfaced a mismatch between the comment and the code: the comment said a
job past its result TTL 404s, but the `catch` treated *every* failure as
completion. So a timeout or a 5xx mid-cleanup would resolve `true` and refresh
against a job still running — the same stale-list symptom the composable exists
to prevent, just behind a rarer trigger.

Now narrowed to a genuine 404 via `axios.isAxiosError(err) && err.response?.status === 404`,
matching how the rest of the codebase discriminates errors. Anything else falls
through to the normal backoff, and `POLL_TIMEOUT_MS` still bounds the wait, so a
persistently failing lookup degrades to "refresh anyway" rather than hanging.

Two tests pin it: a 503 followed by `finished` must produce **two** lookups (the
previous code stopped after one, so it fails against 872afb243), and a
persistently failing lookup must still settle at the deadline.

Points 2 and 3 need no change — the immediate first poll is already asserted by
the "resolves without waiting when the job already finished" test, and the
`MissingGamesSection` fix landed in 872afb243.

One correction to the read on the generation counter: it's right now, but only
because the first cut of that mechanism had a bug — `cancel()` cleared the
pending timer, which orphaned the promise so a cancelled wait never resolved.
The scope-dispose and supersession tests caught it before it was pushed.

Frontend suite 643 passed / 50 files, typecheck and trunk clean.

gantoine (approved):

PR #4078, opened 2026-08-03
perf(roms): sort metadata fields on the indexed roms column

### LINKED ISSUE #4067: [Bug] Sorting a gallery by release date (or other metadata fields) is drastically slower than other sorts
**RomM version**

5.1.0. Affected code unchanged on `master` (79a832247).

**Describe the bug**

Sorting a large gallery by a metadata field — release date is the clearest example
— is dramatically slower than sorting by the default. The difference isn't
marginal; on a large library it's the difference between instant and a 20 second
wait, and it's paid on every page of results.

Most sortable fields live on the ROM record itself. Metadata fields like release
date live in a separate table, so sorting by one makes RomM join that table and
sort the entire library through it. The column has no database index, so there's
nothing to make that sort cheap and the database orders every row from scratch each
time.

**To Reproduce**

1. Use a large library (this was measured on 83,000 games).
2. Open a gallery and set the sort to release date.
3. Compare load time against the default sort.

Directly, with the extra index-building disabled so only the sort is being
measured:

```
GET /api/roms?limit=1&offset=0&with_char_index=false&with_filter_values=false&with_rom_id_index=false
```

once with `order_by=` (default) and once with `order_by=first_release_date`.

**Expected behavior**

Sorting by release date should be comparable to other sorts. Fetching a single row
should not require ordering the whole library through an unindexed join.

**Screenshots**

83,000-game instance, database on SSD, identical single-row request:

| Sort | Time |
| --- | --- |
| default ordering | **0.27 s** |
| `order_by=first_release_date` | **20.3 s** |

Roughly 76x slower for the same one row.

A second measurement isolates the cause. There's a separate code path that doesn't
run a row count at all, and the gap persists there too:

| Sort (index-building path, no count involved) | Time |
| --- | --- |
| default ordering | **7.19 s** |
| `order_by=first_release_date` | **19.45 s** |

Since that path never counts rows, the extra ~12 seconds is attributable to the
join and sort themselves rather than to counting.

**Additional context**

Affected code:

- `backend/handler/database/roms_handler.py` — `get_roms_query()` resolves an
  `order_by` that isn't a `Rom` attribute against `RomMetadata` and adds
  `outerjoin(RomMetadata, RomMetadata.rom_id == Rom.id)`
- `backend/models/rom.py` — `RomMetadata.first_release_date` is a plain
  `BigInteger` column with no index; I found no index on it in any model
  definition or migration

Likely fix is an index on the metadata columns that are actually offered as sort
options. Worth checking the other metadata-backed sorts at the same time, since
they go through the same path and probably share the problem — I only measured
release date.

Sorting by `RomUser` fields goes through a similar branch in the same function and
may be worth a look too.

Spinnich:
**Description**

Fixes #4067

Sorting a gallery by `first_release_date` took **20.3s** against **0.27s** for the
default sort on an 83k-game library. The issue guessed the cause was a missing
index, but the indexes were already there (`idx_roms_generated_first_release_date`
and friends, created by migration `0098`). The real cause is *where the sort key
sits*.

`roms_metadata` is not a table. It is a thin view over the STORED generated
columns on `roms`. Resolving one of its columns as the sort key did:

```python
order_attr = getattr(RomMetadata, order_by)
query = query.outerjoin(RomMetadata, RomMetadata.rom_id == Rom.id)
```

That joins `roms` back to itself and leaves the sort key on the *joined* table.
MariaDB/MySQL can only use an index for `ORDER BY` when the sort column belongs
to the **first** table in the join order, and the right side of a `LEFT JOIN`
never can be. So the planner fell back to a full scan of `roms` plus a filesort
of the entire library, on every page, to return 72 rows.

The three view columns that are actually offered as sorts now resolve straight
to the indexed `roms` column and the join disappears. Measured on a 120k-row
copy of the real `roms` schema (169 MB), with `Rom.metadatum`'s `lazy="joined"`
eager load present in **both** shapes so the comparison is honest:

| Sort | Before | After |
| --- | --- | --- |
| `first_release_date` ASC, LIMIT 72 | 6.02s | 0.004s |
| `average_rating` DESC, LIMIT 72 | 2.16s | 0.002s |

`EXPLAIN` before: `type: ALL`, 113,534 rows, `Using temporary; Using filesort`.
After: `type: index`, key `idx_roms_generated_first_release_date`, 72 rows, no
filesort — the same plan shape as the default `name_sort_key` sort.

No migration, no new index, no schema change: the columns and their indexes
already existed, they just weren't being read.

**Scope note:** the remaining `roms_metadata` columns (`genres`, `franchises`,
`collections`, `companies`, `game_modes`, `age_ratings`) are JSON arrays with no
index to sort on, and nothing in the UI offers them as a sort. Since `/api/roms`
has no `order_by` allowlist and accepts any string, the view-join fallback is
deliberately kept so those keep behaving exactly as they do today.

**Files changed**

| File | What |
| --- | --- |
| `backend/handler/database/roms_handler.py` | New `ROM_METADATA_ORDER_COLUMNS` map, plus a branch in `get_roms_query()` that resolves those three keys to the indexed `roms` column instead of joining the view. |
| `backend/models/rom.py` | Maps `generated_first_release_date`, `generated_average_rating` and `generated_player_count` on `Rom` as read-only (`FetchedValue()`), so the sort has an attribute to reference. |
| `backend/tests/handler/database/test_roms_metadata_sort.py` | New. 11 cases covering both the emitted SQL and the resulting order. |

**Testing notes**

- Full backend suite: 2715 passed, 2 skipped.
- End-to-end against a running app: all six sort paths return 200 with correct
  ordering (`first_release_date`, `average_rating`, `player_count`, each
  direction).
- Perf numbers above measured on a 120k-row seeded `roms` table, since a normal
  dev library is too small to show the difference.

**What a reviewer should look at**

1. **The read-only mapping.** `generated_*` are STORED generated columns owned by
   the engine, mapped with `server_default`/`server_onupdate=FetchedValue()` so
   SQLAlchemy never tries to write them. Worth confirming that reasoning holds:
   `add_rom` goes through `session.merge()`, `update_rom` uses a Core `update()`
   with caller-supplied keys, and `_nullable_columns()` is only ever called with
   `RomFile`/`TrackMeta`, so nothing enumerates `Rom`'s columns to build a write.
2. **No API surface change.** The response schemas in `endpoints/responses/rom.py`
   declare their fields explicitly, so the new attributes don't leak into
   OpenAPI and the frontend types are unchanged.
3. **Alembic.** `env.py` already excludes any `generated_*` column from
   autogenerate, so adding them to the model shouldn't produce a spurious
   revision. I could not run `alembic revision --autogenerate` to prove this —
   it currently fails on `master` for an unrelated reason
   (`NoReferencedTableError` on `saves.origin_device_id`).
4. **`player_count` ordering is lexicographic** (`VARCHAR(100)`, so `"10" < "2"`).
   That is unchanged: the view projects the same `VARCHAR` column, so this PR
   preserves the existing behavior rather than introducing it.

**Out of scope, but found while in here:** the `RomUser` branch of the same
function applies `query.filter(RomUser.user_id == user_id)` on top of an outer
join whose `ON` clause already carries that condition, which collapses it to an
effective inner join. Sorting by `last_played` therefore silently drops every ROM
with no `rom_user` row — 12,000 rows returned instead of 120,000 in my test data.
It is fast only because it discards 90% of the library. Left alone to keep this
PR focused; happy to open a separate issue.

**Checklist**

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this change was written with AI assistance (Claude Code).
The AI performed the root-cause investigation, wrote the tests first, implemented
the fix, and ran the benchmarks and the full test suite. I reviewed the diff, the
benchmark methodology and the test coverage before opening this PR.

### DIFF
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -156,11 +156,7 @@ ROM_FILE_HASH_COLUMNS_BY_DIGEST_LENGTH: dict[int, tuple[QueryableAttribute, ...]
     40: (RomFile.sha1_hash, RomFile.chd_sha1_hash),
 }
 
-# `roms_metadata` is a view over STORED generated columns on `roms`, so these
-# read the same values the view exposes without joining it. Ordering through the
-# view leaves the sort key on a joined table, which the engine cannot serve from
-# an index: it filesorts the whole library for every page. Every column here is
-# indexed on `roms`, so the sort walks the index and stops at the page.
+# Every column here is indexed on `roms`, so the sort walks the index and stops at the page.
 ROM_METADATA_ORDER_COLUMNS: dict[str, QueryableAttribute] = {
     "first_release_date": Rom.generated_first_release_date,
     "average_rating": Rom.generated_average_rating,
@@ -1423,8 +1419,6 @@ class DBRomsHandler(DBBaseHandler):
         elif order_by in ROM_METADATA_ORDER_COLUMNS:
             order_attr = ROM_METADATA_ORDER_COLUMNS[order_by]
         elif hasattr(RomMetadata, order_by) and not hasattr(Rom, order_by):
-            # The rest of the view is JSON arrays with no index to sort on, and
-            # nothing offers them as a sort, so they keep reading the view.
             order_attr = getattr(RomMetadata, order_by)
             query = query.outerjoin(RomMetadata, RomMetadata.rom_id == Rom.id)
         elif hasattr(Rom, order_by):
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -156,6 +156,17 @@ ROM_FILE_HASH_COLUMNS_BY_DIGEST_LENGTH: dict[int, tuple[QueryableAttribute, ...]
     40: (RomFile.sha1_hash, RomFile.chd_sha1_hash),
 }
 
+# `roms_metadata` is a view over STORED generated columns on `roms`, so these
+# read the same values the view exposes without joining it. Ordering through the
+# view leaves the sort key on a joined table, which the engine cannot serve from
+# an index: it filesorts the whole library for every page. Every column here is
+# indexed on `roms`, so the sort walks the index and stops at the page.
+ROM_METADATA_ORDER_COLUMNS: dict[str, QueryableAttribute] = {
+    "first_release_date": Rom.generated_first_release_date,
+    "average_rating": Rom.generated_average_rating,
+    "player_count": Rom.generated_player_count,
+}
+
 # Filter dropdowns read the narrow `roms_facets` mirror instead of `roms`,
 # whose rows carry the raw metadata blobs. Column order matches the unpacking
 # in `_collect_filter_values`.
@@ -1409,7 +1420,11 @@ class DBRomsHandler(DBBaseHandler):
         if user_id and hasattr(RomUser, order_by) and not hasattr(Rom, order_by):
             order_attr = getattr(RomUser, order_by)
             query = query.filter(RomUser.user_id == user_id)
+        elif order_by in ROM_METADATA_ORDER_COLUMNS:
+            order_attr = ROM_METADATA_ORDER_COLUMNS[order_by]
         elif hasattr(RomMetadata, order_by) and not hasattr(Rom, order_by):
+            # The rest of the view is JSON arrays with no index to sort on, and
+            # nothing offers them as a sort, so they keep reading the view.
             order_attr = getattr(RomMetadata, order_by)
             query = query.outerjoin(RomMetadata, RomMetadata.rom_id == Rom.id)
         elif hasattr(Rom, order_by):
--- a/backend/models/rom.py
+++ b/backend/models/rom.py
@@ -12,6 +12,7 @@ from sqlalchemy import (
     BigInteger,
     Boolean,
     Enum,
+    FetchedValue,
     Float,
     ForeignKey,
     Index,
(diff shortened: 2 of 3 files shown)


backend/handler/database/roms_handler.py
```diff
@@ -156,6 +156,17 @@
     40: (RomFile.sha1_hash, RomFile.chd_sha1_hash),
 }
 
+# `roms_metadata` is a view over STORED generated columns on `roms`, so these
+# read the same values the view exposes without joining it. Ordering through the
+# view leaves the sort key on a joined table, which the engine cannot serve from
+# an index: it filesorts the whole library for every page. Every column here is
+# indexed on `roms`, so the sort walks the index and stops at the page.
```
gantoine:
```suggestion
# Every column here is indexed on `roms`, so the sort walks the index and stops at the page.
```

backend/handler/database/roms_handler.py
```diff
@@ -1409,7 +1420,11 @@ def get_roms_query(
         if user_id and hasattr(RomUser, order_by) and not hasattr(Rom, order_by):
             order_attr = getattr(RomUser, order_by)
             query = query.filter(RomUser.user_id == user_id)
+        elif order_by in ROM_METADATA_ORDER_COLUMNS:
+            order_attr = ROM_METADATA_ORDER_COLUMNS[order_by]
         elif hasattr(RomMetadata, order_by) and not hasattr(Rom, order_by):
+            # The rest of the view is JSON arrays with no index to sort on, and
+            # nothing offers them as a sort, so they keep reading the view.
```
gantoine: remove this

backend/models/rom.py
```diff
@@ -401,6 +402,24 @@ class Rom(BaseModel):
         CustomJSON(), default=dict
     )
 
+    # The sortable slice of the STORED generated columns migration 0098 derives
+    # from the blobs above; `roms_metadata` is a thin view over them. Mapped
+    # here so a sort reads the indexed `roms` column directly: going through the
+    # view puts the sort key on a joined table, which rules the index out and
+    # filesorts the whole library. The engine owns the values, so these are
+    # read-only (see `RomMetadata` for the names the API exposes them under).
```
gantoine:
```suggestion
    # Read-only slice of the stored generated columns from the `roms_metadata` view
```

gantoine (changes requested):

gantoine (approved):

Issue #3420, opened 2026-05-24
[Bug] CSRF token not refreshed on OIDC login


CGamesPlay:
**RomM version**
4.8.1

**Describe the bug**
On first visiting Romm, the csrf_token is set with user_id: None, and then I am redirected to OIDC login. After logging in with OIDC, the csrf_token isn't changed, so it still has user_id: None.

**To Reproduce**

1. Go to '/'
2. Click on 'log in with Authelia' or whatever your OIDC provider is
3. Attempt to play a ROM
4. Get logged out

The specific request that causes the logout is `PUT /api/roms/{id}/props 403`

**Expected behavior**
The csrf token should be updated to include the correct user ID when the OIDC login completes.

**Server (please complete the following information):**

- OS: Linux, Docker
- Version `rommapp/romm:4`

**Client (please complete the following information):**

- Device: Computer
- OS: macOS 14.8.5 (23J423)
- Browser: Brave (Chrome) and Orion Browser (Safari)
- Version: Brave v 1.90.122; Orion Version 1.0.6 (145)

**Additional context**

Workaround: delete the `romm_csrftoken` cookie, which forces it to regenerate.

PR #4086, opened 2026-08-03
fix(auth): accept a stale user-bound CSRF token on anonymous requests

### LINKED ISSUE #4088: [Bug] First login attempt rejected with a CSRF error after a server restart
**RomM version**

5.1.1-beta.1 (also reproduces on 5.0.0; `csrf_middleware.py` is unchanged between them)

**Describe the bug**

Logging in fails once with a CSRF error, then succeeds on an immediate retry with the same credentials. It shows up as "I have to log in twice".

`_csrf_tokens_match` in `backend/handler/auth/middleware/csrf_middleware.py` validates the user id embedded in the CSRF token against the *current* request's user:

```python
and decoded_doc_cookie["user_id"] == user_id
and decoded_header_cookie["user_id"] == user_id
```

When the browser still holds a CSRF cookie bound to user N but the server-side session is gone (container restart, session expiry, Redis flush), the login POST is anonymous. `N != None`, so it is rejected. The rejection response then passes through `send()`, which rotates the cookie to an anonymous one, which is why the retry works.

**To Reproduce**

1. Log in normally, so the browser holds a CSRF cookie bound to your user id.
2. Restart the backend, or flush Redis, so the server-side session is dropped while the browser keeps its cookies.
3. Try to log in again.
4. The first POST is rejected with a CSRF error; the second attempt succeeds.

**Expected behavior**

The first login attempt should succeed. A CSRF token left over from a dead session should not cost the user a failed request.

- OS: Windows 11 + WSL2 (Ubuntu 26.04), Docker CE
- Version: 5.1.1-beta.1

- Device: Desktop PC
- OS: Windows 11
- Browser: Opera GX

**Additional context**

This is the mirror of #3420 (fixed by #3424), which handled the *authenticated* direction by rotating the cookie on safe requests. Both share a root cause: the rejection in `__call__` runs before the rotation in `send()` can help, so any user-binding mismatch costs one failed request.

I have a fix open in #4086.

bikeborb:
Logging in failed once with a CSRF error and succeeded on retry.

`_csrf_tokens_match` validates the user id embedded in the token against the current request's user. A browser holding a token bound to user N, whose server-side session is gone (container restart, session expiry, Redis flush), makes an anonymous login POST. N != None, so the request is rejected. The rejection response then rotates the cookie, which is why the second attempt works.

Keep the double-submit check and the cookie/header user agreement unconditionally, and enforce the caller binding only when there is a caller. An anonymous request cannot be forged this way regardless: an attacker cannot read the cookie cross-origin, so it cannot produce the matching header.

This is the mirror of #3420, which fixed the authenticated direction by rotating the cookie on safe requests. That path is untouched here, and test_authenticated_user_binding_still_enforced pins it.

**Description**

Fixes #4088

**Checklist**

### DIFF
--- a/backend/handler/auth/middleware/csrf_middleware.py
+++ b/backend/handler/auth/middleware/csrf_middleware.py
@@ -166,14 +166,29 @@ class CSRFMiddleware:
             decoded_doc_cookie = self.serializer.loads(document_cookie)
             decoded_header_cookie = self.serializer.loads(header_cookie)
 
-            # Verify that the tokens match, the user IDs match
-            # and the user_id matches the authenticated user
+            # The cookie and the submitted header must always agree, that is
+            # the double-submit check itself, and it holds regardless of who
+            # the caller is.
+            if not secrets.compare_digest(
+                decoded_doc_cookie["token"], decoded_header_cookie["token"]
+            ):
+                return False
+            if decoded_doc_cookie["user_id"] != decoded_header_cookie["user_id"]:
+                return False
+
+            # Bind the token to the caller only when there *is* one. An
+            # anonymous request holding a token from a dead session (server
+            # restart, expired session, cleared Redis) would otherwise be
+            # rejected once, and since the rejection response rotates the
+            # cookie, the retry would succeed. That is the "had to log in
+            # twice" bug. Double-submit still protects the anonymous case: an
+            # attacker cannot read the cookie cross-origin, so it cannot forge
+            # the matching header.
+            if user_id is None:
+                return True
+
             return (
-                secrets.compare_digest(
-                    decoded_doc_cookie["token"], decoded_header_cookie["token"]
-                )
-                and decoded_doc_cookie["user_id"] == decoded_header_cookie["user_id"]
-                and decoded_doc_cookie["user_id"] == user_id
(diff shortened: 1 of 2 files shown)


gantoine:
@greptileai

backend/handler/auth/middleware/csrf_middleware.py
```diff
@@ -166,14 +166,29 @@ def _csrf_tokens_match(
...
+                return False
+
+            # Bind the token to the caller only when there *is* one. An
+            # anonymous request holding a token from a dead session (server
+            # restart, expired session, cleared Redis) would otherwise be
+            # rejected once, and since the rejection response rotates the
+            # cookie, the retry would succeed. That is the "had to log in
+            # twice" bug. Double-submit still protects the anonymous case: an
+            # attacker cannot read the cookie cross-origin, so it cannot forge
+            # the matching header.
```
gantoine:
```suggestion
            # Bind the token to the caller only when there *is* one.
            # An anonymous request holding a token from a dead session would otherwise be rejected.
```

gantoine (changes requested):

gantoine (approved):

PR #4087, opened 2026-08-03
feat(stats): track ROM downloads and add an admin statistics tab

bikeborb:
Adds a `download_events` audit log plus denormalized per-ROM counters, so an operator can see what is actually being downloaded and identify content nobody uses.

- New `download_events` table: user, ROM, platform, source, kind, byte count, client IP and user agent, with name snapshots so a row survives the deletions it exists to inform.
- `roms.download_count` / `roms.last_downloaded_at`, incremented in SQL so concurrent downloads of the same ROM cannot lose an increment.
- Recording hooks on the single-ROM, single-file and bulk-zip download endpoints. Inline media (soundtracks, manuals) is excluded, so opening a game page does not inflate its count.
- `HybridAuthBackend` records how a request authenticated, which is what distinguishes a WebUI download from an API client.
- Admin-gated `/api/stats/downloads` endpoints and a Downloads tab in the existing Administration page.
- Optional retention via `DOWNLOAD_EVENTS_RETENTION_DAYS`, off by default.

Recording is best-effort and never fails a download.

**Description**

**Checklist**

image

image


bikeborb:
Following the feedback that this diff is too large, and after re-reading
CONTRIBUTING.md, I've filed #4089 to discuss the feature first and split
the work into two reviewable halves:

- Backend: 23 files, the event table, migration, API and tests
- Frontend: 57 files, but 48 of those are locale key copies and generated
  API types; the hand-written UI is 9 files

I'll open those once there's a steer on #4089, so please don't spend
review time on this PR. Happy to close it now or leave it as context.

gantoine:
sorry this is cool but i cant make heads or tails of the diff, going to close this and will tackle it sometime in the future!

PR #4112, opened 2026-08-04
fix(states): repoint the row when a re-upload changes emulator

TowyTowy:
**Description**

`POST /api/states` is a filename upsert: `get_state_by_filename` matches on `(rom_id, user_id, file_name)` with no emulator scoping, while the bytes are written under a path derived from *this* request's emulator. The upsert branch only wrote `file_size_bytes`, so re-uploading the same filename under a different emulator left the row pointing at the previous emulator's folder while reporting the new file's size. `GET /api/states/{id}/content` then served the old state, and the freshly written bytes were orphaned on disk.

This persists `file_path`/`emulator` alongside the size, and drops the bytes at the old location when no other row still references them — the same shape already applied to `POST /api/saves` in #3833 (`backend/endpoints/saves.py`), including the `still_referenced` guard and the `FileNotFoundError` catch. States were simply missed by that change.

> AI assistance disclosure: this PR was prepared with Claude Code — bug analysis, the fix, and the regression test — reviewed and verified fail-before/pass-after before submitting.

**Checklist**

### DIFF
--- a/backend/endpoints/states.py
+++ b/backend/endpoints/states.py
@@ -105,9 +105,31 @@ async def add_state(
         user_id=request.user.id, rom_id=rom.id, file_name=sanitized_state_filename
     )
     if db_state:
+        # Track file path and emulator to keep the row pointing at the bytes
+        # this request wrote, which land under the requested emulator's folder.
+        stale_full_path = db_state.full_path
         db_state = db_state_handler.update_state(
-            db_state.id, {"file_size_bytes": scanned_state.file_size_bytes}
+            db_state.id,
+            {
+                "file_size_bytes": scanned_state.file_size_bytes,
+                "file_path": scanned_state.file_path,
+                "emulator": emulator,
+            },
         )
+
+        # Delete orphaned bytes only if no other row references the old path.
+        if stale_full_path != db_state.full_path:
+            still_referenced = any(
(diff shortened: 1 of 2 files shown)


gantoine:
@greptileai

backend/endpoints/states.py
```diff
@@ -105,9 +105,31 @@ async def add_state(
...
-            db_state.id, {"file_size_bytes": scanned_state.file_size_bytes}
+            db_state.id,
+            {
+                "file_size_bytes": scanned_state.file_size_bytes,
+                "file_path": scanned_state.file_path,
+                "emulator": emulator,
+            },
         )
+
+        # Delete orphaned bytes only if no other row references the old path.
```
gantoine: why would another row reference this file if it was just moved?

gantoine (changes requested):

backend/endpoints/states.py
TowyTowy: You're right, it can't. States are unique per (rom, user, file_name) and the emulator is part of the path, so the only row that could ever hold the old path is the one being moved. I mirrored the guard from saves.py without checking whether it actually applied here.

Dropped it in f3d6be1 - the stale file is now removed unconditionally once the path changes. Tests still pass.

gantoine (approved):
nice!

PR #4113, opened 2026-08-04
fix(v2): drop a random pick after leaving the platform

### LINKED ISSUE #4104: [Bug] "Random ROM" can throw you into a game from the platform you just left
**RomM version**
5.1.1-beta.1 and current `master` (906a1c053)

**Describe the bug**
On a platform gallery, clicking "Random ROM" starts a request and navigates to the picked game when it returns. It never re-checks whether you are still on the platform you clicked from.

If you click "Random ROM" and then switch to another platform before the pick resolves, the in-flight pick still fires and pulls you into a game from the platform you left. You land on a game you did not ask for, in a platform you already navigated away from.

The collection gallery already guards against exactly this: it captures the collection id and abandons the navigation if it has changed. The platform gallery does not.

**To Reproduce**
1. Open a large platform's gallery, large enough that the pick takes a visible moment.
2. Click "Random ROM".
3. Immediately navigate to a different platform.
4. The app jumps to a game belonging to the first platform.

**Expected behavior**
An in-flight random pick is abandoned if you have navigated away from the platform it was scoped to.

**Additional context**
`frontend/src/v2/views/Gallery/Platform.vue`, in `onRandomGame()`. The fix is the same short guard `Collection.vue` already carries.

The window is currently seconds wide on a large library. PR #4097 collapses the pick into a single fast request, which narrows the window to milliseconds but does not close it, and that asymmetry was flagged there as out of scope.

Spinnich:
**Description**

The v2 platform gallery resolves "Random ROM" over two requests: a count-only
fetch for `total`, then a single-item fetch at a random offset. Neither result
was re-checked against the platform still on screen, so on a large library the
pick could land after the user had already switched platforms and navigate them
into a game from the gallery they just left.

`onRandomGame()` now captures the platform id at click time and compares it
against `currentPlatform` after each await, discarding a stale pick silently.
Because the first check sits between the two requests, leaving the platform now
also skips the second request instead of firing it and throwing the result away.

The same staleness check also gates the error handler: these requests are issued
without an abort `signal`, so navigating away never cancels them, and one that
failed afterwards still raised "Couldn't pick a random ROM" for the platform the
user had left. Thanks to Greptile for catching that path.

The collection gallery already carried the scope capture and the post-await
checks, but had the same unconditional error handler, so it gets the one-line
mirror here to keep the two galleries behaving identically.

Fixes #4104

**Files changed**

- `frontend/src/v2/views/Gallery/Platform.vue`: capture `scopeId` at click time,
  add a `stale()` check after both awaits, and scope both requests to `scopeId`
  rather than re-reading the live platform. The `catch` is gated on the same
  check. Every early return sits inside the existing `try`, so `finally` still
  clears `randomLoading` and the button never wedges.
- `frontend/src/v2/views/Gallery/Platform.test.ts` (new): 8 mount-level cases for
  the random pick. Two hold a request in flight and move the view to another
  platform (one per await point) and assert no navigation happens, one rejects an
  in-flight request after the move and asserts no snackbar, one pins that a
  same-platform reload still navigates, and the rest cover the scoped happy path
  plus the empty, failed, and already-in-flight branches.
- `frontend/src/v2/views/Gallery/Collection.vue`: mirror of the same error-path
  guard. It already captured the collection id and checked both awaits, so the
  unconditional `catch` was the only piece out of step.
- `frontend/src/v2/views/Gallery/Collection.test.ts` (new): three cases for the
  collection random pick, covering the scoped happy path, a stale resolved pick,
  and a stale rejection.

**Testing**

- `npx vitest run src/v2/views/Gallery/`: 11/11. Each new guard was written
  test-first, so every stale case fails against the unpatched view and passes
  after. The two stale-rejection cases also click again afterwards to prove
  `finally` still cleared the loading flag.
- Full frontend suite: 647 passed, 54 files. `vue-tsc` clean. `trunk fmt` and
  `trunk check` clean. Backend suite unaffected and green (2731 passed, 2
  skipped).
- Browser-verified against a local dev stack with Playwright stalling only the
  two `limit=1` `/api/roms` requests to widen the window, then navigating in-app
  (platform 1, platforms index, platform 2) rather than reloading. Before the
  fix the run ended on `/rom/2`, a platform 1 game, while sitting on
  `/platform/2`. After the fix it stays on `/platform/2` and the second request
  is never issued.

**Anything a reviewer should look at**

- The check compares ids, not object identity, on purpose: the view's background
  `platformApi.getPlatform` refresh swaps in a fresh `Platform` record for the
  same platform, and an identity check would wrongly discard a valid pick after
  a same-platform reload. There is a test for exactly that.
- A stale pick is dropped without a snackbar, on both the success and the failure
  path. The user has already moved on, so feedback about a gallery they left
  would be noise.
- Known residual, deliberately out of scope: `stale()` keys off the loaded
  gallery scope, which is only cleared when another gallery view mounts, so
  navigating to a non-gallery route (Home, Settings) with a pick in flight can
  still navigate. Both galleries share that behaviour.
- Overlaps open PR #4097, which rewrites the platform function onto a single
  `getRandomRom` call. Whichever lands second needs a small rebase: on the
  single-request shape this collapses to one `stale()` check, and the two
  `Platform.test.ts` files need merging. Happy to rebase this one on top if
  #4097 goes first.

**AI Assistance**

This PR (code, tests, and description) was written primarily by Claude Code,
reviewed and verified by me locally.

**Checklist**

Not applicable, no visual change. The fix is the absence of an unwanted
navigation and of an unwanted error snackbar.

### DIFF
--- a/frontend/src/v2/views/Gallery/Collection.test.ts
+++ b/frontend/src/v2/views/Gallery/Collection.test.ts
@@ -237,6 +237,37 @@ describe("Collection view random rom", () => {
     expect(push).not.toHaveBeenCalled();
   });
 
+  it("stays quiet when a pick fails after the view moved to another collection", async () => {
+    let failPick: (reason: Error) => void = () => {};
+    getRandomRom.mockReturnValueOnce(
+      new Promise((_resolve, reject) => {
+        failPick = reject;
+      }),
+    );
+
+    const wrapper = await mountView();
+    await wrapper.get("button.random").trigger("click");
+
+    routeState.params = { collection: "2" };
+    routeGuards.forEach((guard) =>
+      guard({ name: "collection", params: { collection: "2" } }),
+    );
+    await flushPromises();
+
+    failPick(new Error("boom"));
+    await flushPromises();
+
+    expect(snackbarError).not.toHaveBeenCalled();
+    expect(push).not.toHaveBeenCalled();
+
+    // The button still works on the collection the user landed on.
+    getRandomRom.mockResolvedValue({ data: rom(7) });
+    await wrapper.get("button.random").trigger("click");
+    await flushPromises();
+
+    expect(push).toHaveBeenCalledWith({ name: "rom", params: { rom: 7 } });
+  });
+
   it("ignores a click while a pick is in flight", async () => {
     let resolvePick: (value: { data: SimpleRom }) => void = () => {};
     getRandomRom.mockReturnValue(
--- a/frontend/src/v2/views/Gallery/Collection.vue
+++ b/frontend/src/v2/views/Gallery/Collection.vue
@@ -288,17 +288,18 @@ async function onRandomGame() {
   if (!c || randomLoading.value) return;
   randomLoading.value = true;
   const scopeId = c.id;
+  // A pick from the collection the user just left leads nowhere useful.
+  const stale = () => currentCollection.value?.id !== scopeId;
   try {
     const { data } = await romApi.getRandomRom(randomScope());
-    // A pick from the collection the user just left leads nowhere useful.
-    if (currentCollection.value?.id !== scopeId) return;
+    if (stale()) return;
     if (!data) {
       snackbar.info(t("collection.empty"));
       return;
     }
     router.push({ name: ROUTES.ROM, params: { rom: data.id } });
   } catch {
-    snackbar.error(t("platform.random-rom-error"));
+    if (!stale()) snackbar.error(t("platform.random-rom-error"));
   } finally {
     randomLoading.value = false;
   }
--- a/frontend/src/v2/views/Gallery/Platform.test.ts
+++ b/frontend/src/v2/views/Gallery/Platform.test.ts
@@ -137,6 +137,17 @@ function deferRandomRom() {
   return (pick: SimpleRom | null) => settle({ data: pick });
 }
 
+/** Rejects the promise the next `getRandomRom` call returns, on demand. */
+function deferRandomRomFailure() {
+  let fail: (reason: Error) => void = () => {};
+  getRandomRom.mockReturnValueOnce(
+    new Promise((_resolve, reject) => {
+      fail = reject;
+    }),
+  );
+  return () => fail(new Error("boom"));
+}
+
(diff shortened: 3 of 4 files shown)


Spinnich:
Valid, thanks @greptileai. Fixed in 2ec1cf240, and mirrored to the collection gallery in 05e9e6dae.

`onRandomGame` doesn't pass an abort `signal`, so navigating away never cancels
these requests and a rejection is a real failure rather than an abort, which
makes the irrelevant toast reachable. Both galleries now gate the `catch` on the
same `stale()` check the success path uses, and it stays inside the `try` so
`finally` still clears `randomLoading`. Added tests on both views that reject an
in-flight request after switching scope and assert no snackbar, no navigation,
and a still-working button on the scope the user landed on.

gantoine (approved):

gantoine:
@Spinnich looks good, fix the conflict and ill merge it

Spinnich:
@gantoine Rebased onto `master` now that #4097 has landed, so the conflict is resolved.

That PR replaced both galleries' two-request pick with a single `getRandomRom`
call, which absorbed a good chunk of what this branch was carrying. The guard
collapses to one check per view, and the diff is down to 4 files, +151/-13.

Where things ended up:

- `Platform.vue` captures the platform id at click time and drops the pick if the
  view has moved on, with the `catch` gated on the same check so a request that
  fails after you switch platforms no longer raises a snackbar for the one you
  left.
- `Collection.vue` already had the equivalent check inline from #4097. I folded it
  into the same small `stale()` closure and gated its `catch` too, so both views
  now read identically. That is the only behavioural change on the collection
  side.
- Both test files keep the #4068 coverage from #4097 unchanged. This adds three
  cases on top: a stale pick that must not navigate, a stale rejection that must
  not raise a snackbar, and one pinning that a same-platform reload still
  navigates (the background `getPlatform` refresh swaps in a fresh `Platform`
  record, so the check has to compare ids rather than object identity).

Each new case was re-verified against the rebased base by reverting the
corresponding guard and confirming it fails, so they do pin the behaviour rather
than passing by construction.

Locally: 655 tests green across 55 files, `vue-tsc` clean, `trunk check` clean.

PR #4120, opened 2026-08-05
fix(v2): keep Playmatch selectable when providers are set to All

sdornan:
**Description**

In the v2 provider selects, "All" is represented by an **empty model**: `RSelect`'s
All-mode strips its own items from the `v-model` rather than listing them. The
Playmatch gate in both scan dialogs read that raw model:

```ts
const igdbSelected = metadataSources.value.some((s) => s.value === "igdb");
```

So with Providers set to All, `igdbSelected` was `false` and the Playmatch switch
stayed disabled with "Select IGDB to enable Playmatch" — even though the scan was
about to run IGDB.

The same read fed the payload, so the bug was not only cosmetic. An All-mode group
contributed nothing to `apis`, and the backend reads an empty `apis` list as *no
sources*, not "all". Setting both provider groups to All in either dialog would
have scanned with no providers at all, and the submit button greyed out because it
gated on `metadataSources.length`.

`Scan.vue` already had the fix — `effectiveMetadataSources` expands an All-mode
group to its enabled providers — but the two dialogs never got it. They each
carried their own copy of the provider list, the localStorage keys, the hash-matcher
definitions and the payload builder, which is how they drifted in the first place.
Rather than add a third copy of the expansion, this moves the whole model into a
shared `useScanProviders` composable and points all three surfaces at it.

**Reproduction**

1. Open any ROM's ⋮ menu → Refresh metadata (or a platform's Scan dialog).
2. In Providers → General, choose **All**.
3. Playmatch under Proxies is disabled, tooltipped "Select IGDB to enable Playmatch".

Picking IGDB explicitly instead of All enabled it, which is what made this look
like a display bug rather than an empty-payload bug.

**Files changed**

- `frontend/src/v2/composables/useScanProviders/index.ts` (new): the shared model —
  provider categorisation, the four `scan.*` localStorage keys, `metadataOptions`,
  `effectiveMetadataSources`, the Hasheous/Playmatch matchers, `buildScanPayload()`
  and `persistSelection()`. Two behavioural notes versus the code it replaces: an
  explicit pick now always wins over the All flag, so the gate is correct even
  before a select has mounted and emitted its initial All state; and the restore
  watch only re-derives the selection when the *stored* list changes, so a heartbeat
  refresh or an admin toggle no longer discards an in-progress selection (it just
  prunes providers that went away). The three surfaces lose 587 lines between them.
- `frontend/src/v2/composables/useScanProviders/index.test.ts` (new): 12 cases. The
  load-bearing one asserts Playmatch is enabled with the general group in All-mode,
  which fails against the old logic. The rest cover explicit picks winning over All,
  the hash-matcher exclusions, the `SKIP_HASH_CALCULATION` gate, the expanded `apis`
  payload, and both restore paths.
- `frontend/src/v2/components/Dialogs/RefreshMetadataDialog.vue`,
  `frontend/src/v2/components/Gallery/ScanPlatformDialog.vue`: consume the composable.
  This is where the reported bug lives. Both also had the empty-`apis` and
  disabled-submit consequences described above.
- `frontend/src/v2/views/Scan.vue`: consume the composable. No behaviour change — it
  already resolved All-mode correctly; this just removes the copy it was resolving with.
- Dropped a stray inline i18n fallback in `ScanPlatformDialog`
  (`t("scan.playmatch-requires-igdb", "Select IGDB to enable Playmatch.")`); the key
  exists in every locale, and the hardcoded default shadowed the translated string
  with English.

**Verification**

`npm run typecheck` clean, eslint and prettier clean on the changed files, and the
full suite passes at 667 (655 before, +12 here). I have not exercised this in a
browser — there is no RomM instance running against this worktree — so the
theme/input pass is unverified and a maintainer may want to eyeball the two dialogs.
The logic is covered by the unit tests above.

**AI Assistance Disclosure**

AI assistance (Claude Code) was used for this contribution: the root-cause analysis,
the implementation, the unit tests and this description were AI-generated, then
reviewed by me before submission. Any replies I post on this PR are written by me
unless stated otherwise.

**Checklist**

### DIFF
--- a/frontend/src/v2/composables/useScanProviders/index.test.ts
+++ b/frontend/src/v2/composables/useScanProviders/index.test.ts
@@ -5,7 +5,7 @@ import { useScanProviders, type HashMatcher } from "./index";
 
 // Provider list the heartbeat would expose with every source enabled.
 const OPTIONS: MetadataOption[] = [
-  { value: "igdb", name: "IGDB + Playmatch", logo_path: "", disabled: "" },
+  { value: "igdb", name: "IGDB", logo_path: "", disabled: "" },
   { value: "ss", name: "ScreenScraper", logo_path: "", disabled: "" },
   { value: "moby", name: "MobyGames", logo_path: "", disabled: "" },
   { value: "ra", name: "RetroAchievements", logo_path: "", disabled: "" },
--- a/frontend/src/v2/composables/useScanProviders/index.test.ts
+++ b/frontend/src/v2/composables/useScanProviders/index.test.ts
@@ -1,0 +1,157 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { nextTick, ref } from "vue";
+import type { MetadataOption } from "@/stores/heartbeat";
+import { useScanProviders, type HashMatcher } from "./index";
+
+// Provider list the heartbeat would expose with every source enabled.
+const OPTIONS: MetadataOption[] = [
+  { value: "igdb", name: "IGDB + Playmatch", logo_path: "", disabled: "" },
+  { value: "ss", name: "ScreenScraper", logo_path: "", disabled: "" },
+  { value: "moby", name: "MobyGames", logo_path: "", disabled: "" },
+  { value: "ra", name: "RetroAchievements", logo_path: "", disabled: "" },
+  { value: "hasheous", name: "Hasheous", logo_path: "", disabled: "" },
+  { value: "playmatch", name: "Playmatch", logo_path: "", disabled: "" },
+];
+
+const heartbeat = {
+  value: {
+    METADATA_SOURCES: {
+      HASHEOUS_API_ENABLED: true,
+      PLAYMATCH_API_ENABLED: true,
+    },
+  },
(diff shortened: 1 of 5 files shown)


frontend/src/v2/composables/useScanProviders/index.test.ts
```diff
@@ -0,0 +1,157 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { nextTick, ref } from "vue";
+import type { MetadataOption } from "@/stores/heartbeat";
+import { useScanProviders, type HashMatcher } from "./index";
+
+// Provider list the heartbeat would expose with every source enabled.
+const OPTIONS: MetadataOption[] = [
+  { value: "igdb", name: "IGDB + Playmatch", logo_path: "", disabled: "" },
```
gantoine: as playmatch is its own source now, do we even need igdb + playmatch as an option?

sdornan: Good catch. Removed.

gantoine:
@greptileai

gantoine (approved):
very nice

PR #4129, opened 2026-08-05
fix(hashing): extract CHD SHA-1 for folder-based ROMs

### LINKED ISSUE #4126: [Bug] CHD SHA-1 is not extracted when ROM files are inside a folder per game
**RomM version**
5.1.1-beta.1

**Describe the bug**
The CHD SHA-1 hash is not extracted from CHD files if the ROM is inside a folder. For instance, a multi-disc PSX game that is all on the same folder with similar naming scheme (without `(Disc N)` indicator). Also happens on single file ROMs.

**To Reproduce**

1. Add CHD rom files inside a folder per game
2. Scan library
3. CHD SHA-1 is not extracted

**Expected behavior**
CHD SHA-1 is extracted for the .chd files inside folder. For multi-disc ROMs, at least the Disc 1 should be extracted.

**Screenshots**
Multi-disc ROM inside folder

Image

Same game, Disc 1 not in folder (directly in `/romm/library/roms/psx`)

Image

Another single-disc ROM inside folder, without CHD SHA-1 extracted

Image

- OS: Ubuntu 26.04
- Docker 29.4.0

sdornan:
**Description**

Fixes #4126.

The CHD disc-data SHA-1 was never extracted for ROMs stored in a per-game folder. In the folder branch of `FSRomsHandler.get_rom_files`, the path handed to `is_chd_file()` / `extract_chd_hash()` was built from the ROM's folder name instead of the file being walked:

```python
for f_path, file_name in iter_files(f"{abs_fs_path}/{rom.fs_name}", recursive=True):
    f_rom_dir = Path(f_path, rom.fs_name)   # .../psx/Final Fantasy VII/Final Fantasy VII
```

`f_path` is the directory holding the file and `file_name` is the file, so the probed path pointed at a file that doesn't exist. `is_chd_file()` found no `.chd` suffix, libmagic raised `OSError` on the missing path and returned `False`, and `chd_sha1_hash` came out empty for every file of every folder-based ROM. This is not specific to multi-disc sets or `(Disc N)` naming; single-disc games in a folder were affected too. Single files sitting directly in the platform folder always used the correct path, which is why those worked.

Beyond the missing column in the UI, this also degraded metadata matching: `HasheousHandler` sends only the CHD disc-data SHA-1 when it is present, because a CHD's raw MD5/SHA-1/CRC are hashes of the container and match nothing in the database. With the field blank, folder-based CHDs were looked up by those container hashes instead.

Changes:

- Resolve the walked file's real path once (`abs_file_path`) and use it for both the raw hashes and the CHD header probe.
- Extract the probe into a `_chd_sha1_hash()` helper shared by the folder and single-file branches, and run it via `asyncio.to_thread` like the surrounding hashing calls, since it now performs real I/O per file.
- Add a regression test covering a two-disc CHD folder plus an `.m3u`: each disc keeps its own header SHA-1, the playlist stays empty.

Note for existing libraries: `_should_get_rom_files` only rebuilds ROM files for newly added ROMs or a **Complete**/**Hashes** scan, so already-scanned folder ROMs need a hashes rescan to backfill the value.

**Checklist**

N/A — backend only, no UI change.

---

**AI assistance disclosure**

This change was written with AI assistance (Claude Code). The AI performed the investigation, wrote the fix and the regression test, and drafted this description; all of it was reviewed by me before submitting. Verification run locally: the new test fails on `master` and passes with the fix, `uv run pytest` green across the backend suite, and ruff/mypy clean on the changed files.

### DIFF
--- a/backend/handler/filesystem/roms_handler.py
+++ b/backend/handler/filesystem/roms_handler.py
@@ -430,9 +430,7 @@ class FSRomsHandler(FSHandler):
                         crc_c,
                         md5_h,
                         sha1_h,
-                        chd_sha1_hash=await asyncio.to_thread(
-                            _chd_sha1_hash, abs_file_path
-                        ),
+                        chd_sha1_hash=_chd_sha1_hash(abs_file_path),
                     )
                 else:
                     file_hash = FileHash(
@@ -572,7 +570,7 @@ class FSRomsHandler(FSHandler):
                 crc_c,
                 md5_h,
                 sha1_h,
-                chd_sha1_hash=await asyncio.to_thread(_chd_sha1_hash, rom_dir),
+                chd_sha1_hash=_chd_sha1_hash(rom_dir),
             )
             rom_files.append(
                 self._build_rom_file(
--- a/backend/handler/filesystem/roms_handler.py
+++ b/backend/handler/filesystem/roms_handler.py
@@ -129,7 +129,7 @@ ARCHIVE_READERS = {
 
 
 def _chd_sha1_hash(file_path: Path) -> str:
-    """Return the disc-data SHA1 embedded in a CHD v5 header, or "" for non-CHD files."""
+    """Return the embedded CHD v5 raw+meta SHA-1, or "" for non-CHD files."""
     return extract_chd_hash(file_path) if is_chd_file(file_path) else ""
 
 
--- a/backend/handler/filesystem/roms_handler.py
+++ b/backend/handler/filesystem/roms_handler.py
@@ -128,6 +128,11 @@ ARCHIVE_READERS = {
 }
 
 
+def _chd_sha1_hash(file_path: Path) -> str:
+    """Return the disc-data SHA1 embedded in a CHD v5 header, or "" for non-CHD files."""
+    return extract_chd_hash(file_path) if is_chd_file(file_path) else ""
+
+
 def _make_file_hash(
     crc_c: int, md5_h: Any, sha1_h: Any, chd_sha1_hash: str = ""
 ) -> FileHash:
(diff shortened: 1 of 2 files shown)


backend/handler/filesystem/roms_handler.py
```diff
@@ -568,9 +572,7 @@ def _hash_raw_archive(crc: int) -> int:
                 crc_c,
                 md5_h,
                 sha1_h,
-                chd_sha1_hash=(
-                    extract_chd_hash(rom_dir) if is_chd_file(rom_dir) else ""
-                ),
+                chd_sha1_hash=await asyncio.to_thread(_chd_sha1_hash, rom_dir),
```
gantoine: why asyncio?

sdornan: From Claude:
> _chd_sha1_hash does blocking disk I/O, so it shouldn't run on the event loop thread: for anything without a .chd suffix is_chd_file() falls through to libmagic's from_file() (opens the file, reads its leading bytes, under a process-wide lock), and extract_chd_hash() then opens the file again for the 124-byte v5 header. get_rom_files is awaited from the scan socket handler, which is concurrently emitting progress and servicing the stop signal, and every other hashing call in this function is already wrapped the same way (_calculate_rom_hashes, _hash_archive_entries, _hash_raw_archive).
> 
> The motivation is really the folder branch: before the fix the probe pointed at a path that didn't exist, so it never opened anything; now it runs once per walked file, which for a large folder ROM is that many magic probes. I used the same helper here so both branches behave identically.

Honestly, front-end development is my day job, so threading and asynchronicity isn't my strong suit. Each read is pretty small, so happy to keep it single-threaded if preferred.

gantoine: yes please

> front-end development is my day job

your frontend PRs are much cleaner IMO, i'd stick to those and let us handle the backend stuff

sdornan: Swapped back

gantoine (approved):
nice!

Issue #4124, opened 2026-08-05
[Feature] Validate ScreenScraper credentials and show quota usage on the metadata sources page


Spinnich:
**Is your feature request related to a problem? Please describe.**

RomM never checks whether your ScreenScraper credentials actually work. It only
checks that the settings are non-empty:

```python
def is_enabled(cls) -> bool:
    return bool(SCREENSCRAPER_USER and SCREENSCRAPER_PASSWORD)
```

That value drives the "ScreenScraper enabled" state in the UI. So an instance with a
wrong password shows ScreenScraper as configured, offers it in the scan picker, and
appears to work.

**The reason this matters is how ScreenScraper's API is structured.** Scraping is
authorized by the *application's* developer credentials, which RomM ships. The user's
own credentials play a much smaller role:

| parameter | role |
|---|---|
| `devid` / `devpassword` | authorizes the request (RomM's, shipped in the image) |
| `ssid` | determines whose quota is charged |
| `sspassword` | **never checked on scraping endpoints** — only on account endpoints such as `ssuserInfos.php` |

Verified directly:

```
valid devid + valid user password   HTTP 200  game data returned
valid devid + WRONG user password   HTTP 200  game data returned
valid devid + NO user credentials   HTTP 200  game data returned
NO devid    + valid user password   HTTP 403  "Vérifier vos identifiants développeur"
```

So if a user's password is wrong, **scraping still succeeds on RomM's developer
credentials, and their account is still charged for every request** — measured at
exactly 1:1 over 150 requests. What they lose is their member entitlement. They are
capped at the much lower unauthenticated allowance while paying full price for every
lookup and every image.

The user gets the worst of both outcomes: their daily allowance is consumed at full
rate, they are cut off far earlier than their account should allow, and RomM never
indicates that anything is wrong with their credentials.

That is exactly what happened in a recent report. Their password was wrong (quotes
preserved literally by Docker's `env_file`, which does not strip them the way a shell
does). Their account was charged over 10,000 requests that morning, they were then
refused with quota errors, and their ScreenScraper account page showed plenty of
quota remaining throughout. It took several days to identify.

Quoting is only one way to get there. A `$` consumed by Docker Compose variable
substitution, a `#` starting a comment, trailing whitespace, or a plain typo all
produce the same silent outcome.

**Describe the solution you'd like**

On the metadata sources page, show ScreenScraper's real status:

- **User credentials: valid / rejected.** Checked with a real call to
[...]

Issue #4123, opened 2026-08-05
[Bug] A single ScreenScraper quota error disables ScreenScraper for an entire scan, and manual matching never recovers


Spinnich:
**RomM version**
5.1.1-beta.1. The behaviour dates back to 5.0.0-beta.2 (introduced in #3686), so it
is not specific to recent releases. Re-verified on `master` at `d10bae866`
(2026-09-05), this time by running the code rather than reading it: still present,
all four items. Item 3 is the only part addressed since, and only half of it; see
the note there.

**Describe the bug**
The first time ScreenScraper returns a daily-quota error (HTTP 430 or 431), RomM
sets a flag that makes every subsequent ScreenScraper request fail instantly for the
rest of the scan. It never re-checks. One error, even a transient one or one that has
nothing to do with the user's own allowance, costs them ScreenScraper metadata for
their entire library.

In a real report this tripped **1.4 seconds into the scan**, on the very first ROM:

```
[2026-08-05T11:17:28.517Z] DEBUG [screenscraper] API request: URL=.../jeuInfos.php?crc=919f23cb&...
[2026-08-05T11:17:28.709Z] WARNING [screenscraper] ScreenScraper daily scrape quota exhausted; skipping ScreenScraper for the rest of this scan (quotas reset at midnight CET)
```

Every ROM after that was saved without ScreenScraper data.

Crucially, that user's account was **not** out of quota in the way RomM implied.
Their account page showed 10,010 of 50,000 daily requests used. Their password was
wrong, so ScreenScraper charged their account for every request but never granted
them their member allowance, capping them at the far lower unauthenticated limit.
RomM told them their daily quota was exhausted; their own account page said
otherwise; both were describing different limits.

Four related problems:

**1. HTTP 431 should never stop a scan, and currently does.** 430 and 431 are treated
identically, but they mean completely different things. A 431 tells you two things at
once: this ROM did not match, and the account has reached its daily limit for
submitting unknown ROMs to ScreenScraper's moderation queue for possible addition to
the database.

The first half is ordinary. For metadata purposes it is the same outcome as a plain
"not found", and continuing costs nothing extra: without the cap that request would
have returned 404, so the request count is identical either way.

The second half is a real loss, but not a scraping failure. The user's unmatched ROMs
simply stop being proposed to the moderators for the rest of the day. Every ROM that
does match still scrapes perfectly. Treating this as grounds to disable ScreenScraper
for the entire scan is wrong on both counts.

[...]

PR #4130, opened 2026-08-05
fix(screenscraper): report rejected credentials instead of a bare 403

### LINKED ISSUE #4122: [Bug] ScreenScraper credential failures are reported as something else, hiding the real cause
**RomM version**
5.1.1-beta.1 (code path unchanged on current `master`)

**Describe the bug**
When ScreenScraper rejects the configured credentials, RomM never says so. The log
shows a generic `403, message='Forbidden'`, the scan continues, and every ROM is
saved without ScreenScraper metadata. Nothing anywhere mentions credentials.

This was confirmed against a real report. The user's `.env` had their username and
password wrapped in quotes, which Docker's `env_file` parser preserves literally, so
RomM was sending `ssid="myusername"` instead of `ssid=myusername`. ScreenScraper
replied with a message that says exactly what was wrong:

```
HTTP 403  Erreur de login : Vérifier les identifiants utilisateurs !
```

RomM logged this instead:

```
[2026-08-05T11:17:27.623Z] ERROR [screenscraper] 403, message='Forbidden', url='https://api.screenscraper.fr/api2/ssuserInfos.php'
```

Diagnosing it took several hours, and the user spent most of that time investigating
their ScreenScraper quota, because that is what RomM told them was wrong.

There are four separate problems, all in `_request()` in
`backend/adapters/services/screenscraper.py`:

**1. RomM already has the correct error message, and it can never run.**
`LOGIN_ERROR_CHECK` is the string `"Erreur de login"`. When RomM finds it in a
response body it raises a clear `Invalid ScreenScraper credentials`. But the check
sits *after* `res.raise_for_status()`, so it only runs on a successful response.
ScreenScraper sends that string with **HTTP 403**, which raises first. The check is
dead code for the exact case it was written for. Both the first attempt and the
retry are affected.

**2. HTTP 403 has no handler.** It falls into the catch-all `else` branch, which
does `log.error(err)` and returns an empty dict. Compare HTTP 426, which gets a
clear, actionable message ("ScreenScraper has blacklisted this application version.
Please update RomM."). 403 deserves the same.

**3. A failed account check produces no warning at all.** `prime_account_limits()`
runs at the start of every scan and is written to log a warning if it cannot read
the account. Because a 403 *returns* an empty dict rather than raising, that warning
never fires. The scan then runs with no account limits loaded, which also suppresses
the "ScreenScraper quota:" line. The result is complete silence.

**4. HTTP 401 tells only half the story.** RomM logs "ScreenScraper API is
temporarily unavailable (server CPU >60%)". ScreenScraper's error table at
https://www.screenscraper.fr/webapi2.php splits that row across two columns:
the *description* is "API fermé pour les non membres ou les membres inactifs"
(API closed to non-members and inactive members), and the *cause* is "Le Serveur
est saturé (utilisation CPU>60%)". RomM reports the cause and drops the
description, so an operator whose membership has lapsed reads a 401 as a passing
server-side problem and waits it out. The message should carry both halves.

**To Reproduce**

1. In your `.env`, put quotes around the credentials (Docker keeps them literally,
   unlike a shell):
   ```
   SCREENSCRAPER_USER="myusername"
   SCREENSCRAPER_PASSWORD="mypassword"
   ```
2. `docker compose up -d` to recreate the container.
3. Start a scan with ScreenScraper enabled.
4. The log shows a bare `403, message='Forbidden'` and the scan continues.
5. Nothing states that the credentials were rejected.

**Expected behavior**

- A 403 is reported for what it is, for example: `ScreenScraper rejected the
  configured credentials. Check SCREENSCRAPER_USER and SCREENSCRAPER_PASSWORD.`
- The `LOGIN_ERROR_CHECK` body check runs before `raise_for_status()` so it can
  actually match, or a dedicated 403 branch replaces it.
- A failed `prime_account_limits()` logs a warning whether the underlying call
  raised or returned empty.
- **The account check reports the refusal but never disables ScreenScraper for the
  scan.** It is diagnostic, and ScreenScraper's developer-credential enforcement
  flaps (see #4124), so a scan whose scraping still works must not lose it to a
  failed check. Only a request the scan actually needs may take the provider out.
  This is the same rule #4123 asks for on the quota side.
- The 401 message carries both halves of ScreenScraper's documented row, so the
  closure and its stated cause are both visible.

**Additional context**

ScreenScraper's 403 body identifies which credential set failed, so RomM can report
it precisely. Verified against `jeuInfos.php`:

```
valid devid + valid user pw   HTTP 200  (game data returned)
valid devid + WRONG user pw   HTTP 200  (game data returned)
NO devid    + valid user pw   HTTP 403  "Vérifier vos identifiants développeur"
BOGUS devid + valid user pw   HTTP 403  "Vérifier vos identifiants développeur"
```

Important caveat: this wording is reliable on `jeuInfos.php` but **not** on
`ssuserInfos.php`, where a missing developer id still produced the "identifiants
utilisateurs" message. If RomM surfaces the distinction, prefer passing
ScreenScraper's own text through rather than interpreting which credential set is at
fault.

Re-measured end to end on 2026-08-05, running a real scan for every permutation, which
confirms the caveat and sharpens it. The two endpoints blame different credential sets
for the same configuration, in the same scan:

| credentials | `ssuserInfos.php` (account check) | `jeuInfos.php` (per ROM) |
| --- | --- | --- |
| valid dev + valid user | 200 | 200, ROMs identified |
| valid dev + **wrong user** | **403** "identifiants **utilisateurs**" | 200, ROMs still identified, charged at the unauthenticated cap (`730/10000` rather than `/100000`) |
| **wrong dev** + valid user | 200, then 403 a second later | **403** "identifiants **développeur**" |
| **missing dev** + valid user | 200, then 403 a second later | **403** "identifiants **développeur**" |

Two things follow. Passing ScreenScraper's own text through is the only honest option:
the same run produces both wordings depending on which endpoint asked. And a wrong user
password does not stop scraping at all, which is why the account check must report the
problem without disabling the provider.

Fixing item 1 alone would have reduced this entire investigation to a single log line.

---

*Edit: item 4 originally claimed RomM's 401 message named the wrong cause. That was
wrong. ScreenScraper's error table has separate Description and Cause columns, and
the CPU saturation RomM reports is the documented Cause for that row. The remaining
problem is narrower: the description half is missing, not incorrect. Rewritten
above.*

*Edit: added the measured permutation table and the requirement that the account
check must not disable ScreenScraper. Both come from running a real scan per
credential permutation: a 403 during the account check can be transient (#4124),
and treating it as final costs a working scan all of its ScreenScraper metadata.*

Spinnich:
Fixes #4122

**Description**

When ScreenScraper refused the configured credentials, RomM logged `403, message='Forbidden'` and carried on. Every ROM was saved without ScreenScraper metadata, the "ScreenScraper quota:" line never appeared, and nothing anywhere mentioned credentials. The report behind #4122 took hours to diagnose, most of it spent looking at the wrong thing (the account's quota), because that is where the silence pointed.

The cause was ordering. `_request()` already had the right message: `LOGIN_ERROR_CHECK` matches the `Erreur de login` string in the response body and raises a clear credentials error. But the check sat _after_ `res.raise_for_status()`, and ScreenScraper sends that string with **HTTP 403**, which raised first. The check was dead code for the exact case it was written for, in both the first attempt and the retry, and 403 had no branch of its own, so it fell to the catch-all `log.error(err)`.

This PR:

- **Reads the body before raising the status.** A refusal explains itself in the body, so a 403 no longer aborts the attempt with a bare "Forbidden" and loses the one line that says what is wrong.
- **Says which of the two credential sets was refused**, which is the one thing an operator needs in order to know where to look:

  ```
  ScreenScraper rejected your user account credentials. Check SCREENSCRAPER_USER and SCREENSCRAPER_PASSWORD.
  ScreenScraper rejected the RomM developer credentials.
  ```

  The developer message deliberately does not name its variables. They are only semi-protected (readable from any published image), and nothing should send an end user hunting for them; naming the set is enough to point troubleshooting at a bad build.
- **Attributes the refusal from the endpoint, not from ScreenScraper's wording.** `ssuserInfos.php` is the only endpoint that checks the account password, so a refusal anywhere else can only be about the developer credentials. ScreenScraper's own text is not a reliable signal by itself: with a bad developer id, `ssuserInfos.php` still answers `Vérifier les identifiants utilisateurs`. Its text is appended to the message rather than interpreted, so a bug report still carries the provider's exact words.
- **Trips a breaker, like the daily-quota path already does.** The failure is logged once per set, the remaining requests short-circuit instead of paying a round trip per ROM to be told the same thing, and the three scan-path lookups in `ss_handler` swallow it so the scan degrades to the other providers instead of dying.
- **The account check reports, but never arms that breaker.** ScreenScraper refuses a developer id it accepted a minute earlier (#4124) while the scraping endpoints keep answering, so letting `prime_account_limits()` leave the breaker armed cost a working scan all of its ScreenScraper metadata. Only a request the scan actually needs can take the provider out. This is #4123's rule, applied to the breaker introduced here; the quota breaker still needs it.
- **Makes `prime_account_limits()` warn when the call returns empty**, not just when it raises. Several errors are swallowed into an empty response, which used to leave a scan with no limits and nothing said about it.
- **Gives HTTP 401 both halves of its documented meaning.** ScreenScraper's error table splits that row into a description ("API fermé pour les non membres ou les membres inactifs") and a cause ("Le Serveur est saturé (utilisation CPU>60%)"). RomM reported the cause and dropped the description, so a lapsed member read a 401 as a passing server-side problem. (Item 4 of the issue originally claimed the message named the wrong cause; the issue has been corrected.)
- **Collapses the duplicated request block.** The two attempts each carried their own copy of the status ladder, and the copies had drifted: 401 and 403 were handled differently between them. Both now go through one `_attempt()` and one `_handle_client_error()`.

**Files modified**

| File | What changed |
| --- | --- |
| `backend/adapters/services/screenscraper.py` | New `SSCredentialSet`, `CREDENTIAL_DETAILS` and `ScreenScraperCredentialsError`. `_credential_set()` attributes a refusal; `_reject_credentials()` reports it once and records the set in `_ScanState.credentials_rejected`, so the requests that short-circuit repeat the same message. `_attempt()` holds the single request path and reads the body before `raise_for_status()`; `_handle_client_error()` is the one status ladder both attempts share. `_error_message()` condenses the provider's reply to a single line and masks anything credential-shaped. `prime_account_limits()` warns on an empty result as well as on a raise, and clears the breaker so the account check can never disable the provider. The 401 message carries the documented description and cause. |
| `backend/handler/metadata/ss_handler.py` | New `_is_provider_exhausted()` covers both the exhausted quota and the refused credentials. The three scan-path lookups (`lookup_rom`, `get_rom`, `get_rom_by_id`) use it, so a scan falls back to the other providers. `get_matched_roms_by_name` deliberately does not, so a manual search surfaces the error. |
| `backend/tests/adapters/services/test_screenscraper.py` | New `TestCredentialErrors` (12 tests: each set attributed to the right endpoint, a scraping refusal never blamed on the account, the developer variables never named, body read before status, provider wording carried through, credentials masked, no retry, a refusal on the retry attempt, later requests short-circuiting, logged once, cleared by the next scan) and `TestApiClosedForAccount` (401 carries both halves). Three new priming tests: an empty result, rejected credentials, and the account check never taking the provider out. Two pre-existing login-error tests retargeted to the new status and message. |
| `backend/tests/handler/metadata/test_ss_handler.py` | New `TestScreenScraperCredentialFallback` pins the fallback at each of the three scan-path call sites. |

**Testing**

- Full backend suite green, run serially: 2750 passed, 2 skipped.
- `trunk fmt && trunk check` clean on all four files.
- **A real scan per credential permutation**, against the live API on a mock platform with ScreenScraper as the only metadata source. `ssuserInfos.php` was probed immediately before each scan, because developer-credential enforcement flaps (#4124) and the scan result has to be read against what ScreenScraper actually did:

  | credentials | account check | scan outcome |
  | --- | --- | --- |
  | valid dev + valid user | 200 | 2/2 identified, quota logged |
  | valid dev + **wrong user** | 403 | ERROR naming `SCREENSCRAPER_USER` / `SCREENSCRAPER_PASSWORD`, **2/2 still identified**, quota line shows the unauthenticated cap (`1000/10000`, not `/100000`) |
  | **wrong dev** + valid user | 200, then 403 during priming | account ERROR, breaker cleared, scan proceeds; `jeuInfos.php` then refuses and the second ERROR names the RomM developer credentials, arming the breaker on the request that mattered |
  | **wrong dev** + **wrong user** | 403 | as above |
  | **missing dev** + valid user | 200, then 403 during priming | as above |

  Before the account-check rule was added, rows 3 and 5 produced zero ScreenScraper metadata on an account whose scraping worked. That is what the rule exists to prevent.
- Non-403 failures (checked with a 400) still return an empty response unchanged.

**Things worth a closer look**

1. **The outbound status is 403, and deliberately not 401.** It follows the existing SS-426 mapping (blacklisted application version): both mean ScreenScraper refused RomM itself and both need an operator change, unlike the transient failures that map to 503 or the budget ones that map to 429. A 401 would have been the obvious reading, but `frontend/src/services/api/index.ts` treats any 401 response as an expired RomM session: it clears `romm_session` and redirects to the login page. The pre-existing (unreachable) credential path used 401, so simply making it reachable would have logged users out mid-search.
2. **A manual name search now fails instead of quietly returning less.** `endpoints/search.py` gathers the providers without `return_exceptions`, so a raised credentials error takes the whole search with it rather than returning the IGDB/MobyGames results with ScreenScraper silently missing. That is the existing behaviour for an exhausted daily quota, a blacklisted version and an offline API, so this is consistent rather than new, and the user gets an actionable message instead of an unexplained gap. Happy to catch it in `get_matched_roms_by_name` instead if you would rather the other providers still answer.
3. **A bad developer id logs two errors, and the first one blames the account.** `ssuserInfos.php` refuses with the account wording whichever set is actually wrong, so the account check cannot tell them apart; the scraping request that follows can, and does. Both lines are in the permutation runs above. Suppressing the first would mean staying silent about a check that genuinely failed, and suppressing the second would lose the accurate attribution, so both stay.
4. **The 403 branch in `_handle_client_error()` is a fallback, not the main path.** `_attempt()` converts a 403 while it still has the body, so it produces the better message. The branch in the ladder catches a 403 arriving any other way, so the hole this PR closes cannot reopen if the body read ever moves.
5. **The breaker is process-wide until the next scan resets it.** Since the credentials are `Final` constants read from the environment at import, a restart is genuinely what clears it, and a manual search hitting the short circuit still raises the full message rather than failing silently.
6. **The body is now read on every request, including failures.** No added cost on the success path (the previous code already called `res.text()` before `res.json()`, and aiohttp caches the body between them); error bodies are a single short line.
7. **`_error_message()` masks credential-shaped substrings** with the same `SENSITIVE_KEYS_REGEX` the log formatter uses. The provider's text now reaches the API response as well as the log, and the credentials travel in the query string, so it should not be echoed back verbatim on trust.

**AI assistance disclosure**

This change was written with AI assistance (Claude Code). The AI wrote the tests first, then the implementation, ran the full backend suite, and ran the live permutation scans reported above. I directed the design decisions (the outbound status code, attributing the refusal to a credential set without naming the developer variables, and requiring that the account check never disable the provider) and reviewed the diff.

**Checklist**

### DIFF
--- a/backend/adapters/services/screenscraper.py
+++ b/backend/adapters/services/screenscraper.py
@@ -29,21 +29,6 @@ from utils import get_version
 from utils.context import ctx_aiohttp_session
 from utils.rate_limiter import ConcurrencyLimiter, RateLimiter
 
-LOGIN_ERROR_CHECK: Final = "Erreur de login"
-
-# ScreenScraper's error bodies are a single short line of plain text. Anything
-# longer, or marked up, is a page rather than a message.
-SS_MAX_ERROR_MESSAGE: Final[int] = 200
-
-# The one endpoint that checks the account password. The scraping endpoints
-# authorize on the developer credentials alone and never look at it.
-SS_ACCOUNT_ENDPOINT: Final = "ssuserInfos.php"
-
-# ScreenScraper names the developer credentials in the refusal itself, but only
-# from the scraping endpoints; the account endpoint blames the account whichever
-# set is actually at fault.
-SS_DEVELOPER_ERROR_CHECK: Final = "développeur"
-
 # ScreenScraper occasionally returns malformed JSON with unescaped backslashes in
 # text fields (e.g. game synopses), which the strict parser rejects with
 # "Invalid \escape" and discards the whole response. Match any backslash that is
@@ -193,7 +178,14 @@ class _ScanState:
     and media downloads go through the limiters without a service at all.
     """
 
+    # The account allowances read from the most recent response, if any. They govern
+    # how fast we may scrape (threads and requests per minute), how much of the
+    # daily quota is left, and how slowly media will download.
     account_limits: SSAccountLimits | None = None
+
+    # The one-shot advisories that are logged once per scan, so the log is not
+    # flooded with the same warning for every ROM. They are reset at the start of
+    # a scan so the next scan can report them again.
     logged_worker_advisory: bool = False
     logged_low_quota_warning: bool = False
 
@@ -250,20 +242,21 @@ def _error_message(body: str) -> str:
     if message.startswith("<"):
         return ""
 
-    return redact_sensitive(message[:SS_MAX_ERROR_MESSAGE])
+    # ScreenScraper's error bodies are a single short line of plain text. Anything
+    # longer, or marked up, is a page rather than a message.
+    return redact_sensitive(message[:200])
 
 
 def _credential_set(url: str, message: str) -> SSCredentialSet:
-    """Work out which credential set a refusal is about.
+    """Work out which credential set a refusal is about."""
 
-    ScreenScraper says so itself when a scraping endpoint refuses. When it does
-    not, the endpoint settles it: only the account endpoint checks the account
-    password, so a scraping endpoint can only be refusing the developer set.
-    """
-    if SS_DEVELOPER_ERROR_CHECK in message.lower():
+    # ScreenScraper names the developer credentials in the refusal itself, but only
+    # from the scraping endpoints; the account endpoint blames the account whichever
+    # set is actually at fault.
+    if "développeur" in message.lower():
         return SSCredentialSet.DEVELOPER
 
-    if SS_ACCOUNT_ENDPOINT in url:
+    if "ssuserInfos.php" in url:
         return SSCredentialSet.USER
 
     return SSCredentialSet.DEVELOPER
@@ -343,7 +336,7 @@ def get_account_limits() -> SSAccountLimits | None:
     return _state.account_limits
 
 
-def _parse_int(value: object, *, minimum: int = 0) -> int | None:
+def _parse_ss_int(value: object, *, minimum: int = 0) -> int | None:
     """Read one of the account's numeric fields, ignoring absent or junk values."""
     try:
         parsed = int(str(value))
@@ -355,15 +348,19 @@ def _parse_int(value: object, *, minimum: int = 0) -> int | None:
 
 def _read_account_limits(ssuser: SSUser) -> SSAccountLimits:
     return SSAccountLimits(
-        max_threads=_parse_int(ssuser.get("maxthreads"), minimum=1),
-        max_requests_per_minute=_parse_int(ssuser.get("maxrequestspermin"), minimum=1),
-        max_requests_per_day=_parse_int(ssuser.get("maxrequestsperday"), minimum=1),
-        requests_today=_parse_int(ssuser.get("requeststoday")),
-        max_ko_requests_per_day=_parse_int(
+        max_threads=_parse_ss_int(ssuser.get("maxthreads"), minimum=1),
+        max_requests_per_minute=_parse_ss_int(
+            ssuser.get("maxrequestspermin"), minimum=1
+        ),
+        max_requests_per_day=_parse_ss_int(ssuser.get("maxrequestsperday"), minimum=1),
+        requests_today=_parse_ss_int(ssuser.get("requeststoday")),
+        max_ko_requests_per_day=_parse_ss_int(
             ssuser.get("maxrequestskoperday"), minimum=1
         ),
-        ko_requests_today=_parse_int(ssuser.get("requestskotoday")),
-        max_download_speed_kbps=_parse_int(ssuser.get("maxdownloadspeed"), minimum=1),
+        ko_requests_today=_parse_ss_int(ssuser.get("requestskotoday")),
+        max_download_speed_kbps=_parse_ss_int(
+            ssuser.get("maxdownloadspeed"), minimum=1
+        ),
     )
 
 
@@ -598,7 +595,7 @@ class ScreenScraperService:
     ) -> None:
         self.url = yarl.URL(base_url or "https://api.screenscraper.fr/api2")
 
-    async def _attempt(self, url: str, request_timeout: int) -> dict:
+    async def _attempt_request(self, url: str, request_timeout: int) -> dict:
         """Make one request, and read the account allowances riding along on it.
 
         A refusal explains itself in the body, so the body is read before the
@@ -621,7 +618,7 @@ class ScreenScraperService:
                 timeout=ClientTimeout(total=request_timeout),
             )
             res_text = await res.text()
-            if LOGIN_ERROR_CHECK in res_text:
+            if "Erreur de login" in res_text:
                 raise _reject_credentials(url, _error_message(res_text))
 
             try:
@@ -653,7 +650,7 @@ class ScreenScraperService:
             raise ScreenScraperCredentialsError(_state.credentials_rejected)
 
         try:
-            return await self._attempt(url, request_timeout)
+            return await self._attempt_request(url, request_timeout)
         except aiohttp.ServerTimeoutError:
             # Retry the request once if it times out
             pass
@@ -676,15 +673,15 @@ class ScreenScraperService:
             return {}
 
         try:
-            return await self._attempt(url, request_timeout)
+            return await self._attempt_request(url, request_timeout)
         except aiohttp.ServerTimeoutError as err:
             log.error(err)
             return {}
         except aiohttp.ClientResponseError as err:
             if err.status == http.HTTPStatus.TOO_MANY_REQUESTS:
-                # Refused twice in a row: the pacing is behind the account's
-                # per-minute budget. Surface it so the ROM is reported as
-                # skipped instead of quietly saved without our metadata.
+                # The pacing is behind the account's  per-minute budget.
+                # Surface it so the ROM is reported as skipped instead
+                # of quietly saved without our metadata.
                 raise ScreenScraperRateLimitError() from err
 
             return _handle_client_error(url, err)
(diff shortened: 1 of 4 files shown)


Spinnich:
Not changing this here @greptileai, for a few reasons.

In production the two operations named can't share the state: scans are enqueued to RQ and run in the worker process (`start_bin_rq_worker`), manual searches run in the gunicorn web process, and `_ScanState` is a module global. The overlap only exists under `DEV_MODE`, where `scan_handler` calls `scan_platforms` inline rather than enqueueing.

Within one process the sharing is correct rather than a leak. The credentials are process-wide `Final` constants read from the environment at import (`config/__init__.py`), so two operations cannot be using different ones; a refusal observed by either is accurate for the other. A manual search inheriting the breaker raises the accurate message instead of issuing a request ScreenScraper has already refused, which is what the breaker is for. The reverse direction costs one request before it re-arms, and re-checking at the start of a scan is intentional.

This also isn't introduced here: `daily_quota_exhausted` lives in the same `_ScanState` with identical process-global semantics. The breaker lifecycle across processes is #4123 (item 4 is exactly "in the web process the flag is never cleared"), and a fix there should cover both breakers together rather than one of them in this PR.

gantoine (approved):

PR #4134, opened 2026-08-05
feat(retroachievements): flag which ROM version RA actually supports

sdornan:
Sorry for opening so many PRs in such quick succession. Just wanting to fix and improve things as I import my library. I intend for this one to be the last one for the time being! Also, a future enhancement I am considering is showing the RA badge or a trophy on the game list. Let me know your thoughts on that!

---

**Description**

RetroAchievements only unlocks achievements for the exact dump it has hashed, but nothing in RomM surfaced which of a game's versions that was. Picking between `Star Fox 64 (USA)` and `Star Fox 64 (USA) (Rev 1)` meant guessing, or opening each one in turn.

`ra_id` can't answer this. On the Hasheous path it comes from the matched game's metadata list (`hasheous_handler.py`), so it identifies the *game*, and every sibling of a title resolves to the same id. A per-version marker driven by it would light up on every row.

So this records the per-file fact instead:

- `RAHandler.hash_is_known` asks RetroAchievements' own hash list (the `ra_hashes_v2.json` index built from `get_game_list(..., include_hashes=True)`), never another provider.
- `scan_handler` now asks it for **every** ROM on an RA platform. Previously a Hasheous-supplied `ra_id` short-circuited the RA lookup entirely, which is how the per-file answer got lost.
- The answer persists as `roms.ra_hash_match` (migration `0108`), nullable so "never checked" stays distinct from "checked, and RA doesn't have it".

Two surfaces read it:

- **Version switcher** shows the RA logo on the versions RA knows, so the choice is visible where it's made.
- **Metadata tab / Verification** and the header badge now read the same field, falling back to Hasheous' `ra_match` only when RA was never asked. A definite *no* from RA outranks a Hasheous hit, since deferring there would promise achievements for a file RA has never seen. `_filter_by_verified` encodes the same precedence so the library filter can't drift from the badges.

Also included: the platform hash list is now memoised per platform with a short TTL. It was previously re-read and re-parsed from disk once per ROM during a scan, and it runs to several megabytes on large systems.

Note that `ra_hash_match` starts as `NULL`, so existing libraries show nothing new until those ROMs are rescanned or metadata-refreshed.

**AI assistance disclosure**

This PR was written with AI assistance (Claude Code). The investigation, implementation, migration, and tests were produced in a session with an agent, and reviewed by me before opening. Per `CONTRIBUTING.md`, disclosing this in full: the agent wrote the code and tests; I directed the approach, caught that `ra_id` was game-level rather than per-file, and chose to persist the flag at scan time rather than compute it per request.

**Checklist**

Verification run locally: `vue-tsc` clean, `npm run build` clean, 279 v2 vitest tests, 1902 backend tests across `tests/handler` and `tests/endpoints`, both i18n scripts, and ruff/black/prettier/eslint on the touched files. Not yet verified in a browser against a live library, so the screenshot below is a faithful mock rather than a capture (see below).

image-1785972754387

Composed mock, not a live capture: built from the real `ra.png` asset and the generated v2 theme tokens, with the CSS copied verbatim from `RMenu`, `RMenuItem`, and `VersionSwitcher`. The RA-marked row is the base USA dump, which is the version RetroAchievements actually hashed for Star Fox 64. The light pane also shows the mark alongside the existing default-version bookmark.

The mark is a bare 14px logo rather than the gallery's tiled provider chip: the menu panel is `width: max-content` with no cap, so anything wider grows the whole dropdown. At 14px it costs exactly what the bookmark already in that slot costs.

### DIFF
--- a/backend/alembic/versions/0108_roms_ra_hash_match.py
+++ b/backend/alembic/versions/0108_roms_ra_hash_match.py
@@ -1,0 +1,39 @@
+"""Record whether RetroAchievements knows a ROM's own hash
+
+`ra_id` is the RetroAchievements *game*, and every version of a game resolves
+to the same one: on the Hasheous path it comes from the matched game's metadata
+list, so a bad dump and a good dump of the same title are indistinguishable by
+it. Achievements, though, only unlock when the file's RA hash is in RA's hash
+list, which is a per-file fact RomM computed during scans and then threw away.
+
+This column keeps it. NULL means never checked -- no RA hash, a platform RA
+doesn't cover, or a ROM last scanned before this existed -- so it stays
+distinguishable from a checked-and-absent hash.
+
+Revision ID: 0108_roms_ra_hash_match
+Revises: 0107_roms_dedup_cover_index
+Create Date: 2026-08-05 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0108_roms_ra_hash_match"
+down_revision = "0107_roms_dedup_cover_index"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("roms", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("ra_hash_match", sa.Boolean(), nullable=True),
+            if_not_exists=True,
+        )
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("roms", schema=None) as batch_op:
+        batch_op.drop_column("ra_hash_match", if_exists=True)
--- a/backend/endpoints/responses/rom.py
+++ b/backend/endpoints/responses/rom.py
@@ -331,6 +331,7 @@ class RomSchema(BaseModel):
     md5_hash: str | None
     sha1_hash: str | None
     ra_hash: str | None
+    ra_hash_match: bool | None
 
     has_simple_single_file: bool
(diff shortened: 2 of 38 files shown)


sdornan:
Good catch, and it was worse than "likely dropped" - it made the feature's headline behaviour unreachable.

The metadata priority loop in `scan_rom` copies truthy values only:

```python
for key, field_value in handler_data.items():
    if field_value:
        rom_attrs[key] = field_value
```

So `ra_hash_match=True` persisted, but `False` was dropped and the column stayed `NULL`. `NULL` means never-checked and falls back to Hasheous, which means the "RetroAchievements has never seen this dump" branch could never fire in production. The precedence logic in `_filter_by_verified` and `romVerification.ts` was correct; the data never arrived. My tests missed it because they set the column directly rather than going through the scan.

Fixed in bb7731158 by applying the tri-state explicitly after that loop. It also can't be gated on `ra_id` the way the loop is (`available_sources` requires a truthy id field), since a ROM can carry the game's id from Hasheous while RA has never seen the file.

Added a `scan_rom` test parametrised over all three states. I confirmed it fails on the True and False cases with the fix disabled, so it's a real guard rather than decoration.

Also took the suppressed em-dash comment: fixed in the same commit. For what it's worth, that was the only em-dash in the added lines of this PR (checked with `git diff | grep '^+' | grep '—'`) - the other hits in those files are pre-existing.

sdornan:
Closing this one since I have other PRs open that I am more invested in.

Issue #3761, opened 2026-07-16
[Bug] Manual matching ROMs grid view truncated in Safari


lnlyssg:
**RomM version✱
5.0.0

**Describe the bug✱
When running a manual match of a ROM, the grid view shows slivers of artwork, rather than the whole picture. This seems to be Safari specific (tested in normal and private browsing with same issue on both) and looks OK in Chromium browsers.

**To Reproduce**

1. Click on a ROM
2. Click the three dots and Match ROM, then search
3. See the slivers of art like the screenshot below

**Expected behavior✱
Larger artwork to be displayed

**Screenshots**
<img width="888" height="474" alt="Image" src="https://github.com/user-attachments/assets/db7d10a4-e285-4afe-aa19-ef62e4205766" />

**Server (please complete the following information):**

- OS: Ubuntu, Docker
- Version: 24

**Client (please complete the following information):**

- Device: MacBook
- OS: macOS 26.5.2
- Browser: Safari
- Version: 26.5.2

**Additional context**

PR #4135, opened 2026-08-06
fix(v2): compute game-card cover width instead of deriving it

sdornan:
**Description**

The Match ROM dialog renders every result cover as a thin vertical sliver in Safari (Chromium is fine). Reported in #3761.

Non-hero `GameCard`s render at a fixed height with a natural width, and `.r-gc__art` took that width from `GameCover`'s `aspect-ratio` via `width: auto`. The width therefore depends on the aspect-ratio transferred size, and inside a shrink-to-fit ancestor (the dialog's `.match-grid__card-cell`) that resolution is circular. WebKit gives up and collapses the box to a few pixels, so `object-fit: cover` paints a one-pixel-wide slice of the cover.

The fix multiplies the height by the cover ratio instead:

```css
.r-gc:not(.r-gc--hero) .r-gc__art {
  width: calc(var(--r-card-art-h) * var(--r-cover-ratio, 0.6667));
}
```

Both values are already CSS variables on that element (`--r-cover-ratio` is set inline with the cover's measured ratio), so the width lands on the same number with nothing left for the engine to resolve.

Note on the earlier fix: 44ef5cf5 addressed this same report by making the grid cell a flex container. That changed *where* the width was resolved but left the aspect-ratio transfer in place, so the covers still collapse on 5.1.0. This PR removes the transfer, and I've reverted the now-inaccurate comment on `.match-grid__card-cell` (the `display: flex` itself stays, since that matches the gallery row).

**Verification**

- Confirmed against the failing case: injecting the new rule into a live 5.1.0 instance in Safari renders the covers correctly.
- No layout regression: I measured old vs. new width resolution across 5 cover ratios x 5 size tiers in 3 layout contexts (gallery row, match-dialog cell, list row with its `max-width` cap), in both WebKit and Chromium. All 110 boxes agree except 16 that differ by <= 0.02px, in the default tier where `--r-card-art-h` is itself fractional (236.98px). That is rounding order, not a layout change. WebKit and Chromium produce identical geometry under the new rule.
- Worth flagging for reviewers: headless WebKit does **not** reproduce this bug (the pre-44ef5cf5 markup renders correctly there), so the automated comparison above can only show equivalence. The fix itself was confirmed manually in Safari.

**AI assistance disclosure**

This PR was written primarily by Claude Code (Claude Opus 5): the diagnosis, the code change, the cross-engine comparison harness, and this description. The failing case and the confirmation that the fix resolves it in Safari were done by me manually.

**Checklist**

<sup>No unit test: this is a CSS-only layout fix whose failure mode reproduces only in real Safari, which is not in the test matrix (`playwright.config.ts` runs Chromium only).</sup>

### DIFF
--- a/frontend/src/v2/components/GameCard/GameCard.vue
+++ b/frontend/src/v2/components/GameCard/GameCard.vue
@@ -576,9 +576,14 @@ function onStaticKeydown(e: KeyboardEvent) {
   display: flex;
   flex-direction: column;
 }
-/* Width follows the cover (overrides GameCover's base `width: 100%`). */
+/* Width follows the cover (overrides GameCover's base `width: 100%`).
+   Multiplied out instead of leaning on GameCover's `aspect-ratio`: a
+   width derived from the aspect-ratio alone is circular inside a
+   shrink-to-fit ancestor (the match-dialog cell), and WebKit collapses
+   that to a few pixels. `--r-cover-ratio` is set inline on this element
+   with the cover's measured ratio, so it wins over the card-level one. */
 .r-gc:not(.r-gc--hero) .r-gc__art {
-  width: auto;
+  width: calc(var(--r-card-art-h) * var(--r-cover-ratio, 0.6667));
 }
 /* Label fills the cover's width and ellipsises, without widening the card. */
 .r-gc:not(.r-gc--hero) .r-gc__label {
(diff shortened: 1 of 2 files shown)


sdornan:
Before:
Screenshot 2026-08-05 at 9 38 21 PM

After:
Screenshot 2026-08-05 at 9 38 37 PM

gantoine (approved):

PR #4136, opened 2026-08-06
feat(v2): search platforms by slug, folder, category and family

### LINKED ISSUE #4108: [UI/UX] Search platform by their short names
When an UI element allows to search per platform (e.g. the selector from the upload rom menu and the library setup wizard), it should also find them by their short names.

A good start is the platform slugs list https://docs.romm.app/latest/platforms/supported-platforms/#platform-slugs-coverage

It does not cover every abbreviation though. The common rom file extensions should also be considered (e.g. sfc for Super Nintendo/Super Famicom).

I'm sure there are also a lot of regional nicknames and shortenings. It'd be nice to have a way to add those later.

It would also benefit from a bit of fuzziness: I never remember if there is a space in "Mega Drive" and always type "Megadrive", which yields no results.

Spinnich:
**Description**

Resolves #4108 

Search on the v2 `/platforms` page only ever matched a platform's **display name**. That
misses the identifiers users actually know their platforms by: the slug they see in URLs
and provider ids, the folder the platform lives in on disk, and the two group-by axes the
list view already exposes as its own columns (category and family).

Typing `psx`, `n64`, `nintendo` or `portable console` now finds the platform you mean.

**What changed**

- The `filtered` computed in `PlatformsIndex.vue` matches the term against
  `display_name`, `slug`, `fs_slug`, `category` and `family_name` instead of the display
  name alone. Same trim / lowercase / `includes` shape as before, just widened.
- **Category is matched twice, on purpose**: both the raw stored value
  (`portable_console`) and the prettified label the UI actually renders
  (`Portable console`, via the existing `prettifyPlatformCategory`). Typing what you see
  on screen has to work, and so does typing what the rest of the app stores.
- `category` and `family_name` are nullable, so they're skipped when absent.
- The search placeholder now names the searchable fields, matching the style already used
  on the `/search` page (`rom.search-placeholder` → "Search by name, filename, hash…"):
  **"Search by name, slug, folder, category…"**. New key `platform.search-placeholder`,
  translated in all 18 locales.

This is a client-side filter over data the platforms store already holds. No API change,
no new requests, no backend change.

**Files modified**

| File | Why |
| --- | --- |
| `frontend/src/v2/views/PlatformsIndex.vue` | New `searchFields()` helper builds the haystack per platform; `filtered` matches any field; placeholder bound to the new key. |
| `frontend/src/v2/views/PlatformsIndex.test.ts` | Search term made mockable (hoisted `searchState` behind the `useTileSearchUrl` mock) + a `describe("search")` block covering each field. |
| `frontend/src/locales/*/platform.json` (18 files) | New `search-placeholder` key, translated per locale using each locale's existing terms for slug / folder / category. |

**Testing notes**

Tests were written first, against the old implementation: 6 of them failed (slug, folder,
raw category, prettified category, family, case/whitespace) before the change and pass
after. The new `describe("search")` block covers:

- match on display name (the pre-existing behaviour, as a regression anchor)
- match on slug / folder name / raw category / prettified category label / family name
- case- and whitespace-insensitivity
- the "no results" empty state
- platforms with `null` category and family (the guard that keeps
  `prettifyPlatformCategory` from being handed `null`)

Fixtures are built so each field is provable on its own, e.g. `psx` appears only as an
`fs_slug` and `n64` only as a `slug`, so nothing can pass by matching through the name.

Verified locally:

- `npm run test` — 676 passed, 56 files
- `npm run typecheck` — clean
- `python3 src/locales/check_i18n_locales.py` and `check_i18n_sorted.py` — both pass
- `trunk fmt && trunk check` — clean
- Browser: light + dark, grid + list layout, 1440px and 390px viewports, typing real
  keystrokes. Confirmed against real library data that family matching works where the
  family name is nowhere in the display name (typing `sony` returns the PlayStation
  platforms, `nec` returns the PC Engine ones), and that the `?search=` query param still
  round-trips so a searched view stays bookmarkable.

**Anything a reviewer should look at**

- **Category matched twice** (raw + prettified) is deliberate, not a leftover. Drop the
  raw value if you'd rather only the visible label be searchable.
- **`platform.search-platform` is intentionally kept**, not renamed: v1's
  `components/common/Navigation/PlatformsDrawer.vue` still consumes it, and v1 is frozen.
  It can be deleted along with v1.
- **The header count is unchanged** and still reports the number of visible platforms
  rather than the number of search results. That's pre-existing behaviour and matches the
  other index views, so I left it alone, but say the word if it should follow the filter.
- Only the v2 page is affected. v1's `filteredPlatforms` store getter is untouched.

**AI assistance disclosure**

This PR was written primarily by Claude Code (Claude Opus 5), under my direction and
review: the tests, the implementation and the locale strings were AI-generated, and I
reviewed and verified them locally in the browser before opening this PR. This PR
description was also drafted by Claude Code.

**Checklist**

image

image

### DIFF
--- a/frontend/src/v2/views/PlatformsIndex.vue
+++ b/frontend/src/v2/views/PlatformsIndex.vue
@@ -277,11 +277,25 @@ onMounted(() => {
   }
 });
 
+// Fields a search term is matched against. Beyond the display name,
+// these are the identifiers a user already knows a platform by: the
+// slug used in URLs and provider ids, the folder it lives in on disk,
+// and the two group-by axes the list view exposes as columns. Category
+// contributes both the raw IGDB value and the prettified label, since
+// the label is what the UI shows ("Portable console") while the raw
+// value is what the rest of the app stores ("portable_console").
+function searchFields(p: Platform): string[] {
+  const fields = [p.display_name, p.slug, p.fs_slug];
+  if (p.category) fields.push(p.category, prettifyPlatformCategory(p.category));
+  if (p.family_name) fields.push(p.family_name);
+  return fields;
+}
+
 const filtered = computed<Platform[]>(() => {
   const term = searchTerm.value.trim().toLowerCase();
   if (!term) return visiblePlatforms.value;
   return visiblePlatforms.value.filter((p) =>
-    p.display_name.toLowerCase().includes(term),
+    searchFields(p).some((field) => field.toLowerCase().includes(term)),
   );
 });
 
@@ -479,7 +493,7 @@ const groupedBuckets = computed<Bucket[] | null>(() => {
         :group-by-items="platformGroupByItems"
         :segment-filters="segmentFilters"
         show-search
-        :search-placeholder="t('platform.search-platform')"
+        :search-placeholder="t('platform.search-placeholder')"
         @update:group-by="groupBy = $event"
         @update:layout="layout = $event"
         @update:sort-dir="gridSortDir = $event"
--- a/frontend/src/v2/views/PlatformsIndex.vue
+++ b/frontend/src/v2/views/PlatformsIndex.vue
@@ -277,11 +277,7 @@ onMounted(() => {
   }
 });
 
-// Fields a search term is matched against. Beyond the display name,
-// these are the identifiers a user already knows a platform by: the
-// slug used in URLs and provider ids, the folder it lives in on disk,
-// and the two group-by axes the list view exposes as columns. Category
-// contributes both the raw IGDB value and the prettified label, since
+// Category contributes both the raw IGDB value and the prettified label, since
 // the label is what the UI shows ("Portable console") while the raw
 // value is what the rest of the app stores ("portable_console").
 function searchFields(p: Platform): string[] {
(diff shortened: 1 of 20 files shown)


frontend/src/v2/views/PlatformsIndex.vue
```diff
@@ -277,11 +277,25 @@ onMounted(() => {
   }
 });
 
+// Fields a search term is matched against. Beyond the display name,
+// these are the identifiers a user already knows a platform by: the
+// slug used in URLs and provider ids, the folder it lives in on disk,
+// and the two group-by axes the list view exposes as columns. Category
+// contributes both the raw IGDB value and the prettified label, since
+// the label is what the UI shows ("Portable console") while the raw
+// value is what the rest of the app stores ("portable_console").
```
gantoine:
```suggestion
// Category contributes both the raw IGDB value and the prettified label, since
// the label is what the UI shows ("Portable console") while the raw
// value is what the rest of the app stores ("portable_console").
```

gantoine (approved):

frontend/src/v2/views/PlatformsIndex.vue
gantoine: @Spinnich you gotta get your agent to drop the self-explanitory comments lol

PR #4138, opened 2026-08-06
fix(v2): drop a random pick that lands after leaving the gallery

### LINKED ISSUE #4114: [Bug] "Random ROM" can still pull you into a game after you leave the gallery for a non-gallery page
**RomM version**

Reported on 5.1.1-beta.1. Re-verified on current `master` (960e18df7), where it
still happens. v2 UI only.

**Describe the bug**

The "Random ROM" button on a platform or collection page takes a moment to
decide which game you get, because it has to ask the server for a pick. On a
big library, or over a slow connection, that can take a few seconds.

If you get bored and click away during those few seconds, the pick can still
land and yank you into a game page you never asked for.

PR #4113 fixed the common half of this: if you switch to a *different platform
or collection*, the pick is now thrown away. But the app only notices that you
have "left" a gallery when you open another gallery. If you go somewhere that
isn't a gallery, such as the home page, the settings page, or an individual
game's page, nothing tells the random pick that you have moved on, so it still
fires and navigates you away from whatever you were doing.

A related, smaller version of the same problem: if the request *fails* instead
of succeeding after you have moved on, you get a "Couldn't pick a random ROM"
error toast on a page that has nothing to do with the gallery you left.

**To Reproduce**

Easiest to hit on a platform with a lot of games, or on a slow connection.

1. Open a platform with a large number of games (or a collection).
2. Click the "Random ROM" button in the header.
3. Immediately, before the game page opens, click through to the home page or
   the settings page.
4. Wait a couple of seconds.
5. The random pick finishes and drops you onto a game's detail page, throwing
   away whatever you had navigated to.

For the error-toast variant, make the request fail (for example stop the
backend) between steps 2 and 4: the "Couldn't pick a random ROM" toast appears
on the home page.

**Expected behavior**

Once you navigate away from the gallery you clicked "Random ROM" in, the pick
should be abandoned, wherever you went. It should not change the page you are
on, and it should not show an error about a gallery you already left.

- OS: Unraid
- Version: RomM 5.1.1-beta.1

- Device: Desktop PC
- OS: Windows 11
- Browser: Brave

**Additional context**

Verified on current `master` (960e18df7) with a headless browser against a dev
stack, by holding `/api/roms/random` open for 12s and navigating in-app during
that window. Results:

| Started in | Navigated to | Result |
| --- | --- | --- |
| Platform gallery | Home | pick landed, hijacked to `/rom/{id}` |
| Platform gallery | A game's detail page | pick landed, hijacked to a different `/rom/{id}` |
| Collection gallery | Home | pick landed, hijacked to `/rom/{id}` |
| Platform gallery, request forced to fail | Home | "Couldn't pick a random ROM" toast shown on the home page |
| Platform gallery | **Another platform gallery** | pick correctly discarded (this is the part #4113 fixed) |

That last row is the control: the guard works when you move between galleries,
and only when you move between galleries.

Hypothesised root cause. Both galleries decide whether a pick is stale by
comparing against the gallery that is *currently loaded* --
`galleryRoms.currentPlatform` in `frontend/src/v2/views/Gallery/Platform.vue:340`,
and the component-local `currentCollection` ref in
`frontend/src/v2/views/Gallery/Collection.vue:292`. The only thing that clears
those is `resetGallery()` (`frontend/src/v2/stores/galleryRoms.ts:290-292`), and
every one of its call sites runs while *another gallery* is loading
(`Platform.vue:271`, `Collection.vue:218`, `Search.vue:45`,
`MissingGamesSection.vue:270,282`). Neither view has an unmount hook, so leaving
for a non-gallery route clears nothing, the staleness check sees nothing wrong,
and both the navigation and the error toast go ahead.

Two plausible directions:

- Also treat the view being torn down, or the route no longer being that
  gallery, as "stale".
- Pass an `AbortSignal` into the pick request and cancel it when the view
  unmounts, which would additionally stop the pointless in-flight request. Note
  that `romApi.getRandomRom` (`frontend/src/services/api/rom.ts:420-439`) does
  not currently accept a `signal`, so this route means adding one first. An
  aborted request rejects, so the existing error toast would need to stay
  suppressed in that case.

Whichever way it goes, the fix needs to cover both `Platform.vue` and
`Collection.vue` -- they reproduce identically.

This was called out as a known residual in PR #4113 rather than fixed there, to
keep that change scoped to #4104.

*Investigation and root-cause analysis for this report were AI-assisted. The
reproduction and the file references above were checked against `master`, but
treat the proposed cause as a hypothesis rather than a verified fix.*

---

*Edited to correct two details that were already out of date when this was
filed: the pick is now a single `/api/roms/random` request (PR #4097), not the
two `getRoms` calls originally described, and the `AbortSignal` suggestion has
been updated because `getRandomRom` takes no `signal`. Added the verification on
current `master` and the client/server details.*

Spinnich:
**Description**

Fixes #4114.

Clicking "Random ROM" in a platform or collection gallery starts a request and
navigates when it resolves. #4113 added a staleness check, but it asks "is a
different gallery loaded now?", not "is the user still where they clicked from?".
Leaving for a route that isn't another gallery (Home, Settings, a game's detail
page) never runs `resetGallery()` and never clears `Collection.vue`'s local
`currentCollection` ref, so the captured scope id still matches: the pick lands,
`router.push` fires, and the user is yanked off the page they navigated to. A
pick that *fails* after they moved on raises a "Couldn't pick a random ROM" toast
on the unrelated page.

The fix treats the view being torn down as stale and ORs that into the existing
`stale()` closure, two lines per view. Both symptoms (the hijack and the stray
toast) go through the same gate, so both are covered.

Worth a closer look from a reviewer:

- **`onBeforeUnmount` is only correct while nothing wraps these views in
  `<keep-alive>`.** There is none in v2 today (`AppLayout.vue` renders a bare
  `<router-view name="v2" />`), but if keep-alive is ever introduced these guards
  need `onDeactivated` too. That is the one way this can silently regress.
- **This also closes a gallery-to-gallery case #4113 missed.** `Collection.vue`
  keeps `currentCollection` in a component-local ref, and regular / virtual /
  smart collections are three separate route records, so hopping between kinds
  unmounts the view and leaves that ref pointing at the old collection forever.
  The id check could never catch it; the unmount flag does.
- **The in-flight request is not aborted.** Adding a `signal` to
  `getRandomRom` was considered and rejected: it touches a shared API service the
  Home `RandomPickWidget` also calls, the server has already run the query by the
  time an abort lands, and an aborted request rejects anyway, so the `!stale()`
  gate on the error toast is required either way. Noted as a possible follow-up.

No new strings, no API or schema change, no migration.

**Checklist**

Tested locally:

- Four view-level tests added (two per view), written first and confirmed failing
  against the unfixed code (4 failed / 15 passed) with `push` called as
  `{ name: "rom", params: { rom: 42 } }` and `snackbarError` called with
  `platform.random-rom-error`. Full frontend suite green afterwards: 56 files,
  671 tests.
- `npm run typecheck`, `npm run build` and `trunk fmt && trunk check` all clean.
- Driven in a real browser with Playwright against the dev stack, stalling
  `/api/roms/random` and navigating in-app before releasing it: **10/10 scenarios
  pass with the fix, 4/10 with it stashed**, the six failures being exactly the
  reported defect. The four controls (leaving for another gallery, staying put,
  the empty-library pick, an error raised while still in the gallery) pass in
  both runs, which is what makes the harness discriminating rather than just red.
  Both light and dark themes and both mouse and keyboard activation were covered.
- One pre-existing e2e failure (`game-media-files.spec.ts:44`, the known v2
  manuals/RDropzone defect) reproduces identically with these changes stashed, so
  it is not a regression from this PR.

The clearest pair is the stray toast: a failed pick after navigating to Home.

| Before | After |
| --- | --- |
| C-failed-pick-no-toast-on-home|C-failed-pick-no-toast-on-home|

And the hijack itself, leaving a platform gallery for Home:

| Before | After |
| --- | --- |
|A-platform-to-home-dark|A-platform-to-home-dark|

---

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written primarily by Claude Code (Claude Opus 5)
under my direction. AI assistance covered the investigation and root-cause
analysis, the code change, the four unit tests, and the Playwright harness used
for browser verification. I reviewed the diff, decided the approach (the unmount
flag over an `AbortSignal`) and the scope, and ran and checked the verification
myself.

### DIFF
--- a/frontend/src/v2/composables/useIsAlive/index.test.ts
+++ b/frontend/src/v2/composables/useIsAlive/index.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { effectScope } from "vue";
+import { useIsAlive } from "./index";
+
+describe("useIsAlive", () => {
+  it("starts alive and flips on scope disposal", () => {
+    const scope = effectScope();
+    const alive = scope.run(() => useIsAlive())!;
+
+    expect(alive.value).toBe(true);
+    scope.stop();
+    expect(alive.value).toBe(false);
+  });
+
+  it("keeps sibling scopes independent", () => {
+    const a = effectScope();
+    const b = effectScope();
+    const aliveA = a.run(() => useIsAlive())!;
+    const aliveB = b.run(() => useIsAlive())!;
+
+    a.stop();
+    expect(aliveA.value).toBe(false);
+    expect(aliveB.value).toBe(true);
+  });
+});
--- a/frontend/src/v2/composables/useIsAlive/index.ts
+++ b/frontend/src/v2/composables/useIsAlive/index.ts
@@ -1,0 +1,21 @@
+// useIsAlive — tracks whether the owning effect scope is still active, so an
+// async handler that resolves late can tell it's talking to a dead component.
+//
+// Replaces the manual pattern:
+//   let unmounted = false;
+//   onBeforeUnmount(() => { unmounted = true; });
+//
+// Cleanup uses `onScopeDispose` (not `onBeforeUnmount`) so it also works from
+// a non-component effect scope, e.g. inside another composable.
+//
+// Note: VueUse's `useMounted` is not a substitute — it only flips to `true` on
+// mount and never back to `false` on teardown.
+import { onScopeDispose, shallowRef, type ShallowRef } from "vue";
+
+export function useIsAlive(): ShallowRef<boolean> {
+  const alive = shallowRef(true);
+  onScopeDispose(() => {
+    alive.value = false;
+  });
+  return alive;
+}
--- a/frontend/src/v2/composables/useIsAlive/index.ts
+++ b/frontend/src/v2/composables/useIsAlive/index.ts
@@ -1,15 +1,5 @@
 // useIsAlive — tracks whether the owning effect scope is still active, so an
 // async handler that resolves late can tell it's talking to a dead component.
-//
-// Replaces the manual pattern:
-//   let unmounted = false;
-//   onBeforeUnmount(() => { unmounted = true; });
-//
-// Cleanup uses `onScopeDispose` (not `onBeforeUnmount`) so it also works from
-// a non-component effect scope, e.g. inside another composable.
-//
-// Note: VueUse's `useMounted` is not a substitute — it only flips to `true` on
-// mount and never back to `false` on teardown.
 import { onScopeDispose, shallowRef, type ShallowRef } from "vue";
 
 export function useIsAlive(): ShallowRef<boolean> {
--- a/frontend/src/v2/views/Gallery/Collection.test.ts
+++ b/frontend/src/v2/views/Gallery/Collection.test.ts
@@ -268,6 +268,48 @@ describe("Collection view random rom", () => {
     expect(push).toHaveBeenCalledWith({ name: "rom", params: { rom: 7 } });
   });
 
+  // Issue #4114: leaving for a route that is not another gallery leaves
+  // `currentCollection` set, so the id check alone cannot tell that the user
+  // walked away.
+  it("drops a pick that lands after the user left the gallery", async () => {
+    let resolvePick: (value: { data: SimpleRom }) => void = () => {};
+    getRandomRom.mockReturnValueOnce(
+      new Promise((resolve) => {
+        resolvePick = resolve;
+      }),
+    );
+
+    const wrapper = await mountView();
+    await wrapper.get("button.random").trigger("click");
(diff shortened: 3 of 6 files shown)


Spinnich:
@gantoine Uh, I think GitHub had a stroke running these tests. As far as I can tell, I don't have permission to fire them off manually myself (which is fine). If something does fail once they run I'll address it (but they should pass as I ran them locally)!

gantoine (approved):

PR #4142, opened 2026-08-06
fix(v2): show a running scan on pages loaded mid-scan

### LINKED ISSUE #4141: [Bug] A page loaded while a scan is running doesn't show the scan
**RomM version**

`master` @ 960e18df7 (v2 UI)

**Describe the bug**

If you open (or reload) a page while a library scan is already running, the frontend doesn't know a scan is in progress. `/scan` renders its "no scan started" empty state with an armed **Start scan** button, and the navbar's scanning pill stays hidden — while the server is happily scanning.

There are two independent causes, and they compound.

**1. Scan state is never reconciled with the server — it's built up purely from socket events**

`scanning` lives only in the `scanning` Pinia store (`frontend/src/stores/scanning.ts`) and is only ever set by the socket handlers in `frontend/src/v2/composables/useScanLifecycle/index.ts`. Nothing asks the backend "is a scan running right now?" on load.

So any fresh page load — hard refresh, new tab, second device, closing and reopening the app — starts at `scanning: false, scanningPlatforms: []`. `Scan.vue` gates the whole live panel on `scanning || scanningPlatforms.length > 0`, so you get the empty state until a *new* event happens to land while you're sitting there.

**2. `scan:update_stats` is the one event that fires continuously, and it doesn't flip `scanning`**

```js
useSocketEvent<ScanStats>("scan:update_stats", (stats) => {
  scanningStore.setScanStats(stats);   // no setScanning(true)
});
```

`scanning_platform` and `scanning_rom` both call `setScanning(true)`; `update_stats` doesn't. That matters because the other two are sparse:

- `scan:scanning_platform` fires **once per platform** — minutes apart on a large library.
- `scan:scanning_rom` only fires for ROMs the scan actually adds (plus restored-missing ones). On a quick/update scan over an already-scanned library, whole platforms go by emitting nothing but batched `scan_stats.increment(scanned_roms=len(skipped_rom_ids))` — i.e. only `update_stats`.

So the common case is: load the page mid-scan, receive stats events, store them silently, and still render "no scan started". The counters tick behind a hidden panel.

Knock-on symptom: when the scan finishes you get the "Scan completed successfully" snackbar for a scan the page never showed.

**To Reproduce**

1. Start a scan of a reasonably large library (`/scan` → **Start scan**).
2. While it's running, hard-refresh the page (or open the app in a second tab / on another device).
3. On `/scan`: the live panel shows the empty state and **Start scan** is enabled, as if nothing were running.
4. On any other page: the navbar scanning pill is absent.

Easiest to hit with a **quick or update scan over an already-scanned library**, where nearly every ROM is skipped and only `update_stats` is emitted — there the page can stay wrong indefinitely. With an all-new library `scan:scanning_rom` fires constantly, so the page self-corrects within ~100 ms and the bug is easy to miss.

> [!IMPORTANT]
> This does **not** reproduce with `DEV_MODE=true`, which is what `DEVELOPER_SETUP.md` tells contributors to set. Under `DEV_MODE` the scan handler short-circuits and runs the scan inline in the web process (`backend/endpoints/sockets/scan.py`, in `scan_handler`), so there is no RQ job at all. Reproducing (and fixing) this needs a production-shaped setup: `DEV_MODE=false` plus a real RQ worker.

**Expected behavior**

A page that loads while a scan is running should show the scan: the live status bar with its counters, the abort button, the config locked, and the navbar pill on other routes.

The server already knows — `GET /api/tasks/status` returns the running job with `task_type: "scan"`, `status: "started"` and its `meta.scan_stats` — so the frontend just needs to ask on load.

The per-platform live log (`scanningPlatforms`) is a separate matter: it only ever exists in the originating tab's memory, since the job meta carries stats but not the platform/ROM list. A reconciled page can show correct status and counters, with the panel list filling in from the next `scan:scanning_platform` onward. Preserving the full log across reloads would need the backend to persist it.

**Server**

- OS: macOS (local dev), reproduced against MariaDB + Valkey with `DEV_MODE=false` and an RQ worker
- Version: `master` @ 960e18df7

**Client**

- Device: desktop
- Browser: Chromium
- UI: v2

**Additional context**

Scope note: `/api/tasks/status` requires the `tasks.run` scope, which is the same scope the `scan` socket handler gates on (`reject_unauthorized_scan`), so anyone who could have started the scan can also read it back. Users without it would stay purely event-driven.

Diagnosed and reported with AI assistance (Claude Code); a fix is on the way.

sdornan:
**Description**

Fixes #4141 — a page loaded while a scan is running now shows the scan instead of its empty state.

Scan state was built up purely from socket events, so a tab that loaded mid-scan (hard refresh, second tab, another device) started at `scanning: false`: `/scan` rendered "no scan started" with an armed **Start scan** button, and the navbar pill stayed hidden while the server was scanning.

Two independent causes, both in `useScanLifecycle`:

**1. `scan:update_stats` didn't flip `scanning` on.** `scanning_platform` and `scanning_rom` both called `setScanning(true)`; `update_stats` only stored the stats. That matters because it's the one event a scan emits continuously — `scanning_platform` fires once per platform, and `scanning_rom` only for ROMs the scan actually adds, so a quick/update scan over a settled library emits nothing else for long stretches. One line, and it closes the recovery window for the common case.

**2. Nothing reconciled with the server.** New `reconcileWithRunningScan()` reads `GET /api/tasks/status` on install and, if a task with `task_type: "scan"` and `status: "started"` is there, seeds `scanning` and `scanStats` from its `meta.scan_stats`. That makes a refresh correct immediately rather than eventually.

Details:

- Gated on the `tasks.run` scope — the same scope `reject_unauthorized_scan` requires of the `scan` socket handler — so anyone who could have started this scan can read it back, and users without it don't eat a 403 on every page load.
- Runs once per login, keyed on user id, since the watch also fires when the user object is replaced by an unrelated profile update.
- Two races guarded: a `scan:done`/`done_ko` landing while the request is in flight won't resurrect a finished scan, and live stats already in the store win over the job's staler snapshot.
- `Scan.vue` needed no change — it already gates the config lock, the CTA, the abort button and the live panel on `scanning`.

**Not fixed here:** the per-platform live log (`scanningPlatforms`) only ever exists in the originating tab's memory — the job meta carries stats, not the platform/ROM list. A reconciled page gets the correct status bar, counters, progress bar and abort button, with the panel list filling in from the next `scan:scanning_platform` onward. Preserving the full log across reloads would need the backend to persist it, which felt out of scope.

> [!IMPORTANT]
> **Don't verify this with `DEV_MODE=true`.** Under `DEV_MODE` the scan handler short-circuits and runs the scan inline in the web process, so there's no RQ job, `/api/tasks/status` reports nothing, and the reconcile correctly finds nothing to reconcile with. Since `DEVELOPER_SETUP.md` tells contributors to set `DEV_MODE=true`, it's easy to test there and wrongly conclude the reconcile is broken. The `update_stats` fix still works in DEV_MODE; the reconcile needs `DEV_MODE=false` plus a real RQ worker.

**Testing**

Unit tests (6 new cases) cover: stats-event-implies-scanning, reconcile from a running job, staying idle when no scan is running, skipping without the scope, and both races.

Manually verified against a production-shaped stack — `DEV_MODE=false`, real RQ worker, MariaDB + Valkey, 4500-ROM library — with a scan running in the worker and `/api/tasks/status` confirmed reporting `task_type: scan, status: started`:

- Hard reload on `/scan` mid-scan → "Scanning" with pulse, live counters (1/3 platforms, 230/4500 ROMs), abort button, config locked, Start button showing its spinner, log streaming.
- Hard reload on `/` (Home) mid-scan → navbar pill present: "Scanning 221 / 4500", `aria-label="Scanning library. Click to view"`, links to `/scan`, keyboard-focusable.
- Both `r-v2-light` and `r-v2-dark`.

Full frontend suite green: typecheck, 673 tests / 57 files, build, Trunk.

Two gaps worth naming: I didn't exercise the skip-path scan (only `update_stats` firing) in the browser — the scan I tested was all-new ROMs, where `scanning_rom` fires constantly — nor did I do an A/B revert to attribute the behaviour change visually. Both are covered by the unit tests. No touch/gamepad or responsive sweep either, since the change adds no markup or CSS; it only makes existing components render in a case where they previously didn't.

**Checklist**

#### AI assistance disclosure

Written with AI assistance (Claude Code, Opus 5), and substantially so: the diagnosis, both code changes, the unit tests, the local verification stack, and this description were all produced by the agent under my direction and review.

### DIFF
--- a/frontend/src/v2/composables/useScanLifecycle/index.test.ts
+++ b/frontend/src/v2/composables/useScanLifecycle/index.test.ts
@@ -1,7 +1,7 @@
 import { mount } from "@vue/test-utils";
 import { createPinia, setActivePinia } from "pinia";
-import { beforeEach, describe, expect, it, vi } from "vitest";
-import { defineComponent, nextTick } from "vue";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { defineComponent, reactive } from "vue";
 import type { ScanStats } from "@/__generated__";
 import taskApi from "@/services/api/task";
 import storeScanning from "@/stores/scanning";
@@ -34,17 +34,23 @@ vi.mock("@/services/api/platform", () => ({
   },
 }));
 
-const authState = {
+// Reactive so the composable's `watch` on `authStore.user` fires when the
+// user arrives after install, which is the real flow: AppLayout installs
+// while /users/me is still in flight.
+const authState = reactive({
   user: { id: 1, oauth_scopes: ["tasks.run"] } as {
     id: number;
     oauth_scopes: string[];
   } | null,
-};
+});
 vi.mock("@/stores/auth", () => ({
   default: () => authState,
 }));
 
 const getTaskStatus = vi.mocked(taskApi.getTaskStatus);
+
+/** Drain pending microtasks so the reconcile's promise chain has settled. */
+const flushPromises = () => new Promise((resolve) => setTimeout(resolve, 0));
 
 function makeStats(overrides: Partial<ScanStats> = {}): ScanStats {
   return {
@@ -77,9 +83,13 @@ function runningScanTask(stats: ScanStats | null) {
 }
 
 // The lifecycle uses `inject` and `onScopeDispose`, so it needs a host
-// component instance.
+// component instance. Tracked so `afterEach` can unmount it: the auth state
+// is reactive and shared, so a leaked host would keep watching it and
+// reconcile again during later tests.
+let host: ReturnType<typeof mount> | null = null;
+
 function install() {
-  mount(
+  host = mount(
     defineComponent({
       setup() {
         installScanLifecycle();
@@ -102,6 +112,11 @@ describe("installScanLifecycle", () => {
     authState.user = { id: 1, oauth_scopes: ["tasks.run"] };
   });
 
+  afterEach(() => {
+    host?.unmount();
+    host = null;
+  });
+
   it("treats a stats event as proof a scan is running", () => {
     install();
     const scanning = storeScanning();
@@ -119,13 +134,33 @@ describe("installScanLifecycle", () => {
     } as never);
 
     install();
-    await nextTick();
-    await nextTick();
+    await flushPromises();
 
     const scanning = storeScanning();
     expect(scanning.scanning).toBe(true);
     expect(scanning.scanStats.scanned_roms).toBe(40);
     expect(scanning.scanStats.total_roms).toBe(100);
+  });
+
+  it("reconciles once the user arrives after install", async () => {
+    // The real flow: AppLayout installs while /users/me is still in flight,
+    // so `user` is null at install and the watch has to catch the arrival.
+    authState.user = null;
+    getTaskStatus.mockResolvedValue({
+      data: [runningScanTask(makeStats({ scanned_roms: 7 }))],
+    } as never);
+
+    install();
+    await flushPromises();
+    expect(getTaskStatus).not.toHaveBeenCalled();
+
+    authState.user = { id: 1, oauth_scopes: ["tasks.run"] };
+    await flushPromises();
+
+    expect(getTaskStatus).toHaveBeenCalledTimes(1);
+    const scanning = storeScanning();
+    expect(scanning.scanning).toBe(true);
+    expect(scanning.scanStats.scanned_roms).toBe(7);
   });
 
   it("stays idle when no scan job is running", async () => {
@@ -134,8 +169,7 @@ describe("installScanLifecycle", () => {
     } as never);
 
     install();
-    await nextTick();
-    await nextTick();
+    await flushPromises();
 
     expect(storeScanning().scanning).toBe(false);
   });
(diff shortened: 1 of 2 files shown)


sdornan:
Thanks — took all three.

**`running.meta` optional chaining** (`index.ts:254`): done. The generated type has `meta` as required and the endpoint always builds it, so this shouldn't be reachable, but it's a JSON boundary and the guard is free. A missing snapshot now means "no counters yet" rather than a throw. (For the record it wouldn't have crashed the app — the throw was inside the `.then` and the chained `.catch` would have swallowed it — but it would have silently skipped seeding the stats.)

**Missing coverage for the deferred-auth flow** (`index.ts:235`): good catch, and the most valuable of the three — that's the actual production path, since AppLayout installs while `/users/me` is in flight. Added `reconciles once the user arrives after install`, asserting nothing fires while `user` is null and that the reconcile happens exactly once on arrival.

Worth noting *why* it was uncovered: the auth mock was a plain object, so `watch(() => authStore.user)` could never fire on change — the existing tests only passed via `{ immediate: true }` with a user already present. Making the mock reactive then exposed a second latent problem: `install()` never unmounted its host, so watchers from earlier tests stayed alive and reconciled again on the now-shared reactive state (the new test failed with 3 calls instead of 1). Now unmounted in `afterEach`, which also exercises the `onScopeDispose` cleanup.

**`nextTick` → `flushPromises`** (`index.test.ts:123`): agreed, the double-`nextTick` worked by accident of microtask count. Replaced with an explicit `flushPromises` helper throughout.

Full suite green: typecheck, 674 tests / 57 files, build, Trunk.

gantoine (approved):

PR #4146, opened 2026-08-07
fix(metadata): identify archives and folder ROMs by the right file

### LINKED ISSUE #3742: [Bug] Hasheous and Playmatch hash matching broken for multi-file archives since 4.9.0 (composite hash sent, but they index by largest file)

Spinnich:
**Description**

Fixes #3742.

Since 4.9.0 an archive's `RomFile` carries a **composite** digest streamed across every
entry in the archive (`a170649fe`, PR #3412). The metadata handlers were never updated
and kept reading their lookup digests straight off that record, so they silently
switched from "the ROM inside the zip" to "a hash of the whole archive". No ROM database
indexes that composite, so any zip with more than one member (arcade/MAME sets, but also
an ordinary ROM zipped next to a readme or patch) stopped matching by hash and fell back
to filename matching. Single-file zips were unaffected, which is why most libraries
never noticed.

This is a re-land of #4059, which implemented exactly this and was closed unmerged on
2026-08-02. The review request on it is still open and no formal review was ever
submitted, so I've kept half A deliberately close to that diff to keep the re-land
legible.

**What changed**

*Which hashes (the 4.9.0 regression).* New `RomFile.lookup_hashes` decides what a ROM
database should be queried with: `chd_sha1_hash` for CHDs, else the largest
`archive_members` entry, else the file's own digests. Hasheous and Playmatch both read
it. CHD behaviour for Hasheous is byte-identical to before; only the archive case moves.

*Which file (pre-existing, not a 4.9.0 regression).* Playmatch picked its file with
`next(file for file in files if file.file_size_bytes > 0)` — the *first* file with bytes,
not the largest, with no `is_top_level` filter. `iter_files` is a bare `os.walk` with no
sort and `Rom.files` has no `order_by`, so for a multi-disc folder ROM the file Playmatch
was asked about was arbitrary filesystem order and could differ between two machines
scanning the same library. It was also frequently the `.m3u` playlist, which carries no
game data and can only ever miss.

To be clear about scope: this half is **not** part of the 4.9.0 regression. 4.8.1 carries
the same three lines verbatim (`ca58028eb`, "playmatch should just use the first file",
2025-07-10). It's in here because fixing only the hash selection would leave multi-disc
folder ROMs still failing on Playmatch, and the two questions are the same question.
A new `_select_lookup_file` replaces both copies of that expression (lookup and
suggestion) with the largest top-level file, tie-broken on path.

**Worth looking at hardest**

1. `_select_lookup_file`'s `is_top_level` filter is a behaviour narrowing, not just a
   tie-break. A folder ROM whose files all live in subdirectories (`Game/Disc 1/game.bin`)
   now gets **no** Playmatch lookup at all, where before it got one against an arbitrary
   nested file. Hasheous and ScreenScraper already drop those libraries, so this aligns
   the three, but it is a real loss for that layout. Happy to add a last-resort fallback
   to the largest sizeable file if you'd rather keep the old reach.

2. `submit_manual_match_suggestion` now contributes the *member's* digests alongside the
   *archive's* `fileName`/`fileSize`. Harmless for lookups (a hash match outranks
   `FileNameAndSize`), but the contributed triple is internally inconsistent. Sending the
   member's name and size would be the fuller fix; I left it out of scope here and it's
   worth its own issue. Relatedly, the suggestion no longer falls back to
   `rom.md5_hash`/`rom.sha1_hash` when no file qualifies — it just returns. Contributing
   nothing to a public index beats contributing a composite that is a digest of no single
   file.

3. Playmatch now looks CHDs up by `chd_sha1_hash` instead of the container digests. This
   is right in principle and matches Hasheous, but Playmatch's CHD behaviour is the one
   leg I couldn't confirm with a live match.

**Deliberately unchanged**

ScreenScraper keeps sending the archive's own composite digests. Its `jeuInfos` lookup
falls back to `romnom` (the filename), which for an arcade zip is the set name, so those
still resolve; moving it would shift match results across many libraries. There's a new
test pinning that, so a future change there has to be a choice rather than a side effect.
RetroAchievements is a separate RAHasher pipeline and is untouched.

**No migration, no re-hash.** `archive_members` has been populated since 4.9.0 and
already carries the member digests, so this reads correct data off rows already in the
database.

**Correction to the issue body:** #3742 claims the arcade fixtures demonstrate the
Playmatch half. They don't — Playmatch does not index MAME sets at all and returns
`NoMatch` for them on *either* hash. An ordinary No-Intro ROM zipped with a readme is
what reproduces the Playmatch half; both are covered below.

---

> **AI assistance disclosure.** This PR was investigated, implemented, and written
> primarily by Claude Code — the root-cause analysis, the code, the tests, and this
> description. I directed the work, made the scope calls, reviewed every line, and ran
> the verification below myself.

**Checklist**

<details>
<summary>Verification</summary>

**Live lookups against the real Hasheous and Playmatch APIs**, driving the real
`FSRomsHandler.get_rom_files()` (only `RAHasherService.calculate_hash` patched out) into
the real `lookup_rom`:

| Fixture | md5 sent before | md5 sent now | Hasheous | Playmatch |
| --- | --- | --- | --- | --- |
| `aerofgt.zip` (8 members) | `c5347ceb…` | `694853ae…` | **'Aero Fighters' hasheous 48064 / igdb 3147** (was HTTP 404) | `NoMatch` (expected) |
| `ga2.zip` (19 members) | `c6ed4e03…` | `0cd4f9af…` | **'Golden Axe: The Revenge of Death Adder' 51687 / igdb 5565** (was 404) | `NoMatch` (expected) |
| SMB3 zip (1 member, control) | `e883e0d4…` | identical | 'Super Mario Bros. 3' 25611 (unchanged) | `SHA1` igdb 1068 (unchanged) |
| SMB3 + readme (2 members) | `9f018a5b…` | `e883e0d4…` | **'Super Mario Bros. 3'** (was 404) | **`SHA1` igdb 1068** (was `NoMatch`) |

The two arcade `NoMatch`es are not a failure: Playmatch doesn't index MAME on either
hash. The SMB3+readme row is the fixture that demonstrates the Playmatch half.

Multi-disc folder ROM (`Metal Gear Solid (USA) (Rev 1)/`, two CHDs plus an `.m3u`):
picks `Disc 1`, not the playlist; sends the CHD's disc-data SHA-1, not the container's;
identical selection across 4 list orderings (order-dependent on master).

**Tests.** 15 new tests. All 7 in `tests/models/test_rom.py` failed on master with
`ImportError: cannot import name 'LookupHashes'`; the Hasheous payload test failed
sending `compositemd5/sha1/crc`; 6 of the 7 Playmatch tests failed on master (composite
hashes, `fileName='game.m3u'`, order-dependent payload, a nested file selected, a
composite POSTed as a suggestion). Edge cases covered: single-member archive unchanged,
`archive_members` `None` and `[]`, a member with a missing or `None` size, CHD
precedence, nested-file exclusion, and no-eligible-file.

Full backend suite: **2762 passed, 2 skipped**. `trunk fmt` + `trunk check --no-fix` on
all 7 files: no issues.

**No OpenAPI change** — proven, not assumed: dumped `app.openapi()` from a `master`
worktree and from this branch and diffed them. Identical, 26,574 lines each. So no
`npm run generate` and no frontend typecheck.

</details>

N/A — backend only. No UI, no i18n, no schema change. Evidence is the live-lookup table
above.

### DIFF
--- a/backend/models/rom.py
+++ b/backend/models/rom.py
@@ -5,7 +5,7 @@ import enum
 import re
 from datetime import datetime
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, TypedDict
+from typing import TYPE_CHECKING, Any, NamedTuple, TypedDict
 
 from sqlalchemy import (
     TIMESTAMP,
@@ -104,6 +104,14 @@ class RomArchiveMember(TypedDict):
     sha1_hash: str
 
 
+class LookupHashes(NamedTuple):
+    """The hashes a ROM database should be queried with."""
+
+    crc: str | None
+    md5: str | None
+    sha1: str | None
+
+
 class RomFile(BaseModel):
     __tablename__ = "rom_files"
 
@@ -164,6 +172,28 @@ class RomFile(BaseModel):
         from handler.filesystem import fs_rom_handler
 
         return fs_rom_handler.parse_file_extension(self.file_name)
+
+    @cached_property
+    def lookup_hashes(self) -> LookupHashes:
+        """The hashes to identify this file by against a ROM database.
+
+        Often not the file's own digests: a CHD is indexed by the disc data
+        embedded in its header, and a multi-file archive by its largest member
+        (the ROM itself, next to readmes and the like). The file's own hashes
+        cover the container, which no database holds.
+        """
+        if self.chd_sha1_hash:
+            return LookupHashes(crc=None, md5=None, sha1=self.chd_sha1_hash)
+
+        if self.archive_members:
+            largest = max(self.archive_members, key=lambda m: m.get("size") or 0)
+            return LookupHashes(
+                crc=largest.get("crc_hash"),
+                md5=largest.get("md5_hash"),
+                sha1=largest.get("sha1_hash"),
+            )
+
+        return LookupHashes(crc=self.crc_hash, md5=self.md5_hash, sha1=self.sha1_hash)
 
     @cached_property
     def is_nested(self) -> bool:
--- a/backend/handler/metadata/hasheous_handler.py
+++ b/backend/handler/metadata/hasheous_handler.py
@@ -270,17 +270,12 @@ class HasheousHandler(MetadataHandler):
         # against any of them.
         data: list[dict] = []
         for file in filtered_files:
-            file_hashes: dict[str, str | None]
-            if file.chd_sha1_hash:
-                # CHD files are indexed by disc-data SHA1 only
-                # Raw file MD5/CRC are hashes of the container and won't match
-                file_hashes = {"shA1": file.chd_sha1_hash}
-            else:
-                file_hashes = {
-                    "mD5": file.md5_hash,
-                    "shA1": file.sha1_hash,
-                    "crc": file.crc_hash,
-                }
+            hashes = file.lookup_hashes
+            file_hashes: dict[str, str | None] = {
+                "mD5": hashes.md5,
+                "shA1": hashes.sha1,
+                "crc": hashes.crc,
+            }
 
             # Drop empty hashes and skip files that have none.
             file_hashes = {key: value for key, value in file_hashes.items() if value}
--- a/backend/handler/metadata/playmatch_handler.py
+++ b/backend/handler/metadata/playmatch_handler.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+from collections.abc import Iterable
 from enum import Enum
 from typing import Final, NotRequired, TypedDict
 
@@ -62,6 +63,21 @@ PLAYMATCH_LOOKUP_ROM_ATTRS: frozenset[str] = frozenset(
 PLAYMATCH_SUPPORTED_SOURCES: frozenset[str] = frozenset(
     {"igdb", "moby", "ss", "launchbox", "sgdb"}
 )
+
+
+def _select_lookup_file(files: Iterable[RomFile]) -> RomFile | None:
+    """The single file a ROM is identified by: the biggest top-level one, which
+    is how Hasheous and ScreenScraper choose theirs.
+
+    Equally sized files break the tie on path, so a two-disc set resolves to
+    disc 1. The scanner walks the filesystem unsorted, so without that two
+    machines scanning the same library ask about different files.
+    """
+    return min(
+        (file for file in files if file.file_size_bytes > 0 and file.is_top_level),
+        key=lambda file: (-file.file_size_bytes, file.full_path),
+        default=None,
+    )
 
 
 class GameMatchType(str, Enum):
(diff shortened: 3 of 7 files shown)


backend/models/rom.py
```diff
@@ -165,6 +173,28 @@ def file_extension(self) -> str:
...
+
+        if self.archive_members:
+            largest = max(self.archive_members, key=lambda m: m.get("size") or 0)
+            return LookupHashes(
+                crc=largest.get("crc_hash"),
+                md5=largest.get("md5_hash"),
+                sha1=largest.get("sha1_hash"),
+            )
+
+        return LookupHashes(crc=self.crc_hash, md5=self.md5_hash, sha1=self.sha1_hash)
```
gantoine, line 196: very clean approach

gantoine (approved):

PR #4149, opened 2026-08-07
refactor(i18n): name gallery filters by property instead of "Show X"

sdornan:
**Description**

The gallery's bool filters are tri-state (All / Yes / No), but seven of them were labelled with a verb prefix. That made "Show missing" set to *No* read as a double negative, and the shared "Show" prefix was really just the section heading hoisted into each row.

They're now named after the property being filtered, matching the `Has saves` / `Has save states` / `Has soundtrack` trio that already read correctly:

| Before | After |
| --- | --- |
| Show matched / Show unmatched | Matched / Unmatched |
| Show favourites | Favorite (en_US) · Favourite (en_GB) |
| Show versions | Has versions |
| Show playables | Playable in browser |
| Show missing | Missing from disk |
| Show verified | Hash verified |
| Show Retroachievements | Has RetroAchievements |

Two labels also named the wrong thing: `verified` is a DAT/hash match via Hasheous rather than generic verification, and `missing` is specifically `missing_from_fs`. Neither was guessable from the old label.

The section heading moves from **Show** to **Properties** so it sits parallel with the drawer's other two headings (Platforms, Tags), and `platform.show` is renamed to `platform.properties` to match its value. Wording is picked per locale rather than mapped from one word: CJK takes the native 属性 / 屬性 / 속성 over the katakana loan, which reads as a file-properties dialog.

This also corrects the RetroAchievements casing, which disagreed with every other string referencing it.

> [!NOTE]
> These keys are shared with the **v1** filter drawer, so v1's labels change too. Forking them into v2-only keys would mean 8 duplicate keys across 18 locales for a UI slated for deletion, so the shared rename seemed like the right call. v1 was verified visually.

A second commit fixes an unrelated dialect problem found while auditing: `en_US` carried British spellings (*favourites*, *catalogues*), `en_GB` carried American ones (*color*).

All 18 locales are updated in the same change. Where a translation had drifted from the English meaning it was corrected rather than carried over: `bg_BG` had *playables* as "изпълнимите" (executable), `de_DE` had *matched* as "zugewiesene" (assigned).

**Verification**

`vue-tsc`, 667/667 vitest, `npm run build`, ESLint, i18n parity + sort, and Trunk all pass. Manually verified in the browser: v2 light/dark at 1440px plus 390px and 320px, the v1 drawer in both themes, and live renders in fr/de/ja/hu/ru with no label wrapping or clipping in any locale.

**AI assistance disclosure**

Written with Claude Code (Opus 5), which did the full change: label selection, all 18 locale translations, the key rename, and verification. Reviewed by me before submission.

**Checklist**

- [x] I've updated relevant comments (n/a, no comments affected)
- [x] I've added unit tests that cover the changes (updated the existing `smartCollectionCriteria` assertion; no new logic to test)

#### Screenshots

image-1786117691456
image-1786117697399
image-1786117702801
image-1786117707456


Spinnich:
There's probably a few other spots in the system that use only `favourite` / `favourites` as an fyi for consistency. Probably out of scope, and likely tricky for the default standard collection that gets seeded during the first game that is favorited.

gantoine (approved):
excellent change

PR #4153, opened 2026-08-07
fix: refresh the v2 gallery when a ROM is edited or matched

sdornan:
> [!NOTE]
> This PR was written primarily by Claude Code (code, tests and this description). I reviewed the approach and directed the scope at each step.

**Description**

The v2 gallery ignored its own writes: editing or matching a ROM left the card showing the old name and cover.

The v2 gallery renders its cards from `galleryRoms.byPosition`, but every v2 write site only patched v1's `stores/roms`, whose `_allRoms` the gallery never reads. Editing or matching a ROM replaces it with a fresh object from the API response, so the card kept showing the pre-edit name and cover until its window happened to be refetched (navigating away and back, a filter change, a scan).

Optimistic toggles (favourite, playing status) appeared to work without this only by accident: they mutate the cached object in place, and the gallery holds that same reference.

`galleryRoms.update()` already existed for exactly this purpose and had no callers.

This adds `useRomSync` and routes the 20 v2 write sites through it:

- **`syncRom`** fans a write out to both stores. Used by the optimistic toggles and the asset tabs, which keep the cheap in-place update — invalidating the windows costs skeletons and the scroll position, too much for a favourite flip or a screenshot upload.
- **`applyRomWrite`** (edit / match dialogs) additionally refetches when an in-place swap would leave the list lying:
  - _Membership_ — any active filter forces a refetch. A match rewrites provider ids, name, and the genre / company / tag metadata behind half the drawer, so guessing which filters it moved would mean re-deriving backend predicates on the client. This is what makes "filter to games not matched in IGDB → match one → it leaves the list" work.
  - _Order_ — compares only the value the gallery is currently ordered by, so a rename under name-ascending moves the card while a summary-only edit leaves the gallery (and its scroll position) alone.
- **`refreshAfterUserStateChange`** covers favourite / status writes, guarded narrowly on just the filters those can move (Favourites collection, favourites filter, status filter). Notably the existing "drop it from the Favourites view" branches in `useFavoriteToggle` and `SelectionBar` check v1's `currentCollection`, which v2 never sets — so un-hearting a game while viewing Favourites left the card in place. The bulk bar reconciles once per action rather than once per selected ROM.

The one remaining `romsStore.update` in `useScanLifecycle` is deliberate: it's the v1 gallery branch, and the v2 gallery is already refreshed by `refreshGallery()` a few lines above.

v1 is untouched.

**Checklist**

**Testing**

11 unit tests in `useRomSync/index.test.ts`. Manually verified in a local instance (mock library, SNES + GBA), in both light and dark themes:

| Case | Result |
| --- | --- |
| Rename a ROM | Card shows the new title and cover immediately |
| Same rename with the fix reverted | Card stays stale — confirms the diagnosis |
| Rename under name-ascending, unfiltered | Card moves to its new position; AlphaStrip follows |
| Summary-only edit | Updates in place, no refetch, scroll preserved |
| Give a ROM an IGDB id while filtered to "not matched in IGDB" | Row disappears |
| Unmatch while filtered to matched | Row disappears |

**Not verified, needs a reviewer with provider credentials:** the real Match ROM search-and-pick flow. My test instance had no IGDB/ScreenScraper/MobyGames keys, so provider search returned nothing. I exercised the identical `applyRomWrite` path through the edit dialog with the same membership change (gaining an `igdb_id`), and the match dialog's call is a single unconditional line after `syncRom` — but the search flow itself is untested.

**Verification run:** frontend typecheck, 686 frontend tests, `npm run build`, the full backend suite (2798 passed, 2 skipped), and `trunk fmt && trunk check` — all clean. No locale, token, or `/lib` primitive changes, and no response-schema or route-signature change, so the i18n check, `build:tokens`, Storybook, and `npm run generate` don't apply.

**Out of scope: cross-client freshness.** Everything here is same-client. There's no Socket.IO event for ROM mutations, so a second open browser stays stale until it refetches on its own. That follow-up is built and waiting on [`claude/rom-mutation-broadcast`](https://github.com/sdornan/romm/tree/claude/rom-mutation-broadcast) — deliberately kept out of this PR because it adds a backend event contract, a new request header, and a change to the shared axios interceptor, which is a different blast radius from a frontend cache fix, and because it hasn't been verified with two live browsers yet. Happy to open it as a follow-up PR once this lands. This PR is most of its client-side groundwork: the listener just hands each refetched ROM to `applyRomWrite`.


sdornan:
**Resolved:** the broadcast has been split back out — this PR is frontend-only again. The work lives on [`claude/rom-mutation-broadcast`](https://github.com/sdornan/romm/tree/claude/rom-mutation-broadcast) and can follow as its own PR once this lands.

Splitting because that change has a different blast radius: a backend Socket.IO event contract, a new `X-Socket-Id` request header, and a change to the shared axios interceptor every v1 and v2 request goes through. It also hasn't been verified with two live browsers yet, and I'd rather that uncertainty didn't sit on top of a frontend cache fix that has been.

<details><summary>Earlier notes on the broadcast design</summary>

The two design points, for whenever it gets reviewed:

- **Ids, not serialized ROMs** — `rom_user` is scoped to the requesting user, so a shared payload would show every client the actor's playing status, rating and last-played. Each client refetches and resolves its own.
- **Echo suppression per connection, not per user** — two tabs of one account are two clients that each need the update, and on a single-user instance every client is the same user, so matching on the user would suppress the event everywhere.

</details>

sdornan:
Thanks both. Went through the four findings; three were valid and are fixed in the amended commit, one I'd like to scope out with reasoning.

**Fixed: `useRomSync` mock missing `refreshAfterUserStateChange`** (@copilot-pull-request-reviewer). Correct and worth catching. `useGameActions` destructures it, so it was `undefined` in tests. Nothing crashes today only because no test exercises `setStatus` / `favorite` yet, which made it a landmine rather than a failure. The mock now matches the composable's full surface.

**Fixed: em-dashes violate CLAUDE.md:64.** Also correct. Six added lines across the two new files, plus the commit message. Cleaned; I left the pre-existing em-dashes in the files I only touched lightly alone rather than reformatting unrelated comments.

**Fixed: last-played ordering stays stale** (@greptile-apps). Valid, and the interesting one. `removeFromContinuePlaying` clears `last_played` (a sort key) but went through `syncRom`, which never reorders. Worth noting the obvious fix does not work: swapping it to `applyRomWrite` would still miss it, because that call site mutates the cached ROM in place, so the before/after comparison sees one object and finds no change. Added `refreshIfOrderedBy(key)` for exactly this shape, where the caller names the field it wrote, and called it there. Two tests cover it.

**Scoping out: failed refresh leaves the gallery empty** (@greptile-apps). Real, but not introduced here and not fixable in the right place from this diff. `invalidateWindows()` followed by a `fetchInitialMetadata()` that only logs on failure is the pre-existing pattern in `galleryRoms.remove()` and `useScanLifecycle.refreshGallery()`; this PR reaches it from more places but doesn't create it. A proper fix needs new store state (the store tracks `failedWindows` but has no equivalent for a failed metadata bootstrap) plus a view-level retry or error surface, which is a bigger change than a cache-sync fix should carry. Happy to open a separate issue for it if a maintainer agrees that's the right split.

Re-verified after the changes: typecheck, 680 frontend tests, full backend suite (2798 passed, 2 skipped), and `trunk fmt && trunk check` all clean.

gantoine (approved):

PR #4177, opened 2026-08-10
perf(scan): narrow the scan skip-decision query

sdornan:
**Description**

A scan visits every entry in a platform, but only the entries it decides to scan need a full `Rom`. The per-batch lookup (`get_roms_by_fs_name`) loaded ORM entities, so **every ROM in the library** had all ten metadata JSON columns pulled out of the database and JSON-decoded, plus a joined `platform` and a joined `roms_metadata` view, purely to answer "does a row exist for this filename, and which sources matched it?".

For a quick scan of an unchanged library that is the whole per-ROM cost, since existing entries are skipped and nothing else touches them.

This splits the lookup in two:

- **`get_rom_scan_states(platform_id, fs_names)`** returns a narrow `RomScanState` row per entry. It selects columns rather than entities, so no `Rom` is constructed and the `lazy="joined"` loads never fire. The emitted SQL is now just:

  ```sql
  SELECT roms.id, roms.fs_name, roms.igdb_id, roms.moby_id, roms.ss_id, roms.sgdb_id,
         roms.ra_id, roms.launchbox_id, roms.hasheous_id, roms.tgdb_id,
         roms.flashpoint_id, roms.hltb_id, roms.gamelist_id, roms.libretro_id
  FROM roms WHERE roms.platform_id = ? AND roms.fs_name IN (?)
  ```

- **`get_roms_for_scan(ids)`** hydrates full rows, with `platform` eager-loaded, for only the entries being acted on. A quick scan of an unchanged library hydrates none, and an empty id list issues no query at all.

`should_scan_rom` now takes the narrow row, which is enforced by its type signature rather than by convention, so it cannot start depending on a full `Rom` by accident.

`is_identified` also gets a single definition (`has_metadata_match` / `IDENTIFYING_ID_FIELDS` in `models/rom.py`), shared by `Rom` and `RomScanState`, so adding a metadata source updates one tuple instead of two predicates that could silently disagree. `tgdb_id` and `sgdb_id` still do not count toward identification, unchanged.

**Impact**

Measured over 5000 fully matched ROMs (IGDB + SS + MobyGames + LaunchBox + RA, ~49 KB of metadata JSON each), MariaDB-shaped payloads on SQLite so the numbers isolate ORM hydration and JSON decode:

| | per ROM | 10k library | 100k library |
| --- | --- | --- | --- |
| before | 194 us | 1.9 s | 19.4 s |
| after | 3 us | 0.03 s | 0.3 s |

On top of that sits the database I/O this avoids: at ~49 KB per ROM, a 10k library was shipping roughly 490 MB per scan and a 100k library roughly 4.9 GB. That part is negligible for a local warm MariaDB and significant for a remote one, or one whose buffer pool is smaller than the `roms` table. The narrow query reads about 100 bytes per ROM and is served by the existing `idx_roms_platform_id_fs_name`.

`COMPLETE` and `HASHES` scans hydrate every row as before, so their total work is unchanged; they now do two queries per 200-ROM batch instead of one.

Scan behaviour is otherwise untouched: this changes only *when* a full row is loaded, not which ROMs get scanned.

**AI assistance disclosure**

This change was written with AI assistance (Claude Code). The agent did the investigation, the implementation, the benchmark, and the test updates; I reviewed the result.

**Checklist**

<sup>Verification: full backend suite green on a clean database (2820 passed, 2 skipped; `master` is 2819, the extra one being the new `is_identified` parity test). `trunk fmt` and `trunk check` clean on the changed files. `mypy` reports an identical error set before and after (all pre-existing in this tree). The emitted SQL for both queries is asserted, including that the decision query reads no metadata blob and joins no relationship.</sup>

<sup>One reviewer note: two expectations in `test_comprehensive_scenarios` changed. That parametrization passed a `Mock(spec=Rom)` even for its `rom_exists=False` rows, so the `NEW_PLATFORMS` and `QUICK` "no rom" cases never exercised the `None` path and asserted `False`. Passing a real `None` makes both `True`, which is what the neighbouring `test_new_platforms_scan_with_no_rom` and `test_quick_scan_with_no_rom` already assert.</sup>

Not applicable, backend only.


gantoine:
i see the improvents but since scans are async i dont agree with the extra complexity to make them faster, not worth the tradeoff

PR #4184, opened 2026-08-10
feat(v2): show IGDB ports on the game overview

sdornan:
**Description**

IGDB has always returned a `ports` list of related games, and the backend already fetches, stores and serves it (`igdb_handler.py` builds it with `game_type="port"`, and it ships on every ROM as `igdb_metadata.ports`). The v2 game overview just never rendered it, so the data was fetched and dropped on the floor.

This adds a **Ports** section to the Overview tab, alongside Expansions / DLC / Remakes / Remasters / Similar games:

- `GameDetails.vue` gains a `ports` computed off `igdb_metadata` and passes it to `OverviewTab`.
- `OverviewTab.vue` gains a `ports` prop, a new section, and folds `ports` into `hasRelated` so a game whose *only* related entries are ports now renders the block instead of nothing.

Ports go through the same `RelatedGamesGrid` / `RelatedGameCard` path as every other category, so they inherit the IGDB to RomM cross-reference for free: a port that's in your library shows the "Owned" pill and links to its RomM detail page, otherwise the card opens the IGDB page in a new tab.

**One thing worth a reviewer's attention:** the five existing category headings were hardcoded English strings in the template. Rather than add a sixth untranslated string, I moved all six to `rom.related-*` keys and translated them across all 18 locales. That's why the diff touches 18 locale files for an otherwise small feature. Happy to split that into its own PR if you'd prefer this one stay minimal.

**AI assistance disclosure**
<sup>Per CONTRIBUTING.md.</sup>

This PR was written with AI assistance (Claude Code). The change was AI-authored end to end: locating the unused `ports` field, the component wiring, the unit tests, and the non-English translations. I reviewed the result before opening the PR. Reviewers may especially want to sanity-check the translations of "Ports", "Remakes" and "Remasters" in their own language, since gaming terminology in these varies by locale and I'd rather a native speaker corrected them than have them ship wrong.

**Checklist**

<sub>On "tested locally": static verification is green (`npm run typecheck`, `npm run test` at 696 passing across 59 files, both `check_i18n_*.py` scripts, and `trunk check` via the pre-commit hook, with no new ESLint warnings against the baseline file). I did **not** verify it in a running browser, so the section spacing and the `mdi-swap-horizontal` icon choice have not been eyeballed against the real design in either theme. Flagging that explicitly rather than ticking the box.</sub>

None: see the note above about not having run this in a browser.


gantoine (approved):
simple change, nice

PR #4185, opened 2026-08-10
fix(i18n): translate pt_BR strings left in English

sdornan:
**Description**

A large number of `pt_BR` values were still the raw `en_US` source, so Brazilian Portuguese users saw English in the UI. This translates **159** of them.

Worst-affected areas:

| File | Translated | Notes |
| --- | --- | --- |
| `settings.json` | 101 | Permission groups, permission overrides, client API tokens, and the whole device-pairing flow |
| `rom.json` | 39 | Manual upload/scrape flow, soundtrack upload, saves/states sections |
| `logs.json` | 18 | The entire namespace was untranslated |
| `common.json` | 1 | `server-offline-retrying` |
| `patcher.json` | 1 | `validation-warning` |

Terminology reuses what `pt_BR` already established elsewhere rather than inventing new terms: `trilha sonora` (soundtrack), `escaneamento` (scan, the dominant form in `settings.json` at 10 vs 7), `área de transferência` (clipboard), `Excluir` (delete, 15 vs 6 over `Remover`), `Comunidade`, and `Meus saves` / `Meus states` to match the existing `screenshots-section-mine` → `Minhas capturas`.

**What is deliberately left unchanged**

Identical-to-English is not the same as untranslated. 46 values are already correct Brazilian Portuguese or established loanwords, and changing them would make the UI worse:

- Portuguese words that happen to match English: `Manual`, `Volume`, `Interface`, `Local`, `Material`, `Humor`, `Status`
- Established loanwords already used across `pt_BR`: `Firmware`, `Logs`, `Slug`, `Tag` / `Tags`, `Slot`, `Saves`, `Hashes`, `Widgets`, `Marquee`, `Fan art`, `Logo`, `Proxies`
- Scene/technical ROM categories kept per the existing file, which already translates only the translatable ones (`category-game` → `Jogo`, `category-prototype` → `Protótipo`) while keeping `Cheat`, `Hack`, `Mod`, `Patch`, `Demo`, `DLC`

**Dead keys found along the way (not translated)**

While scoping this I checked whether the strings are actually rendered. **109 of 1,673 `en_US` keys have no i18n call site anywhere in `frontend/src`** (~6.5%); 79 of those have no trace of their name at all. Nine fall in this locale's untranslated set, so I skipped them rather than polishing strings nothing displays:

`common.local`, `common.server-reconnected`, `platform.tab-firmware`, `rom.manual-files-upload-failed`, `settings.client-token-confirm-regenerate`, `settings.exclusions-item`, `settings.hidden-platforms-hint`, `settings.perm-own-only`, `settings.permission-group-hint`

Pruning all 109 across 18 locale directories is a separate change and not attempted here. Happy to open an issue if maintainers want it tracked.

**Verification**

```
python3 frontend/src/locales/check_i18n_locales.py   # ✅ All translations are complete!
python3 frontend/src/locales/check_i18n_sorted.py    # ✅ All locale files are sorted alphabetically!
```

`trunk` (incl. Prettier) passes on all 5 changed files. I also verified every one of the 1,673 `pt_BR` strings against `en_US` for interpolation safety: **0 placeholder-name mismatches**, so no `{n}` / `{error}` / `{count}` was dropped or renamed. Changes were applied by a script guarded to only touch values that still exactly equal the `en_US` source, so the diff is exactly 159 replaced lines with no reordering or formatting churn.

Not run: `npm run typecheck` / `npm run test`, because this worktree has no `node_modules`. Neither is affected by locale JSON content — vue-i18n glob-loads these files at runtime and nothing type-derives from them. CI will cover both.

**AI-assistance disclosure**
<sup>Per `CONTRIBUTING.md`.</sup>

This PR was written with AI assistance (Claude Code, Claude Opus 5). The AI did the full change: auditing which values still matched `en_US`, deciding which were genuine gaps versus legitimate loanwords, producing the Brazilian Portuguese translations, running the dead-key analysis, and running the validators and linter. **The translations are AI-produced and would benefit from a native pt-BR speaker's review** — flagging that explicitly given this locale previously received machine translations that introduced errors (see #4183, which fixes two Breton strings from that same source).

**Checklist**

<sup>No comments needed (locale data only). No unit tests added: the repo's i18n gate is `check_i18n_locales.py` / `check_i18n_sorted.py`, which both pass; translation-value correctness isn't machine-assertable, hence the request for native review above.</sup>

N/A — string-value only, no layout change.


sdornan:
Thanks — both addressed in 05200097f.

1. **`permission-groups` casing** — fixed to `Grupos de permissões`. I checked the rest of the diff for the same problem and this was the only title-case value I'd introduced.

2. **`rom.upload-tracks`** — good catch, and it exposed a gap in how I found these. My sweep only flagged values *exactly* equal to the `en_US` source, so `"Add tracks"` slipped through twice over: it's English, *and* it's stale against the current `en_US` (`"Upload tracks"`). Now `Enviar faixas`, consistent with `upload-soundtrack-files` → `Enviar arquivos de trilha sonora` and `chip-track-n` → `Faixa {n}`.

After that I re-swept `pt_BR` for the whole missed class — English-looking values that *differ* from `en_US` and so were invisible to an equality check — and `upload-tracks` was the only real instance; the other hits were Portuguese sentences containing `{from}`/`{to}` placeholders.

Worth flagging for maintainers: the same check across the other locales surfaces ~130 candidates (highest: `it_IT`, `pl_PL`, `tr_TR`, `ja_JP`, `ko_KR`). Most look like the same false-positive pattern, but that's a class the i18n CI can't currently catch, since `check_i18n_locales.py` only validates key parity, not values. Out of scope here — happy to open an issue if useful.

sdornan:
Correction to my previous comment: it referenced `05200097f`, which no longer exists.

While I was preparing that fix, the Copilot autofix pushed `e861ac2c5` with a byte-identical `permission-groups` change, so my commit and the branch diverged. I rebased on top of it rather than force-overwriting; git dropped the duplicate hunk, leaving my commit as just the `rom.json` change. Correct state on the branch now:

- `permission-groups` → `Grupos de permissões` — from Copilot's `e861ac2c5`, not mine
- `upload-tracks` → `Enviar faixas` — `d2c781b1f`

Both i18n validators pass at `d2c781b1f`. Everything else in my previous comment stands.

gantoine (approved):

PR #4204, opened 2026-08-15
fix(v2): drop the roms from the gallery when they leave the collection on screen

### LINKED ISSUE #4210: [Other] Extract the duplicated v2 gallery cache-removal block into useRomSync
### Summary

Three v2 sites reconcile the ROM caches by hand with the same three calls. That block belongs in `useRomSync`, which was added in #4153 for exactly this job ("fan a ROM mutation out to every cache that holds it") but currently has no removal counterpart.

### Where

```
gallerySelectionStore.removeIds(ids);
romsStore.remove(roms);
galleryRomsStore.remove(roms);
```

- `frontend/src/v2/components/Gallery/SelectionBar.vue` (bulk unfavourite, added in #4151)
- `frontend/src/v2/components/Dialogs/DeleteRomDialog.vue` (delete)
- `frontend/src/v2/components/Dialogs/ManageCollectionsDialog.vue` (collection removal, added in #4204)

Each was written independently against the same problem: a v2 write that removes ROMs from the current view has to reach the gallery window cache, the v1 store, and the selection, or the cards stay on screen. Missing one of the three is what #4151 and #4204 each fixed, one site at a time. A shared helper is what stops the next site getting it wrong.

### Proposal

Add `removeCachedRoms(roms: SimpleRom[])` to `useRomSync` and route all three call sites through it.

### One design note

The three calls are identical across the sites, but the intent is not:

- `SelectionBar` and `ManageCollectionsDialog` mean "these ROMs left this view" — the ROMs still exist.
- `DeleteRomDialog` means "these ROMs are gone entirely", and separately prunes `recentRoms` and `continuePlayingRoms` around the block.

`removeCachedRoms` should wrap **only** the three cache calls, with `DeleteRomDialog` keeping its own pruning outside it. Folding the pruning into the helper would be wrong for the other two callers: a game removed from a collection should stay in Continue Playing.

### Sequencing

Blocked on #4204, which adds the third call site. Doing it before that merges would convert two sites and let the third land raw straight after, needing a second pass.

### Scope

Pure refactor, no behaviour change. The existing tests in `SelectionBar.test.ts` and `ManageCollectionsDialog.test.ts` already assert the reconcile fires (and doesn't over-fire), so they should pass unchanged; `useRomSync/index.test.ts` gains direct coverage for the new function.

<sub>Filed with Claude Code, from a review thread on #4204.</sub>

sdornan:
> [!NOTE]
> This PR was written primarily by Claude Code (code, tests and this description). I reviewed the approach and directed the scope.

**Description**

Removing ROMs from the collection you're currently looking at left their cards on screen.

`ManageCollectionsDialog.toggle()` updated `collectionsStore` and stopped there. The v2 gallery renders from `galleryRoms`, which never heard about the membership change, so the cards sat there until the gallery happened to refetch on its own (navigating away and back, a filter change, a scan).

This is the same bug #4151 fixed in `SelectionBar`, in the other v2 site that writes collection membership. The fix is the same branch: when the write is a *removal* and the gallery is showing that collection, drop the ROMs from the gallery, the v1 store, and the selection.

Two details worth a reviewer's eye:

- **Removal only.** Adding is unreachable for the collection on screen: every selected ROM is already a member, so the row reads "all" and the tri-state cycle only offers "off". Reconciling on add would be dead code.
- **`targetRoms` snapshot.** The dialog is a singleton, so a fresh `showManageCollectionsDialog` can replace `roms` while the request is in flight. `romIds` was already read before the `await`; this extends that to the ROM objects the response applies to.

The second commit fixes a related staleness on the **failure** path, raised in review. It restored `prevState` on error, which pins the row to a value computed against whichever ROMs the dialog held when the click landed. Since a reopen swaps in both a fresh `roms` and a fresh `optimistic` map, a failure arriving after a reopen wrote the old session's answer into the new one. Clearing the override instead falls back to real membership: the same value in the ordinary case (a failed write changed nothing), the correct one after a reopen, and consistent with what the success path already did.

The third commit removes the duplication the first one added, `Closes #4210`. Three v2 sites now reconciled the caches by hand with the identical three calls:

```js
gallerySelectionStore.removeIds(ids);
romsStore.remove(roms);
galleryRomsStore.remove(roms);
```

in `SelectionBar` (bulk unfavourite, #4151), `DeleteRomDialog` (delete), and this dialog. Each was written independently against the same problem, and missing one of the three is exactly what #4151 and this PR each had to fix, one site at a time. They now all call `useRomSync().removeCachedRoms(roms)`, which is where #4153 put the "fan a ROM mutation out to every cache that holds it" job.

One deliberate boundary: `removeCachedRoms` wraps **only** those three calls. `DeleteRomDialog` keeps its own `recentRoms` / `continuePlayingRoms` pruning outside it, because the intent differs. `SelectionBar` and this dialog mean "these ROMs left this view" and the ROMs still exist; only a real delete should take a game out of Continue Playing.

v1 is untouched.

**Checklist**

**Testing**

5 unit tests in `ManageCollectionsDialog.test.ts`, covering the fix and its guards:

| Case | Expected |
| --- | --- |
| Remove, while viewing that collection | Cards dropped from gallery + v1 store, selection cleared |
| Remove, while viewing a *different* collection | Gallery untouched |
| Add, while viewing that collection | Gallery untouched |
| Remove fails | Gallery untouched, selection intact, error snackbar, row back to real membership |
| Remove fails *after* the dialog is reopened | Row reflects the new selection, not the old one |

Both bug tests were confirmed to fail against the code they fix (the second with `expected 'all' to be 'off'`); the guard tests pass either way by design, pinning down that the branch doesn't over-fire.

2 more in `useRomSync/index.test.ts` for the extracted helper: that it clears all three caches, and that it leaves Home's recent / continue-playing rows alone. The refactor is behaviour-preserving, so the existing `SelectionBar.test.ts` and `ManageCollectionsDialog.test.ts` reconcile tests pass unchanged, which is the real regression proof. Deleting any one of the three calls from the helper was confirmed to fail tests at all three call sites, which is the point of centralising it.

**Verification run:** rebased onto current `master` (needed: `useRomSync` landed in #4153 after this branch was cut). `npm run typecheck` clean, full frontend suite 736 passed across 64 files, eslint and prettier clean on all five changed files, and the trunk pre-commit hook clean. No locale, token, or `/lib` primitive changes, and no backend or response-schema change, so the i18n check, `build:tokens`, Storybook, and `npm run generate` don't apply.

**Not verified in a running instance.** This is a cache-reconcile fix with no visual change, and it's covered by the tests above, but I didn't exercise it against a live library. Worth a click-through by a reviewer who has one up: open a collection, select a few games, and un-toggle that collection in the manage-collections dialog. The refactor touches delete and bulk-unfavourite too, so those are worth a pass as well.

**Related:** #4153 fixed the neighbouring problem, v2 write sites not reaching the gallery cache at all, and added the `useRomSync` composable. The fix here stays independent of its refresh helpers deliberately: `refreshAfterUserStateChange` guards on `currentCollection?.id === favoriteCollection.id`, so it returns early for an arbitrary collection and would leave this bug in place. Only the removal path is shared, via the new `removeCachedRoms`.

### DIFF
--- a/frontend/src/v2/components/Dialogs/DeleteRomDialog.vue
+++ b/frontend/src/v2/components/Dialogs/DeleteRomDialog.vue
@@ -13,9 +13,8 @@ import romApi from "@/services/api/rom";
 import storeConfig from "@/stores/config";
 import storeRoms, { type SimpleRom } from "@/stores/roms";
 import type { Events } from "@/types/emitter";
+import { useRomSync } from "@/v2/composables/useRomSync";
 import { useSnackbar } from "@/v2/composables/useSnackbar";
-import storeGalleryRoms from "@/v2/stores/galleryRoms";
-import storeGallerySelection from "@/v2/stores/gallerySelection";
 
 defineOptions({ inheritAttrs: false });
 
@@ -24,8 +23,7 @@ const router = useRouter();
 const route = useRoute();
 const show = ref(false);
 const romsStore = storeRoms();
-const galleryRomsStore = storeGalleryRoms();
-const gallerySelectionStore = storeGallerySelection();
+const { removeCachedRoms } = useRomSync();
 const roms = ref<SimpleRom[]>([]);
 const romsToDeleteFromFs = ref<number[]>([]);
 const excludeOnDelete = ref(false);
@@ -117,10 +115,9 @@ async function deleteRoms() {
       }
     }
     romsStore.resetSelection();
-    // Drop the deleted ROMs from the gallery selection
-    gallerySelectionStore.removeIds(deletedRoms.map((rom) => rom.id));
-    romsStore.remove(deletedRoms);
-    galleryRomsStore.remove(deletedRoms);
+    removeCachedRoms(deletedRoms);
+    // Deletion is permanent, so unlike the other `removeCachedRoms` callers
+    // this one also takes the ROMs out of Home's rows.
     romsStore.setRecentRoms(
       romsStore.recentRoms.filter(
         (r) => !deletedRoms.some((rom) => rom.id === r.id),
--- a/frontend/src/v2/components/Dialogs/ManageCollectionsDialog.test.ts
+++ b/frontend/src/v2/components/Dialogs/ManageCollectionsDialog.test.ts
@@ -6,6 +6,7 @@ import { ref } from "vue";
 import storeAuth from "@/stores/auth";
 import storeCollections, { type Collection } from "@/stores/collections";
 import storeRoms, { type SimpleRom } from "@/stores/roms";
+import type { User } from "@/stores/users";
 import type { Events } from "@/types/emitter";
 import storeGalleryRoms from "@/v2/stores/galleryRoms";
 import storeGallerySelection from "@/v2/stores/gallerySelection";
@@ -67,10 +68,24 @@ function collection(romIds: number[]): Collection {
 async function toggleRow(roms: SimpleRom[]) {
   const emitter = mitt<Events>();
   const wrapper = mountDialog(emitter);
-  emitter.emit("showManageCollectionsDialog", roms);
-  await flushPromises();
+  await open(wrapper, emitter, roms);
   await wrapper.get("[data-test-row]").trigger("click");
   await flushPromises();
+  return { wrapper, emitter };
+}
+
+async function open(
+  wrapper: VueWrapper,
+  emitter: Emitter<Events>,
+  roms: SimpleRom[],
+) {
+  emitter.emit("showManageCollectionsDialog", roms);
+  await flushPromises();
+}
+
+/** The tri-state the row is currently painting. */
+function rowState(wrapper: VueWrapper) {
+  return wrapper.get("[data-test-row]").attributes("data-state");
 }
 
 function mountDialog(emitter: Emitter<Events>): VueWrapper {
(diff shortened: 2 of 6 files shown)


sdornan:
Thanks — both findings were valid and are fixed in `2d04f6fb`.

**Fixed: the error path pinned the row** (@copilot-pull-request-reviewer). The interesting one, and it's the same class of bug as the `targetRoms` snapshot in this PR — I fixed the success path's staleness and left the failure path with the identical hazard.

`prevState` is computed against whichever ROMs the dialog held when the click landed. Because the dialog is a singleton and `openHandler` swaps in a fresh `roms` and a fresh `optimistic` map, a failure arriving after a reopen wrote the old session's answer into the new session's map, pinning that row. Clearing the override instead falls back to real membership, which is the same value in the ordinary case (a failed write changed nothing) and the correct one after a reopen. It also matches what the success path already does.

Covered by a new test that drives the actual ordering — click, reopen over a game the collection doesn't hold, *then* reject. It asserts `off` and fails with `expected 'all' to be 'off'` against the previous code.

**Fixed: `as never` in the test setup.** Fair. Now `as User`, which typechecks and says what it means.

Worth noting what this doesn't fix, since it's adjacent and pre-existing: `pendingCollections` is also reset on open, so a reopen can start a second toggle for the same collection while the first is in flight, and the first request's `finally` then clears the second's pending flag. Properly fixing that wants a session token captured at open that every post-await write checks, which is a bigger change than this PR should carry. Happy to open it as an issue if a maintainer agrees.

**Re-verified:** typecheck, the full frontend suite (707 passed), eslint, prettier, and the trunk pre-commit hook, all clean.

sdornan:
Filed the duplication follow-up as #4210, blocked on this PR since it adds the third call site. Keeping it out of here so this stays a scoped bug fix.

gantoine (approved):

gantoine:
@greptileai

PR #4228, opened 2026-08-18
fix(ss): don't use the Switch icon URL as the manual URL

TowyTowy:
**Description**

Both Switch TitleDB fallbacks in `ss_handler.get_rom()` set `url_manual` to the icon URL, the same value as `url_cover`:

```python
url_cover=index_entry.get("iconUrl", ""),
url_manual=index_entry.get("iconUrl", ""),
```

The TitleDB index has no manual, so this is just the cover line copied with the key renamed. `igdb_handler.py` and `moby_handler.py` have the identical fallback blocks and set only `url_cover`, so ScreenScraper is the only handler doing this.

Effect: any Switch ROM resolved through the title ID or product ID fallback gets an icon URL stored as its manual, and `_store_manual()` then fetches that icon on scan while trying to download a manual. The content type check in `resources_handler` rejects a PNG, so no bogus PDF is written, but the request is made and thrown away every time, and the wrong value is still persisted and served over the API.

`url_manual` is `NotRequired` on `BaseRom`, so dropping the line is enough; nothing needs a placeholder.

**Checklist**

Added `test_switch_titledb_fallback_does_not_set_icon_as_manual`, which drives `get_rom()` through the title ID fallback with a mocked TitleDB entry and asserts the icon lands in `url_cover` but not in `url_manual`. It fails before the change (`assert not 'https://example.net/icon.png'`) and passes after. Full `tests/handler/metadata/` suite is green at 417 passed.

**AI disclosure**

Written with Claude (Claude Code), as required by `CONTRIBUTING.md`. I verified the failing-before and passing-after states locally and checked the sibling handlers to confirm which behaviour is the intended one.

### DIFF
--- a/backend/handler/metadata/ss_handler.py
+++ b/backend/handler/metadata/ss_handler.py
@@ -940,7 +940,6 @@ class SSHandler(MetadataHandler):
                     name=index_entry["name"],
                     summary=index_entry.get("description", ""),
                     url_cover=index_entry.get("iconUrl", ""),
-                    url_manual=index_entry.get("iconUrl", ""),
                     url_screenshots=index_entry.get("screenshots", None) or [],
                 )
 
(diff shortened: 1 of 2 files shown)


gantoine (approved):
nice catch

gantoine:
@greptileai

PR #4240, opened 2026-08-19
fix(backend): make filters match against all grouped versions instead of only the main version of each game entry

darkpaul91:
**Description**
If you have a game with mutiple versions, those may be shown under one single game entry, but the library filters try to match just against the one marked or detected as the primary version of the game.

The fix just changes the order of operations already in the code, allowing to apply the filters against all the versions before the grouping has taken effect, instead of doing it afterwards.

**Example scenario**
Let's say our library has these three rom files for the same Nintendo Entertainment System game:
- library/roms/nes/Double Dragon (Europe) (En)/Double Dragon (Europe) [!].nes  -> (Marked as Primary or Main version)
- library/roms/nes/Double Dragon (Japan) (Ja)/Sou Setsu Ryuu (Japan) [!].nes
- library/roms/nes/Double Dragon (USA) (En)/Double Dragon (USA) [!].nes

Before the fix, if you filter the list of games by Regions="Japan", Regions="USA" or Languages="Japanese" the game won't show up at all even if those versions do in fact exist in the library.

**Checklist**

#### Screenshots

example 1: no filters enabled
no-filters

example 2: filters -> regions="Japan" before fix (all entries disappear except the one whose main version matches the filter)
filters-region-japan-without-fix

example 3: filters -> regions="Japan" after fix (the entries appear again, showing the version that matches the filter)
(double dragon III doesn't have any other version in this example and is therefore correctly still missing)
filters-region-japan-with-fix

### DIFF
--- a/backend/handler/database/roms_handler.py
+++ b/backend/handler/database/roms_handler.py
@@ -1234,6 +1234,36 @@ class DBRomsHandler(DBBaseHandler):
         if updated_after:
             query = query.filter(Rom.updated_at > updated_after)
 
+        # Apply metadata and rom-level filters efficiently
+        # Moved before applying group_by_meta_id to avoid missing titles when
+        # filters don't match the primary ROM version in a group but would match a different version instead.
+        filters_to_apply = [
+            (genres, genres_logic, self._filter_by_genres),
+            (franchises, franchises_logic, self._filter_by_franchises),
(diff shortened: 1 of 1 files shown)


gantoine (approved):

gantoine:
@greptileai

PR #4247, opened 2026-08-20
fix(metadata): match dot-separated filenames like underscored ones

sdornan:
**Description**

A dump that spells its title with periods reached the metadata providers with those periods intact, while the underscored equivalent was already normalized into words. `_normalize_search_term` replaced underscores unconditionally, but only dropped periods as part of the blanket punctuation removal that ScreenScraper (`ss_handler.py`), HLTB and Flashpoint deliberately opt out of (`remove_punctuation=False`, because those APIs match on punctuation).

So a library holding `jazz_jackrabbit_2_secret_files_windows_gog_(85159).zip` matched everywhere, while `Final.Fantasy.VII.iso` was sent to ScreenScraper as `final.fantasy.vii`.

Two changes:

1. **`_split_dotted_name` in `handler/metadata/base_handler.py`** reads periods as word breaks alongside underscores, so it applies whatever the punctuation flag. It only fires on a name that uses nothing else as a separator and holds at least two multi-character segments, which leaves real punctuation alone for the providers that want it.

2. **`EXTENSION_REGEX` in `models/base.py`** fed that path an already-truncated stem. Its compound-extension group (`([a-z]+\.)*`) walked back over every lowercase segment, so `final.fantasy.vii.iso` parsed as extension `fantasy.vii.iso` and stem `final`, while the title-cased spelling of the same name kept its words. The compound branch is now the tarball family only (`(?:tar\.)?`), matched case-insensitively.

| file name | extension | stem | search term |
| --- | --- | --- | --- |
| `Final.Fantasy.VII.iso` | `iso` | `Final.Fantasy.VII` | `final fantasy vii` |
| `final.fantasy.vii.iso` | `iso` (was `fantasy.vii.iso`) | `final.fantasy.vii` (was `final`) | `final fantasy vii` (was `final`) |
| `castlevania.symphony.of.the.night.chd` | `chd` | full stem | `castlevania symphony of the night` |
| `game.tar.gz` / `GAME.TAR.GZ` | `tar.gz` / `TAR.GZ` | `game` / `GAME` | `game` |
| `F.E.A.R.` / `Dr. Mario` / `Sonic 3.5` | unchanged | unchanged | periods kept |

Knock-on changes from limiting the compound branch, all in the same direction (only the last dot-run is an extension): `file.with.dots.txt` now parses extension `txt` and stem `file.with.dots` (was `with.dots.txt` / `file`); likewise `game.nds.hash.txt` → `txt` and `game.bin.ecm` → `ecm`. Two assertions in `tests/handler/filesystem/test_base_handler.py` encoded the old greedy result and are updated. Extension-based exclusions are unaffected: `exclude_single_files` matches configured suffixes with `endswith`, not this regex.

Not linked to an issue; found while looking at how filenames with periods are matched.

**Checklist**

**AI assistance disclosure**

Written with Claude Code (Opus 5), driven interactively. The AI did the codebase investigation, wrote the implementation and the tests, and ran the verification below; every design decision was reviewed and directed by me, including scoping the change down from a larger extension-parser rewrite to these two targeted fixes.

**Verification**

Full backend suite green (`2982 passed, 2 skipped`) and `trunk fmt && trunk check` clean on the diff. New coverage: dot-separated separators plus the initialism/abbreviation/version guard in `tests/handler/metadata/test_base_handler.py`, and the lowercase/title-cased dotted names plus `.TAR.GZ` casing in `tests/models/test_base.py`.

### DIFF
--- a/backend/handler/metadata/base_handler.py
+++ b/backend/handler/metadata/base_handler.py
@@ -50,6 +50,26 @@ NON_WORD_SPACE_PATTERN = re.compile(r"[^\w\s]")
 MULTIPLE_SPACE_PATTERN = re.compile(r"\s+")
 
 
+def _split_dotted_name(name: str) -> str:
+    """Turn the periods of a dot-separated name into spaces.
+
+    A scene-style dump ("Final.Fantasy.VII") spells its title with periods where
+    another dump uses underscores, and ScreenScraper, HLTB and Flashpoint are
+    searched with punctuation left in, so those periods reach the provider as
+    part of the words. Names that separate words some other way are left alone,
+    keeping the periods of an initialism ("F.E.A.R."), an abbreviation
+    ("Dr. Mario") and a version ("Sonic 3.5") intact for the same providers.
+    """
+    if "." not in name or len(name.split()) != 1:
+        return name
+
+    segments = [segment for segment in name.split(".") if segment]
+    if sum(len(segment) > 1 for segment in segments) < 2:
+        return name
+
+    return " ".join(segments)
+
+
 class BaseRom(TypedDict):
     name: NotRequired[str]
     name_sort_key: NotRequired[str | None]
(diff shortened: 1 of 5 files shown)


sdornan:
Going to close this one because I'm not sure it's worth the risk of regression to fix something that probably affects a very small number of people. The best solution is for them to standardize their file names. Feel free to reopen if necessary.

PR #4264, opened 2026-08-22
Fix N+1 and entity sort in GET /api/roms

tmgast:
## Description

`GET /api/roms` builds every page by doing work it then throws away. This removes eight instances of it.

Nothing here changes what the endpoint returns. Response bodies are byte identical before and after, across the gallery, sync and sidecar shapes, so there is no schema change and no frontend regeneration. On a 23,873 rom library the endpoint is 1.9x to 10x faster, and a page of 100 ROMs on a platform averaging 104 files per ROM falls from 10,456 statements to 45.

### Findings

The eight are four loader options that pull relationships the list path never reads, one N+1 on the page's files, one sort over whole entities, one derivation replacing three correlated subqueries per row, and one missing index. They do not all show up in the same request, so the table separates them by request shape: what a client paging with `with_files` sends, what the same request costs at depth with the rom id index on and off, and what the v2 gallery sends on load and on scroll.

Measured against a 23,873 rom library (50,250 `rom_files`, 24,678 `track_meta`, 19,662 sibling rows) on MariaDB 11.4.9 with the stock 256 MiB buffer pool. Median of three, warmed. `q` is the server side statement count, read from MariaDB's `Questions` counter rather than from SQL echo, so the instrument does not perturb the timing.

| Request                              | What it shows                        |            Before |           After | RTT improvement |
| ------------------------------------ | ------------------------------------ | ----------------: | --------------: | --------------: |
| Sync page, 104 files/ROM             | one query per file for track metadata | 4370ms / 10456q  |  477ms /  45q  | 9.2x |
| Sync page 2, 104 files/ROM           | same, past the first page            | 4155ms / 10432q  |  537ms /  45q  | 7.7x |
| Sync page, 1.8 files/ROM             | the same N+1 at ordinary file counts | 374ms /   179q   |   79ms /  25q  | 4.7x |
| Page at offset 0, id index on        | page served by primary key           | 628ms /   125q   |  192ms /  25q  | 3.3x |
| Page at offset 0, id index off       | page sorted as whole rows            | 1709ms /   122q  |  218ms /  23q  | 7.8x |
| Page at offset 3000, id index on     | "                                    | 532ms /   125q   |  187ms /  25q  | 2.9x |
| Page at offset 3000, id index off    | sort cost grows with depth           | 1784ms /   122q  |  183ms /  23q  | 9.8x |
| Page at offset 6000, id index on     | "                                    | 572ms /   125q   |  208ms /  25q  | 2.8x |
| Page at offset 6000, id index off    | deepest page, worst sort             | 2046ms /   122q  |  204ms /  23q  | 10.0x |
| Gallery window, first paint          | what the web UI sends on load        | 348ms /    20q   |  129ms /  20q  | 2.7x |
| Gallery window, after scrolling      | what the web UI sends on scroll      | 1339ms /    20q  |  156ms /  20q  | 8.6x |
| Gallery window, sibling heavy        | discarded sibling and rom_user loads | 110ms /    20q   |   58ms /  20q  | 1.9x |

Two causes dominate the table. The first is the files query, which lazy loaded `track_meta` once per file: a single page of 100 ROMs issued 10,456 statements on a platform averaging 104 files per ROM, and now issues 45. Every row that requests files pays some of that. The gallery rows request none and sit unchanged at 20 statements either side, so their gains come from the discarded loader options and the sort.

The second is the page sort. The `id index on` and `id index off` rows at each offset are the same request one parameter apart, and opting out cost 2.7x at offset 0 and 3.6x at offset 6000 while the statement count fell, 125 to 122. Query volume was never the problem, sorting whole rows was. This is not only a sync path: the v2 gallery sends `with_rom_id_index=false` on every window fetch, so a scrolled window was paying that sort at 1339ms. It now serves in 156ms.

### Changes

- **The files query skipped track metadata.** `get_files_for_roms` was a bare `select(RomFile)`, while the seven other call sites in that file all eager load `RomFile.track_meta`. The relationship is a plain `lazy="select"`, unlike every relationship on `Rom`, so building the response cost one query per file. That scales with files per ROM, not rows per page.

- **Paging sorted whole entities.** With the rom id index off, the page came from `OFFSET n LIMIT m` over the ordered entity, dragging every JSON metadata blob through the sort. It now orders bare ids and fetches the rows by primary key, which is what the index-on branch already did with its sliced id list.

- **The notes indicator loaded every note in full.** `RomNote.rom` and `RomNote.user` are both `lazy="joined"`, so eager loading `Rom.notes` pulled a full `roms` row and a `users` row per note, plus every note body (`Text`), to resolve one boolean. One batched query resolves it instead. The test database contained no notes, so no metrics could be aggregated for this one.

- **Nothing reads `RomUser.user`.** `RomUserSchema` carries no user fields, and the adjacent `noload(RomUser.rom)` was already there.

- **`get_siblings_for_roms` joined two tables it never reads.** `load_only` narrows columns but not relationships, and both `Rom.platform` and `Rom.metadatum` are `lazy="joined"`, so each sibling row pulled a platforms row and six JSON columns from `roms_metadata` to fill a four field schema.

- **The list path hydrated siblings twice.** It resolves them through `get_siblings_for_roms` and passes them to the schema explicitly, so `from_orm_with_request` never reads the relationship, yet `selectinload(Rom.sibling_roms)` ran anyway and hydrated full `Rom` entities for the same siblings. Callers that resolve siblings themselves now opt out. Every other caller still gets the whole relationship. Narrowing it inside a shared helper would hand the next caller a silent per object lazy load instead of an error.

- **The file stat flags ran three correlated subqueries per row.** That is 300 on a 100 row page. When `with_files` is set the page's files already answer all three, so they are derived instead. `has_soundtrack` stays a `column_property` because `filter_roms` filters on it.

- **`rom_files` had no index covering `category`.** `has_soundtrack` renders as `EXISTS`, which enumerates a ROM's files to conclude it has none. The composite answers that from the index, and makes the `has_soundtrack=true` filter indexable, which it was not. Tidiness for most libraries, headroom as soundtrack collections grow.

### Tests

- A differential test asserts the derived file stat flags and the subqueries answer identically for a flat file, a folder with two top level files, a nested single file, a soundtrack folder, and a ROM with no file rows at all, so neither predicate can be edited without the other. The derivation does not reuse `RomFile.is_top_level`, which answers a different question and disagrees on nested files.
- A query count test asserts a ROM's files load in a fixed number of statements rather than one per file.
- Checked against a real library as well. The derived flags match the original SQL expressions over 3,029 ROMs on five platforms covering every file layout. Six captured response shapes are byte identical to `master`, apart from the per request `rom_user` placeholder timestamps, which differ between two runs of `master` too.
- The migration applies and reverses on MariaDB against a populated table and on PostgreSQL from an empty schema. Adding the index took about 1.2s on 50,250 rows.

> [!NOTE]
> Spotted while testing and left alone, since it predates this change. `RomFile.last_modified` is nullable on the model but `RomFileSchema.last_modified` is a required datetime, so a file row with a NULL there raises a validation error and 500s the endpoint when `with_files=true`. The test database contained no such rows, so it was not observed in practice.

---

> AI Disclosure

**Checklist**


sdornan:
Whoa!

gantoine (approved):

PR #4271, opened 2026-08-23
fix: Preserve nested arcade ROM filenames

### LINKED ISSUE #1893: [Bug] ROMs can't be opened in FBNeo/mame from subfolders
**RomM version**
3.9.0

**Describe the bug**
ROMs can't be opened in FBNeo from subfolders
There's another detail: non-merged or merged ROMs only load correctly if the neogeo.zip BIOS is compressed inside another ZIP file.
For example: bios/neogeomvs/neogeo.zip/neogeo.zip.
The outer ZIP file can have any name, as long as it contains the actual neogeo.zip inside it.
Example: bios/neogeomvs/any_name.zip/neogeo.zip.
**To Reproduce**
Placing the ZIP file in a subfolder (e.g., rom/neogeomvs/Art Of Fighting/aof.zip)
Loading the game normally from the page
It says the ROM set is unknown.

Image

Note: If I place it as rom/neogeomvs/aof.zip, it works correctly.

Image

I’ve tried both full non-merged and non-merged sets. In both cases, aof.zip doesn’t load unless it's directly in the neogeomvs folder.

**Expected behavior**
"I should be able to play without unnecessary complications."

System: Windows 11
Browser: Edge

mvanhorn:
**Description**

`getDownloadPath` derived the content path segment from `rom.fs_name` whenever
no explicit file selection was supplied. For a nested single-file ROM — an
arcade entry whose directory holds exactly one file — that produced the
directory name rather than the real filename, so the download arrived named
after the folder (`.../content/Art Of Fighting` instead of `.../content/aof.zip`).

This adds a `nestedFile` branch: when no `file_ids` were selected, the ROM is
flagged `has_nested_single_file`, and it has exactly one entry in `files`, that
entry's `file_name` becomes the content segment. Every other branch is
untouched — flat single-file ROMs, explicitly selected discs, multi-file ZIP
downloads, and unknown file IDs all resolve exactly as before.

The decision stays in the shared helper rather than adding player-specific
configuration, so the existing EmulatorJS, console, Ruffle, and download call
sites all receive a correctly named payload with no new wiring.

**Checklist**

Unit tests in `frontend/src/utils/index.test.ts` cover the new nested
single-file case and pin the unchanged flat single-file and multi-file
behaviors. The other boxes are left unchecked because those steps were not
performed.

Not applicable — this changes a download URL path, with no UI surface.

Fixes #1893

AI was used for assistance.

### DIFF
--- a/frontend/src/utils/index.test.ts
+++ b/frontend/src/utils/index.test.ts
@@ -12,11 +12,41 @@ function makeRom(overrides: Partial<SimpleRom>): SimpleRom {
 }
 
 describe("getDownloadPath", () => {
-  it("uses fs_name when no file is selected", () => {
-    const rom = makeRom({ id: 14, fs_name: "Maniac Mansion (1989).adf" });
+  it("uses fs_name for a flat single-file rom", () => {
+    const rom = makeRom({
+      id: 14,
+      fs_name: "Maniac Mansion (1989).adf",
+      has_nested_single_file: false,
+      files: [
+        { id: 19, file_name: "Maniac Mansion (1989).adf" },
+      ] as SimpleRom["files"],
+    });
     expect(getDownloadPath({ rom })).toBe(
       "/api/roms/14/content/Maniac Mansion (1989).adf",
     );
+  });
+
+  it("uses the file name for a nested single-file rom", () => {
+    const rom = makeRom({
+      id: 21,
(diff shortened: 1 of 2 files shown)


gantoine (approved):

gantoine:
@greptileai

Issue #4186, opened 2026-08-10
[Bug] Scan already in progress, ignoring request


MadMullins:
**RomM version**
5.1.0

**Describe the bug**
When trying to perform any kind of scan, an error message appears that says "Scan failed: A scan is already in progress"
In the logs: "Scan already in progress, ignoring request"
Waiting any amount of time shows no scan progress in the logs.

**To Reproduce**
Perform any form of scan with any variation of selected options.

**Expected behavior**
A new scan is performed.

**Screenshots**

**Server (please complete the following information):**

- OS: Linux Mint 22.3. 
- Running via Docker Compose
- Romm version 5.1.0

**Client (please complete the following information):**

- Device: PC
- OS: Linux Mint
- Browser Zen Browser 1.21.10b

florianduros:
I'm reproducing the same issue

greyluked:
I have the same issue. My info:

**Server:**

OS: TrueNAS Community Edition 25.10.5 - Goldeye
Running as TrueNAS app version: 1.2.32
Romm version 5.1.0

**Client:**

Device: PC
OS: Bazzite
Browser: Vivaldi 8.1.4087.62

EDIT:

I disabled filesystem watcher and enabled scheduled scans instead, and now when trying to scan manually the error I get is "Scan failed: You are not authorized to run scans". This is done from the admin account, which should have all the permissions.

When disabling both filesystem watcher and scheduled scans, manual scans work.

Mernst1:
same problem for me to.

Update:
Problem solved.
It was a corruption in the Redis Database for romm
Need to delete the corrupted jobs in the database.

This was in the logs from romm, which hinted me to redis.

> `WARNING:  [RomM][init][2026-08-21 22:28:09] rq_scheduler exited unexpectedly, restarting it
> INFO:     [RomM][init][2026-08-21 22:28:09] Starting RQ scheduler
> 22:28:10 Registering birth
> 22:28:14 Registering death
> Traceback (most recent call last):
>   File "/src/.venv/lib/python3.13/site-packages/rq/job.py", line 575, in _deserialize_data
>     self._func_name, self._instance, self._args, self._kwargs = self.serializer.loads(self.data)
>                                                                 ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
>   File "/usr/local/lib/python3.13/enum.py", line 727, in __call__
>     return cls.__new__(cls, value)
>            ~~~~~~~~~~~^^^^^^^^^^^^
[...]

PR #4272, opened 2026-08-23
fix(scan): stop refusing scans a stalled scheduler left pending

sdornan:
**Description**

Scans have been refused with "Scan failed: A scan is already in progress" on 5.1.0, with `Scan already in progress, ignoring request` in the logs and no scan ever appearing. Waiting does not help, because nothing is running to wait for.

The cause is a stalled rq-scheduler. It reads a job's function name before it takes the job out of its registry, so one job left behind by an older version, holding an argument that no longer deserializes, crashes it on every poll and stays there for the next one to trip on ([the `'unidentified' is not a valid ScanType` traceback in #4186](https://github.com/rommapp/romm/issues/4186), from a delayed job created before `ScanType` was refactored in 9fa15d20f). Nothing scheduled runs again, and every watcher rescan piles up in the registry behind it. `_get_queued_scan_jobs` counted those as scans waiting to start, so the concurrency guard added in 99e214a8b refused every manual scan for as long as the scheduler stayed broken. That matches all three reports in the issue: watcher off makes scans work again, and clearing Redis fixes it.

What this changes:

- **Clear jobs the scheduler cannot read at startup**, before it polls them again, so an affected instance recovers on upgrade instead of needing Redis wiped by hand.
- **Split scan discovery**, so the guard consults only scans sitting on a worker queue. Delayed scans block nothing now: the scheduler is the only thing that releases them, so a scheduler that is down must not be able to refuse scans.
- **Drop delayed scans more than an hour past due at startup.** A recovered scheduler would otherwise release its whole backlog at once and run the same library scan over and over. The standing cron entry is left alone, since it reschedules itself.
- **Stopping a scan now drops delayed scans out of the scheduler's registry**, not just cancels the job. Cancelling alone left the id in the registry, and the scheduler queued it anyway once the delay was up, running the scan that was just stopped.
- **Tolerate jobs that are already gone.** A worker killed mid-scan points at a job that may no longer exist, which raised `NoSuchJobError` out of the socket handler with no reply to the client, and out of `GET /tasks/status` as a 500. Reading a job's status has the same problem once its hash expires, so it goes through a wrapper alongside `get_job_func_name`.
- **Say what is in the way** when refusing: "Quick Scan is already running", "… is already queued", or "… is still stopping, try again in a moment".

No API or schema change, so no frontend type regeneration. v2 already reconciles with a running scan via `/tasks/status`; extending that to a refusal is a separate UI change.

Related defect found while investigating, not fixed here: the watcher's own dedupe in `get_pending_scan_jobs` never matches, because it filters on `job.args[0]` while the watcher enqueues with keyword arguments only. That is why the backlog grows one job per filesystem change.

**Checklist**

**Testing**

Full backend suite green (3015 passed, 2 skipped), `trunk fmt && trunk check` clean.

New unit tests cover the guard (a scheduled watcher scan, a cancelled queued scan, a job whose status is gone, and a worker holding a job that no longer exists all stop blocking), the refusal messages, dropping delayed scans out of the scheduler on stop, the unreadable-job purge, and the stale-backlog cleanup.

Verified against a real Redis and a real rq-scheduler registry as well, since mocks cannot exercise the rq-scheduler interaction: a job with unreadable data does crash `enqueue_jobs()` with `DeserializationError` and remains in the registry; the purge clears it and the next poll succeeds; a three-scan backlog is dropped while a scan still waiting out its delay survives; and a delayed scan contributes nothing to the queued set the guard checks.

**AI assistance disclosure**

This change was written with AI assistance (Claude Opus 5 via Claude Code). The issue diagnosis, the fix, the tests, and this description were AI-generated, then reviewed and verified by me: I confirmed the rq-scheduler and RQ behaviour in their sources, ran the test suite and linters, and ran the real-Redis verification described above.

### DIFF
--- a/backend/endpoints/sockets/scan.py
+++ b/backend/endpoints/sockets/scan.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 import asyncio
 from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
 from itertools import batched, chain
 from typing import Any, Final
 
 import pydash
 import socketio  # type: ignore
 from rq import Worker, get_current_job
+from rq.exceptions import InvalidJobOperation, NoSuchJobError
 from rq.job import Job, JobStatus
 from sqlalchemy.exc import IntegrityError
 
@@ -52,6 +54,7 @@ from handler.metadata.ss_handler import log_quota as log_ss_quota
 from handler.metadata.ss_handler import log_scan_summary as log_ss_scan_summary
 from handler.redis_handler import (
     get_job_func_name,
+    get_job_status,
     high_prio_queue,
     low_prio_queue,
     redis_client,
@@ -79,6 +82,10 @@ from utils.gamelist_exporter import GamelistExporter
 from utils.pegasus_exporter import PegasusExporter
 
 STOP_SCAN_FLAG: Final = "scan:stop"
+
+# A delayed watcher scan this far past due was left behind by a scheduler that
+# stopped releasing them, and the change it reacted to has long since settled.
+STALE_SCHEDULED_SCAN_AGE: Final = timedelta(hours=1)
 
 
 def _scan_platforms_func_name() -> str:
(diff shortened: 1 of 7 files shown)


sdornan:
Superseded by #4274, which replaces `rq-scheduler` with RQ's own cron and delayed-job scheduling.

Everything this PR fixed is carried there: the `NoSuchJobError` guards on the socket handler and `/tasks/status`, the safe job-status read, scheduled scans no longer counting as scans in flight, the refusal message naming what is in the way, and the stale-backlog sweep. The two workarounds it needed are deleted instead of kept — RQ's scheduler cannot stall on a job it cannot read, and `Job.cancel()` alone clears the scheduled registry.

`Fixes #4186` has moved to #4274.

PR #4273, opened 2026-08-23
chore(deps): upgrade rq to 2.11

sdornan:
**Description**

`uv.lock` sat on rq 2.9.0 while `pyproject.toml` allowed anything under 3.0, so the floor now names the version the code is actually tested against.

2.11 is the release the scheduler work ahead needs:

- `RQScheduler` acquires and refreshes its lock before enqueueing, which matters as soon as delayed jobs are released by a worker rather than a separate process.
- Each `CronJob` has a name and keeps the ids of the jobs it created (`cron_job.get_job_ids()`) — the introspection a move to `rq cron` would otherwise cost, since cron state stops living in Redis.
- Calling `create_cron()` more than once no longer duplicates jobs.

2.10 also added webhook notifications and a stable scheduler identity, and 2.9.1 covers `redis-py` >= 8.

The constraint stays under 3.0 deliberately: that release moves `get_current_job()` to `contextvars`, reworks dependency handling behind a `ReadyJobRegistry`, and changes the `Worker.handle_job_success()` signature, which `RomMWorker` subclasses around.

**Note on the quarantine window**

2.11.0 was published on 2026-08-17, so it is still inside the rolling 7-day `exclude-newer` window and needs a per-package exclusion to resolve today. The window reaches it on **2026-08-24**, after which the entry can be deleted and the constraint resolves on its own — happy to drop the override and let it land that way instead, if you would rather not bypass the window at all.

**Checklist**

<sup>No tests added: this is a dependency bump with no behaviour change of its own.</sup>

**Testing**

Full backend suite green (3003 passed, 2 skipped), `trunk fmt && trunk check` clean, and the lockfile diff touches nothing but rq.

Also verified against a real Redis, since the suite's coverage of the scheduler is mostly mocked. Every rq internal the `rq-scheduler` fork reaches into still exists in 2.11 (`rq.utils.backend_class`, `rq.utils.import_attribute`, `rq.logutils.ColorizingStreamHandler`, `Queue.from_queue_key`, `Queue.enqueue_job`, `job.enqueue_at_front`), and each scheduler API RomM uses still behaves: `enqueue_in` and `cron` produce `SCHEDULED` jobs with the expected `func_name`, `get_jobs(with_times=True)` returns due times, a past-due job is queued by `enqueue_jobs()`, and `cancel()` clears the registry entry.

One thing that surfaced while testing, unrelated to this bump: `rq-scheduler` calls `crontab.CronTab.next()` without `default_utc`, which now emits a `FutureWarning` about the coming `utcnow()` change in crontab 0.22+.

**AI assistance disclosure**

This change was written with AI assistance (Claude Opus 5 via Claude Code). The changelog review, the compatibility checks, and this description were AI-generated, then verified by me against the rq sources and a local run of the suite plus the real-Redis checks described above.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
   "python-magic ~= 0.4",
   "python-socketio ~= 5.16",
   "redis ~= 6.2",
-  "rq ~= 2.7",
+  "rq ~= 2.11",
   # TODO: Move back to upstream `rq-scheduler`, when support for username and SSL settings is added.
   #       Related PR: https://github.com/rq/rq-scheduler/pull/325
   "rq-scheduler @ git+https://github.com/adamantike/rq-scheduler.git@feat/script-options-username-ssl",
@@ -146,7 +146,10 @@ package = false
 exclude-newer = "7 days"
 # vcrpy >= 8.2.0 is required for aiohttp 3.14 compatibility (the removal of
 # `AsyncStreamReaderMixin`); allow it past the rolling 7-day window.
-exclude-newer-package = { vcrpy = "2026-06-17" }
+# rq 2.11.0 carries the scheduler locking and cron job history the scheduler
+# migration builds on. The rolling window reaches it on 2026-08-24, after which
+# this entry can go.
+exclude-newer-package = { vcrpy = "2026-06-17", rq = "2026-08-17" }
 
 [tool.ty.environment]
 root = ["./backend"]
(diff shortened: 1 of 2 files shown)


sdornan:
Folded into #4274, which is now the only consumer of the bump.

Correcting the claim in the description while I am here: the scheduler migration does not *need* 2.11. Every API it uses exists in 2.9.0, which was already locked. 2.11 is what it benefits from — the scheduler refreshing its lock before enqueueing, which matters once a worker rather than a separate process releases delayed jobs, and `CronJob` names and job history, which is the introspection that moving cron out of Redis would otherwise cost.

The `exclude-newer-package` entry for rq carries over to #4274 and is deletable from 2026-08-24, when the rolling window reaches 2.11.0.

PR #4313, opened 2026-08-31
Show a clear message instead of the generic error for threaded cores over non-secure origins

TRusselo:
**Description**

Threaded EmulatorJS cores (`dosbox_pure`, `ppsspp`, `azahar`, and any custom threaded core) need `SharedArrayBuffer`, which browsers only expose on a secure context (HTTPS, or `localhost`) — regardless of what COOP/COEP headers the server sends. When a user hits Play for one of these cores on a non-secure origin (a bare LAN IP over plain HTTP is the common case for self-hosted instances), EmulatorJS's own `downloadGameCore()` shows a bare, unhelpful on-canvas message: "Error for site owner / Check console". The actual reason only appears as a `console.warn` most users will never open.

This adds a pre-flight check in both the v1 and v2 players (right where each already knows the selected core and is about to hand off to EmulatorJS's loader) using the same condition EmulatorJS itself checks internally, and shows a clear, actionable snackbar instead: "Must be run over HTTPS. Contact the site owner." — before ever attempting to load the emulator.

- v1 (`views/Player/EmulatorJS/Base.vue`): uses the existing `snackbarShow` emitter event, matching the rest of v1's error-reporting convention.
- v2 (`v2/views/Player/EmulatorJS.vue`): uses the `useSnackbar()` composable and a new i18n key (`play.https-required`), added only to `en_US/play.json` — vue-i18n's fallback locale is already `en_US`, so every other language shows this (universally-recognizable) English string until someone translates it.

Found and fixed while integrating a custom core into a self-hosted instance and hitting this exact confusing error myself.

**Checklist**

No existing test files cover `Base.vue` or `v2/views/Player/EmulatorJS.vue` (large, heavily browser/canvas-integrated components with no current unit test harness), so no new tests were added, consistent with the existing coverage pattern in this area. The full existing suite (901 tests) passes, `vue-tsc --noEmit` is clean, and the fix was verified live against a running instance with a real threaded core over both HTTP (toast now shows) and HTTPS (game boots normally, unaffected).

N/A (text-only toast change)


zurdi15 (dismissed):
Thank you for the PR. Two things that need to be fixed:

- v1 frontend is in freeze, every new fix/change must be done to v2 only.
- New locales must be added in every supported language.

You can see both in the workflow runs summary

gantoine (approved):

Issue #3758, opened 2026-07-16
[Bug] 5.0.0: virtual collections return HTTP 500 on MariaDB 11.6+ (Illegal mix of collations in virtual_collections view)


kthistlewood:
**RomM version**
5.0.0 (upgraded from 4.9.2; alembic migrated cleanly 0082 -> 0095)

**Describe the bug**
After upgrading to 5.0.0, all virtual collections (genre/franchise/company/collection/mode) disappear from the UI. `GET /api/collections/virtual?type=<any>` returns HTTP 500. Manual and smart collections still work. Backend log shows:

```
sqlalchemy.exc.OperationalError: (mariadb.OperationalError) Illegal mix of collations
(utf8mb4_general_ci,COERCIBLE) and (utf8mb4_uca1400_ai_ci,COERCIBLE) for operation '='
```

Root cause (verified on the affected instance): 5.0.0 defines `virtual_collections` as a SQL view (JSON_TABLE over `roms_metadata`). The view's string expressions resolve to `utf8mb4_general_ci`, while MariaDB 11.6+ resolves utf8mb4 expressions to `uca1400_ai_ci` by default (the new `character_set_collations` mapping), and comparing the two throws.

**To Reproduce**

1. MariaDB >= 11.6 with default `character_set_collations` (mine: 11.8.8)
2. RomM 5.0.0 (fresh or migrated from 4.9.x; my DB collation is `utf8mb4_unicode_ci`)
3. Open the collections page, or `curl /api/collections/virtual?type=genre`
4. HTTP 500 with the error above

**Expected behavior**
Virtual collections list as in 4.9.x.

**Screenshots**
n/a (API traceback above; can provide the full traceback on request)

**Server (please complete the following information):**

- OS: Unraid 7.2.4, Docker (`rommapp/romm:latest`, digest `91f6611e`)
- DB: external MariaDB 11.8.8 (`lscr.io/linuxserver/mariadb`), shared instance, dedicated `romm` database

**Client (please complete the following information):**

- Any (API-level failure); confirmed with curl and the web UI

**Additional context**
Verified workaround: restore the pre-11.6 mapping on the MariaDB server and restart RomM. All virtual types return 200 afterward.

```
SET GLOBAL character_set_collations='utf8mb4=utf8mb4_general_ci';
```

(persist under `[mysqld]` in a cnf to survive restarts)

[...]

PR #4374, opened 2026-09-06
fix(db): make virtual_collections and generated columns collation-agnostic

gantoine:
Stacked on #4371 — base will retarget to `master` once that merges. Review the second commit.

**Description**

Following up on #4350, I added a utf8mb3 cell to the migrations workflow and it immediately failed on two *more* collation bugs that CI has never covered. A MariaDB server started with `--character-set-server=utf8` makes the `roms` table, string literals and JSON_TABLE columns utf8mb3, and three separate places assume otherwise.

**1. `0096` / `0101` — `COLLATION 'utf8mb4_general_ci' is not valid for CHARACTER SET 'utf8mb3'`**

The #3758 fix pins `COLLATE utf8mb4_general_ci` on the `virtual_collections` output columns. An explicit utf8mb4 collation on a utf8mb3 value is rejected outright, so on such a server the upgrade dies at `0096`, before it even reaches `0098`. Each operand now goes through `CONVERT(... USING utf8mb4)` first, which keeps the strong coercibility #3758 needed. I confirmed the view still reports `utf8mb4_general_ci` for `name`, `type` and `description` on a utf8mb3 database, so that fix is intact.

The single-argument `CONCAT(j.genre)` wrappers existed only to carry the `COLLATE`; `CONVERT` supersedes them, so they're gone.

**2. `0098` — `Illegal mix of collations ... for operation 'in'`**

Beyond the `JSON_TYPE` case in #4371, `JSON_UNQUOTE` output is compared against string literals in `NOT IN`, `REGEXP` and `NULLIF`. `JSON_UNQUOTE` resolves under the connection charset while a generated-column literal takes the *table's* collation, so those are an illegal mix too. They now route through one helper, `_maria_text`, which wraps the value in `CAST(... AS CHAR)`.

Worth noting this is version-dependent, which is why #4371 alone looked sufficient: on 12.2.2 `JSON_UNQUOTE` returns `utf8mb4_bin` and coerces cleanly, so only the `JSON_TYPE` comparison fails. On 11.3.2 it returns the connection collation and the `NOT IN` lists fail as well.

**3. CI coverage**

`migrate-mariadb` becomes a matrix over `utf8mb4_general_ci` and `utf8mb3_unicode_ci`. Service containers only take `docker create` options and cannot pass `--character-set-server`, so the server is started from a step instead.

MariaDB also moves to `11.3.2` everywhere in CI (was `10.11`), matching `docker-compose.yml`. That image dropped the `mysqladmin` symlink, so the service health checks move to the image's own `healthcheck.sh --connect --innodb_initialized`.

**Testing**

On `mariadb:11.3.2`, both matrix cells reach `0121_state_disc_file`. Seeding identical roms into a `utf8mb4_general_ci` and a `utf8mb3_unicode_ci` database produces byte-identical `generated_genres`, `generated_age_ratings`, `generated_first_release_date`, `generated_average_rating` and `generated_player_count`, and identical `virtual_collections` rows — so there is no behavior change on a default setup.

Each of the three failures was also observed first without the fix, at the migration named above.

**Checklist**

<sup>"Tests" here is the CI matrix rather than pytest: reproducing any of this needs a MariaDB server started with a non-default `--collation-server`, which the pytest fixtures have no way to express. The matrix fails on every one of the three bugs without the accompanying fixes.</sup>

**AI assistance disclosure**

This change was written with AI assistance (Claude Code). The failures were found by running the new matrix cell locally against real MariaDB containers, and each fix was verified the same way before being committed.


gantoine:
Folded into #4371 — both commits now live on `fix/generated-columns-json-type-collation`, and #4371's description covers the full set of collation fixes. GitHub auto-closed this as merged once the commit landed on its base branch.

PR #4377, opened 2026-09-06
test(streaming): refuse broker connections instead of waiting out timeouts

sdornan:
**Description**

The pytest workflow went from about 4 minutes to about 20 after #4314 landed. Parsing the CI job log timestamps showed one module, `backend/tests/endpoints/test_streaming.py`, accounting for roughly 1,250s of worker time, with 18 tests each stalling for 60 to 65 seconds.

Root cause: the module points every fixture container at a broker host nothing answers (`192.168.1.10`). Any broker call a test left unpatched hung for its full socket timeout: 5s for the stop ack inside the test body, and 60s for the detached save pull that the release teardown spawns, which the `TestClient` shutdown then waited on in fixture teardown.

Changes:

- `test_streaming.py`: add an autouse fixture that makes `urllib.request.urlopen` raise a connection-refused `URLError` for the whole module and zeroes `PULL_RETRY_DELAY`. The unreachable-broker path is exercised exactly as before, just immediately. Tests that need a broker reply keep their own inner `urlopen` patches, which take precedence.
- `test_db_handler.py`: seed the 1,050 ROMs of the `bulk_mark_present` chunking test in one transaction instead of a commit per row (40s down to under 1s).
- `pytest.yml`: add `--durations=25` so the slowest tests show up in the CI log and a regression like this is visible.

Measured with `-n 4` against MariaDB: the streaming module runs in about 50s wall (no single test above 1s), and the whole suite in 3m23s, versus the roughly 20 minutes the CI job currently takes.

**AI assistance disclosure:** this PR was written primarily by Claude Code (investigation, code changes, and this description), with the author reviewing.

**Checklist**

N/A

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -88,4 +88,4 @@ jobs:
         run: |
           cd backend
           # GitHub-hosted Linux runners have 4 vCPUs; run one worker per core.
-          uv run pytest -n 4 -vv --maxfail=10 --junitxml=pytest-report.xml --cov --cov-report xml:coverage.xml --cov-config=.coveragerc .
+          uv run pytest -n 4 -vv --maxfail=10 --durations=25 --junitxml=pytest-report.xml --cov --cov-report xml:coverage.xml --cov-config=.coveragerc .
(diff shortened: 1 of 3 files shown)


sdornan:
I saw there is already a PR that basically does the same thing as this one. Feel free to close.

Issue #4395, opened 2026-09-08
[Bug] High ram usage in idle


NotaInutilis:
**RomM version**
5.2.0

**Describe the bug**
High RAM usage in idle. 1,3Go after startup, went as high as 3Go. I had to limit it to 2Go.

**To Reproduce**
No particular steps.

**Expected behavior**
https://docs.romm.app/5.2.0/about/faqs/#how-much-ramcpu-does-it-need
Doc states it should need no more than 500Mo for my usecase: 1 user an 400 files.

**Screenshots**

**Server (please complete the following information):**

- OS: DSM / docker
- Version 7.4
Using postgresql

Found no hints in the logs

NotaInutilis:
I've managed to get some more info. After 23h up, the container was stabilized to 300Mo of ram, which is acceptable.

I found a way to look at the different processes in the containers, so I rebooted it. At boot time, the valkey process directly starts with 750Mo and ramps up to 1Go, then stays there even after startup's activity peak with little to no CPU activity.

titandrive:
Do you have launch box enabled?

Redux28:
I am also having this issue, container was using over 4 GB of ram, restarted the container yesterday and now its at 1.71 GB. 
Launchbox is enabled in mine.

PR #4426, opened 2026-09-10
perf(metadata): compress the cache-only metadata dump records

sdornan:
**Description**

Takes the metadata dump stores from 499MB to 309MB. Part of #4395, and now rebased onto master since #4416 merged.

The LaunchBox and Switch TitleDB stores are read by exact key and never scanned, so each value can be compressed on its own. Values become zstd level 3; a payload under 256 bytes stays plain JSON, because below that a frame header plus a poor ratio on a short payload costs more than it saves. That threshold is what leaves the id-valued title, product id and alternate name indexes untouched.

Compression changes the shape a store holds, so **both stores go to `version=2`** on the `VersionedCacheStore` #4416 introduced. A store an older release wrote is dropped at startup and re-imported, which is what that stamp is for: without the bump, an install with the scheduled update off would keep an uncompressed store forever and never see the saving.

`decode` still detects the zstd frame, but for a different reason than compatibility: a payload under the threshold stays plain JSON, so one store mixes both encodings by design.

A zstd frame is not UTF-8, so the dump keys move to a cache client that leaves values as bytes. `async_cache` still serves everything else, including the schema stamp and the existence checks.

**Measured** on a full import of both real dumps (`Metadata.zip`, `blawar/titledb` `US.en.json`) into the bundled valkey 7.2.12, via the new `backend/tools/measure_dump_cache.py`:

| Store | Fields | Before | After | Saved |
| --- | --- | --- | --- | --- |
| `launchbox_metadata_database_id` | 188,081 | 201.2MB | 124.6MB | 38% |
| `launchbox_metadata_image` | 181,891 | 147.4MB | 71.4MB | 52% |
| `switch_titledb` | 37,350 | 82.7MB | 46.0MB | 44% |
| `launchbox_metadata_name` | 188,060 | 17.9MB | 17.9MB | 0% |
| `launchbox_metadata_folded_name` | 187,507 | 17.4MB | 17.4MB | 0% |
| `launchbox_files` | 85,382 | 12.3MB | 12.2MB | 1% |
| `launchbox_metadata_alternate_name` | 62,383 | 4.9MB | 4.9MB | 0% |
| `launchbox_mame` | 48,781 | 4.3MB | 4.3MB | 1% |
| `switch_product_id` | 29,508 | 2.7MB | 2.7MB | 0% |
| `launchbox_platforms` | 190 | 0.2MB | 0.1MB | 40% |
| **`used_memory`** | | **499.0MB** | **309.4MB** | **38%** |
| **`used_memory_rss`** | | **516.4MB** | **323.0MB** | **37%** |

Stored value bytes go from 491.1MB to 301.6MB. The stores that barely move are the ones whose values sit at or below the threshold, by design. RSS is measured one pass per freshly started server (`--only`), because the pages a delete frees stay resident and cannot be netted out.

These figures are lower than an earlier revision of this description quoted, because #4416's later commits projected three stores down to the fields their readers read before compression touched them: `launchbox_mame` fell 32.3MB to 4.3MB, `launchbox_files` 18.5MB to 12.3MB, and `launchbox_metadata_alternate_name` 9.9MB to 4.9MB. The tool mirrors those projections, or it would measure records production no longer writes.

**Cost.** Reads pay decompression on a hit and nothing on a miss, and most lookups miss:

| Store | `decode` before | after | mean stored bytes |
| --- | --- | --- | --- |
| `launchbox_metadata_database_id` | ~4.6us | ~7.8us | 976 -> 581 |
| `launchbox_metadata_image` | ~4.1us | ~6.5us | 697 -> 319 |
| `switch_titledb` | ~8.0us | ~13.9us | 2,059 -> 1,122 |

Each run samples different records, so these move run to run; the added cost is roughly 2-6us per record read, less than the `json.loads` already on that path, against provider HTTP calls of 100ms and up during a scan. The import went from 38s to 41s.

The measurement tool touches only the ten dump stores, refuses a server holding anything else unless forced, and reports counts and byte totals read back off the hashes rather than tallied while writing, since the dump collides on ~9,000 fields.

**On the codec choice**, measured on 15,000 real `<Game>` records so it is not a guess: the zstd level barely matters here (3 to 19 buys 2.5% for 13x the compression cost, because a ~1KB payload gives a deeper search almost nothing to work with), zlib-6 is 4.9% smaller but 5x slower to decompress and 6.7x slower to compress, and lzma is worse on both axes at this size. The real headroom is a trained dictionary, which reaches 2.92x against 1.65x on a disjoint sample and decompresses faster, but it needs the dictionary stored, versioned and looked up by the id in each frame, with a mismatch treated as a cache miss. Now that `VersionedCacheStore` exists that is a smaller job than it was, but it is still a follow-up rather than this PR.

**Checklist**

On testing: the measurements above are real, taken against the real dumps after merging master.

The full suite could not be run in the authoring environment, where the only CPython 3.14 available is 3.14.0rc2 and the pinned pydantic fails on it during type evaluation (`assert isinstance(value, typing.ForwardRef)`). That reproduces on an untouched master commit, so it is environmental, but it does mean CI is the first place the tests run. It has earned its keep twice: it caught a test that pinned `json.dumps`'s exact output before the codec existed, and, after the merge, two test modules still importing the store keys from a module the merge had emptied of them.

N/A, backend only.

---

**AI assistance disclosure**

This change was written with AI assistance (Claude Code), to a substantial extent: the investigation, the code, the tests, the measurement tool and this description were produced by the agent. The memory, latency and codec figures are measurements the agent took by importing the real LaunchBox and TitleDB dumps into a local valkey, before and after the change.

https://claude.ai/code/session_01CQNNQHk7uyS3eoB9yAxoM9


sdornan:
Addressed the Greptile finding on `backend/tools/measure_dump_cache.py` in cbd1c1f48.

Valid, and the risk was worse than a wrong-URL hazard: RomM's sessions and RQ queues share the database with the dump stores, so `flushall` here would have dropped live job state, not just a rebuildable cache.

- Deletes only the ten keys it writes, via a `STORE_KEYS` tuple that is the sole thing it is allowed to delete. No `flushall` anywhere now.
- Refuses a server holding any key outside those stores, with `--force` to override, since even the scoped delete makes a running RomM re-import its LaunchBox data.
- Each pass records what the server uses *without* those stores and reports figures net of that baseline, so a non-empty server still measures what the dumps themselves cost rather than needing an empty one.

Verified: against a server seeded with one unrelated key the tool exits 1 and leaves that key in place; on a clean server the measurement reproduces, with per-store figures identical and `used_memory` 538.0MB to 334.8MB (the ~1MB shift from the earlier run is the baseline now being netted out). The tables in the description are updated to match what the tool prints today.

---

sdornan:
Both Copilot findings on `backend/tools/measure_dump_cache.py` were correct. Fixed in 5fa244bac.

**Write attempts counted as fields.** Verified against the loaded server before changing anything, and the collisions are real: the dump repeats a title on one platform and an alternate name across platforms, so 8,997 writes landed on a field another record already held. The alternate name store reported 70,297 fields against 62,384 actually stored, and `folded_name`, `files` and `name` were off by 608, 254 and 21. Counts and byte totals now come from `HLEN` and a scan of each hash, so they cannot drift from what the store holds.

**RSS offset by a `used_memory` baseline.** Correct, and following the suggestion exposed something worse: netting RSS is invalid *however* the baseline is captured, because the pages a delete frees stay resident. Capturing an RSS baseline separately made the encoded pass report **-56.1MB**. RSS is now reported absolute, and a new `--only` flag loads a single encoding so RSS can be read off a server started fresh for that pass.

Neither error touched the headline figures, since per-store sizes come from `MEMORY USAGE ... SAMPLES 0` and `used_memory` from `INFO`, and neither was affected. What did change:

| | before | after |
| --- | --- | --- |
| total fields | 1,017,887 | **1,008,890** |
| stored value bytes | 419.3MB -> 234.5MB | **418.6MB -> 233.9MB** (1.79x) |
| `used_memory` | 538.0MB -> 334.8MB | unchanged |
| `used_memory_rss` | 553.3MB -> 350.2MB | **555.8MB -> 348.9MB**, one pass per fresh server |

The description is updated to these numbers. The per-record decode timings move by 1-2us between runs because each run samples different records, so treat them as approximate.

---

sdornan:
Trimmed in ec002748c. Copilot is right on the substance: `CLAUDE.md:65` allows a docstring one sentence plus `Args:`/`Returns:`, and the module docstring had grown to four paragraphs. Every docstring in the file is now one prose line, except `clear_stores` which is one sentence plus a `Returns:` block.

The reasons worth keeping moved to one-line comments at the code they explain, which is where a reader looking at that code needs them, rather than being dropped:

- why the dump walk is duplicated rather than imported, at `iter_launchbox`
- why counts come off the hashes instead of the write loop, at `store_stats`
- why it is not `flushall`, at `clear_stores`
- the `SAMPLES 0` rationale, already a one-line comment at `measure`

Two corrections for the record, neither of which changes the outcome:

- The secondary line references are off. Lines 45 and 71 are single-line comments (`# Mirrors the keys the two update tasks write.`), inside the limit. The other actual offenders were `store_stats` and `clear_stores`, and both are trimmed.
- Moving the operational detail to `--help` would have been a no-op, because `ArgumentParser(description=__doc__)` makes the module docstring *be* the `--help` text. Shortening the prose was the fix; the operational detail a reader needs at the command line is carried by the `--force` and `--only` help strings and by the refusal message.

`--help` still names the purpose, the usage line and every flag, and the tool runs unchanged.

---

sdornan:
Took Copilot's duplicated-constants finding on the store keys, in 97f5cd42d.

Agreed, and it reads worse than a style point now the tool *deletes* those keys: after a production rename it would have deleted stale hashes, missed the new ones, and counted the real stores as foreign data in its own safety guard.

The ten keys now live in `handler/dump_cache.py`, which already holds the codec that reads and writes them and depends on nothing but the cache client. The importers, the lookups and the tool all take them from there, and `DUMP_STORE_KEYS` is the list the tool measures and clears. That also dissolves the reason I had copied them: `launchbox_handler/types.py` pulls in fastapi through `BaseRom`, so a tool cannot import it.

Two things fell out of it:

- `base_handler` was importing two constants from `tasks/scheduled/update_switch_titledb`, so a handler depended on a task module. That import is gone, and with it the cycle risk that made the ordering of those two imports load-bearing.
- `launchbox_handler/types.py` keeps `LAUNCHBOX_METADATA_INITIAL_IMPORT_KEY`, which marks an import in progress rather than naming a store, so it is not part of the moved set.

Verified: no `romm:launchbox_*` or `romm:switch_*` literal is declared outside `dump_cache` now, `dump_cache` still imports only `json`, `typing`, `zstandard` and the cache client, and a full re-measurement is byte-identical (538.0MB to 334.8MB, every per-store figure the same).

**Declining the `GAME_IMAGE_FIELDS` half of that finding.** It selects which XML fields the importer keeps, which is an importer concern rather than a cache-key one, so centralising it in the codec module would put it at the wrong layer. It is also #4416's constant, and relocating it from this PR would edit that change. The tool's copy is one line and annotated as mirroring it.

---

gantoine (approved):

Issue #4423, opened 2026-09-10
[Bug] missing_from_fs can never become true for a save or a state


Spinnich:
**RomM version**

5.2.0 (`master`, verified at `bdf42435d`). Not a regression — as far as I can tell it has never worked.

**Describe the bug**

`missing_from_fs` on `saves` and `states` can never become `true`. The functions that would set it exist but are never called, and the scan that would call them never looks at save or state files in the first place.

Delete a save or state from disk and RomM keeps listing it as present. Downloading it then fails at the filesystem.

**Where it breaks**

`RomAsset.missing_from_fs` is flipped by a family of five `mark_missing_*` handlers. Three of them are wired up; the two asset ones are not:

| Handler | Defined | Called from production code |
| --- | --- | --- |
| `mark_missing_roms` | `handler/database/roms_handler.py:2126` | yes — `endpoints/sockets/scan.py:961`, `:1044` |
| `mark_missing_firmware` | `handler/database/firmware_handler.py:105` | yes — `scan.py:1052` |
| `mark_missing_platforms` | `handler/database/platforms_handler.py:128` | yes — `scan.py:1284` |
| **`mark_missing_saves`** | `handler/database/saves_handler.py:210` | **no — zero call sites anywhere** |
| **`mark_missing_states`** | `handler/database/states_handler.py:130` | **no — zero call sites anywhere** |

`grep -rn "mark_missing_saves\|mark_missing_states" backend/` returns only the two `def` lines. Not a caller, not a test, nothing.

The root cause is a level up: **the scan never enumerates saves or states on disk.** There is no `scan_save` or `scan_state` call in `endpoints/sockets/scan.py`, so there is no `saves_to_keep` / `states_to_keep` list to hand these functions even if someone wanted to call them. Both take `(rom_id, user_id, *_to_keep)`, so they'd also need a per-user walk that the current per-platform scan loop doesn't do.

**To Reproduce**

1. Play a game in the browser player so RomM writes a save and a state.
2. Delete the save file from the library on disk.
3. Run any scan type.
4. The save is still listed in the ROM's Save data tab with no missing indicator, and `GET /api/saves` still reports `missing_from_fs: false`.

**Expected behavior**

Either the scan reconciles saves/states against the filesystem the way it does ROMs, firmware and platforms — or, if that is deliberately out of scope, the two dead handlers and the `missing_from_fs` field on `SaveSchema` / `StateSchema` should go, so the API stops advertising a signal it never produces.

**Additional context**

[...]

Issue #4421, opened 2026-09-10
[Feature] Let companion apps sync save states, not just in-game saves


Spinnich:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when my save states stay stuck on whichever device made
them. In-game saves follow me between my handheld, my phone and the browser.
Save states don't — they only move if the browser player made them.

RomM's sync protocol covers saves only. The `/api/sync/negotiate` flow takes a
list of the client's saves, compares each against the server's, and returns
upload / download / conflict / no-op per save. Save states are not part of it at
any point — the sync endpoints, the sync watcher, the push-pull task and the
comparison handler never touch state records at all.

So a companion app has two bad options: ignore save states entirely, or hand-roll
its own sync on top of the plain `/api/states` endpoints, with no conflict
detection, no pairing, and no way to tell whether the server's copy is the same
one it already has.

In practice they pick the first. From #4278, describing a three-device setup
(browser, Android, Steam Deck):

> RomM can store both types of saves and already does so for EmulatorJS, while
> Argosy and Tender currently sync only in-game saves.

That's the gap. A save state made on a handheld never reaches RomM, and a save
state RomM has never reaches the handheld.

**Describe the solution you'd like**

Extend the sync protocol to cover save states, so a companion app can negotiate
them the same way it negotiates saves: tell the server what states it has, get
back what to upload, what to download, and what's already in sync.

For a self-hoster this is the difference between "my progress follows me" and "my
progress follows me unless I used a save state." If you play on a handheld with
auto-resume and then open the same game in the browser, you should land where you
left off.

Alongside it, states would want the same provenance saves get: which device
produced a state, and which devices currently have it.

**Describe alternatives you've considered**

- *Let each companion app roll its own state sync on `/api/states`.* Possible
  today — those endpoints even have a `rom_ids` parameter described as being
  "for clients syncing a known set of ROMs" — but every client would reinvent
  conflict resolution, and they'd each get it subtly differently. The whole point
  of the negotiate flow is that the server decides.
- *Sync states as if they were saves.* Tempting, since the machinery exists, but
  they don't behave alike. Saves pair on `(rom_id, slot)` and a null slot means
  "archival"; states have no slot concept in the same sense, are far more
  numerous (RomM keeps every capture), and are tied to a specific emulator and
  often a specific core version in a way saves aren't. Pairing and retention
  rules need thinking through rather than inheriting.
- *Solve it at resume time instead* (#4278). That issue is about an older save
[...]

Issue #4420, opened 2026-09-10
[Feature] Show which device a save came from and where it's synced


Spinnich:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I look at a save in the web UI and can't tell where it
came from or whether my handheld has it yet.

If you sync saves with a companion app, RomM already records this. When a device
uploads a save it stores which device it was (`Save.origin_device_id`), and every
time a device syncs a save it records that too, in a `device_save_sync` table
holding the device, the timestamp, and whether the sync is current.

None of it is visible in the GUI. Both fields exist on `SaveSchema` and the
save-list endpoint even computes a friendly `device_syncs` payload — device name,
last synced at, `is_current` — but only for API callers that authenticate *as a
device*. A browser is not a device, so it always gets an empty list, and
`origin_device_id` arrives as a bare id with nothing rendering it.

So the person most likely to want the answer — someone sitting in front of the
web UI wondering "did my Deck get this save yet?" — is the one person who can't
get it.

**Describe the solution you'd like**

Show save provenance and sync status in the Save data tab, and degrade gracefully
when it isn't there:

- **Where a save came from.** If `origin_device_id` is set, show the device's
  name. Saves uploaded through the browser or written by the built-in player have
  no origin device — those simply show nothing extra.
- **Where a save has synced to.** If a save has sync records, show the devices and
  how current each one is ("Steam Deck, up to date"; "Phone, 3 days ago"). A save
  no device has touched shows nothing.

The rule throughout: render what's populated, show nothing when it isn't. Nobody
who doesn't use a sync client should see an empty "Devices" row, and nobody who
does should have to guess.

Worth being explicit about what this costs, because the two halves differ:

| | What's needed |
| --- | --- |
| **Origin device** | **Frontend only.** `origin_device_id` is already on `SaveSchema` and already reaches the browser. Resolving it to a name needs `GET /api/devices`, which is in the default read scopes and already returns only the caller's own devices. |
| **Sync status** | **Needs a backend change.** `device_syncs` is deliberately left empty for non-device callers, so the ROM detail response would have to start populating it. No migration — the `device_save_sync` table already holds everything. |

**Describe alternatives you've considered**

- *Show the raw `origin_device_id`.* It's a device id, not a name — useless on
  its own, and it's already technically "in the UI" in the sense that it's in the
  page's data. The name is the whole point.
- *Leave sync status to the companion apps.* They can already see it, since
[...]

Issue #4419, opened 2026-09-10
[Feature] Expose which slot and which disc a save state belongs to


Spinnich:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I look at a list of save states for a streamed
multi-disc game and can't tell which disc — or which slot — any of them belongs
to. RomM knows both. It just doesn't tell anyone.

Two concrete cases:

- **Discs.** Streaming supports disc swapping, and when you capture a state RomM
  records which disc was mounted, so a later resume can put the same one back
  (`State.disc_file_id`, added in migration `0121`; used by
  `handler/streaming/launch.py` on resume). That's genuinely useful behaviour —
  but the value is invisible. Looking at four states for a two-disc PS2 game,
  nothing tells you which are on Disc 1 and which are on Disc 2. You pick one and
  find out by launching it.
- **Slots.** Streaming emulators write states into numbered slots, and RomM
  already parses the slot out of the filename to tell the broker what to load
  (`slot_from_state_filename()` in `handler/streaming/states.py`, covering PCSX2
  `.NN.p2s`, Dolphin `.sNN`, xemu `.xNN` and RetroArch `.stateNN`). It derives
  the number, uses it, and throws it away. Saves expose `slot` on their schema;
  states don't.

So in both cases the server has the answer and no client can get it. Neither
field is on `StateSchema`, so this affects the web UI and any API consumer
equally.

**Describe the solution you'd like**

Add `slot` and `disc_file_id` to `StateSchema`.

Both are nullable and additive — nothing that reads states today would change.
Neither needs a database migration: the slot is derived from the filename by code
that already exists, and the `disc_file_id` column has been on the `states` table
since `0121`.

For a self-hoster this is the difference between a save-state list you can read
and one you have to launch to understand: "Disc 2, slot 3" instead of a filename
that's been ellipsised down to the game's name.

**Describe alternatives you've considered**

- *Have clients parse the slot out of the filename themselves.* That means every
  client reimplementing the per-emulator filename rules the server already
  implements, and re-doing it whenever a new emulator lands. Wrong side of the
  wire.
- *Return the disc's name instead of its id.* Probably unnecessary —
  `DetailedRomSchema.files` is already on the ROM detail response and the Files
  tab already reads it, so a client can resolve the id locally. Happy to be told
  otherwise if a flat name is preferred for non-web consumers.
- *Fold this into #4418.* That one also adds a field to `StateSchema`, so it was
[...]

PR #4429, opened 2026-09-10
feat(v2): show save states with the same detail as saves

### LINKED ISSUE #4320: [UI/UX] Saves and save states don't show the information RomM already has about them
Saves and save states both carry more information than the v2 GUI shows. In some
places the information is fetched and thrown away; in others it's rendered but
truncated to the point of being unreadable.

**This issue is only about data the API already returns today.** No backend
change, no schema change, no migration — every item below is a rendering fix in
`frontend/src/v2/`. Things that need the backend to expose something first are
listed at the bottom as explicitly out of scope.

**How save states are currently displayed in the v2 GUI**, note the lack of
accessible information.
Image

**How saves are currently displayed in the v2 GUI.**
Image

---

### The problem

**1. The states subtab renders strictly less than the saves subtab — and less
than the old UI did.**

The two subtabs go through different components: saves through `AssetList`
(rows), states through `AssetStrip` (tiles). The state tile renders filename,
relative date and size. It does *not* render the emulator, and it does not render
the exact timestamp. The saves row renders both.

Both values are already in `StateSchema` and already arrive in the browser. They
are simply never painted.

The old UI painted them. `components/common/Game/AssetCard.vue` shows, for the
same state object: the full wrapping filename, an orange emulator chip, a size
chip, and `Updated: <exact timestamp>` with the relative date underneath. So this
is a regression in the new UI, not a feature that was never built.

**2. The truncated filename is unreachable, not just clipped.**

The tile name is `text-overflow: ellipsis` inside a `minmax(150px, 1fr)` grid.
Normally you'd hover for the rest — but the tooltip is gated on `v-if="selectable"`,
and the Save data tab renders in manage mode (`:selectable="false"`). There is no
tooltip, and no `title` attribute either. No hover, no focus, no keyboard path to
the full name.

That's worse for states than it sounds, because both naming schemes put the
*distinguishing* part last, exactly where the ellipsis eats it:

- Browser player: `Super Mario World [2026-09-01 13-35-45-123].state` — every
  state for a game shares the ROM-name prefix, and the timestamp that tells them
  apart is the tail.
- Streaming: `SLUS-12345 (ABCD1234).20260901-133545123456.03.p2s` — same shape,
  stamp and slot at the tail.

So you get a column of tiles that all read identically.

**3. Save hashes are fetched and discarded.**

`SaveSchema.content_hash` is returned on every save and is not rendered anywhere
in the frontend — even though the UI already has a click-to-copy `HashChip`
component, used by the metadata, files and firmware tabs. If you want to check
whether the server's copy of a save matches the one on your handheld, the value
is already sitting in the page's data and there's no way to see it.

### Describe the solution you'd like

For a self-hoster with a pile of save states on one game, this is the difference
between a wall of identical tiles and a list you can actually read: tell two
states apart at a glance, see which emulator made one before you try to resume
it, and read a save's hash without opening devtools.

Four changes, all frontend, all against data already in the response:

1. **Render `emulator` on the state tile**, as a chip, the way the saves row and
   the old UI both do.
2. **Render the exact timestamp** alongside the relative one on states, the way
   the saves row and the old UI both do.
3. **Give the Save data tab a tooltip in manage mode**, so the full filename and
   exact timestamp are reachable by hover *and* keyboard focus. (Un-gating the
   tooltip that already exists helps the saves rows too, which truncate long
   filenames the same way.)
4. **Show `content_hash` on save rows** using the existing `HashChip`.

Plus one layout change that makes 1 and 2 fit:

5. **Render states through the same row component saves use**, moving the
   capture into the row's leading cell (widened to 16:9) rather than leaving it
   as a 150px tile. The screenshot survives, the filename gets the full row
   width, and the emulator chip and exact timestamp have somewhere to go.

> **Revised 2026-09-10.** Item 5 originally proposed a persisted layout toggle
> for the states subtab, mirroring the streaming player's. I've since talked
> myself out of it and swapped in the row component instead; the reasoning is in
> Alternatives below.

### Describe alternatives you've considered

- **A layout toggle for the states subtab**, the way the streaming player
  already has one: keep the tile view, add a denser list view beside it. This is
  what item 5 originally asked for. I've dropped it because a toggle asks every
  user to find and change a setting in order to fix a default that is simply
  worse, and it leaves two layouts to maintain. Once the row carries the capture,
  there is nothing the tile view does better.
- **Render states with the saves component and drop the screenshot entirely.**
  Wrong, and for exactly the reason #3655 was filed: people navigate by
  thumbnail. Item 5 keeps the capture, just in the row's leading cell at 96px
  instead of as a 150px tile.
- **Cram the extra chips into the 150px tile and leave the layout alone.** At
  11px there isn't room for name + emulator + size + relative + exact. It would
  just trade one truncation for another. That's why item 5 is here.
- **Only fix the tooltip and leave the tile alone.** Cheapest option, and it does
  make the data reachable — but it needs a hover per tile to answer "which of
  these is the Dolphin one", which is a poor substitute for showing it.
- **Solve identification with labels/renaming instead** (#942, #2653). I'd still
  like both, but they ask the user to hand-annotate things RomM already knows.
  This is the cheap half, and it makes those features less necessary rather than
  competing with them.

### Deliberately out of scope

Things that would also be worth showing, but that the GUI **cannot** show today
because the backend doesn't send them. Each is tracked separately so this issue
stays a frontend-only change:

| | Tracked in |
| --- | --- |
| A state's **slot**, and which **disc** it was captured on | **#4419** — both are absent from `StateSchema`. No migration needed (the slot is already derived by `slot_from_state_filename()`, and `states.disc_file_id` has existed since migration `0121`), but the backend has to send them before anything can render them. |
| A state's **`content_hash`** | **#4418** — the column doesn't exist at all, so this one needs a migration. |
| Which **device** a save came from, and where it's **synced** | **#4420** — `origin_device_id` needs a device-name lookup, and `device_syncs` is currently only populated for device-scoped API callers. |
| Device provenance for **save states** | **#4421** — states aren't part of the sync protocol at all, so there's nothing recorded to show yet. |

One thing genuinely rejected rather than deferred: `missing_from_fs`. It's on
every asset schema, but it is never set on saves or states —
`mark_missing_saves` / `mark_missing_states` are defined and never called — so a
"missing" badge would decorate a flag nothing writes. That may be its own bug,
but it isn't this issue's.

### Additional context

One possible approach for item 5, offered as a starting point rather than a
demand — the maintainers should decide the shape:

- The layout toggle already exists elsewhere. `views/Player/Stream.vue` defines
  `STATE_LAYOUTS` (row / grid / list) persisted in `localStorage`, with the
  comment *"Every capture is kept, so a heavy save-stater ends up with a history
  the horizontal strip buries."* That's this issue, already solved once — in the
  player, not in the Save data tab. `AssetStrip`'s `list` layout even describes
  itself as *"rows with no thumbnail and the meta split into columns. Fits the
  most entries per pixel, for long save-state histories."* The `play.states-view-*`
  translation keys already exist in all 18 locale directories, so reusing this
  needs no new strings.
- One snag: `AssetStrip`'s `list` layout hardcodes `max-height: 340px; overflow-y:
  auto`. `AssetList` solved the same thing with a `scrollable` prop that the tab
  sets to `false`. `AssetStrip` would want the same prop, or the tab ends up with
  a scrollbox inside a scrollbox.

---

*This analysis was AI-assisted: the codebase investigation and prior-art search
behind this issue were done with Claude Code, and reviewed by me before filing.*

### LINKED ISSUE #4422: [UI/UX] The Play view hides screenshots that in-game saves actually have
### Summary

The v2 Play view assumes in-game saves never carry a screenshot. They do. Every save the browser player writes has one, and the pre-game UI throws it away in favour of a generic floppy-disk icon.

### The assumption

Two v2 components state it outright:

- [`src/v2/components/shared/AssetStrip.vue`](https://github.com/rommapp/romm/blob/master/frontend/src/v2/components/shared/AssetStrip.vue#L122) gates the tile thumbnail on `type === 'state'`:

  ```vue
  <div v-if="type === 'state' && screenshotOf(asset)" class="r-asset-strip__thumb-img" ... />
  ```

  `screenshotOf()` itself is type-agnostic and would happily return a save's screenshot; the template refuses to ask.

- [`src/v2/components/Player/AssetPreview.vue`](https://github.com/rommapp/romm/blob/master/frontend/src/v2/components/Player/AssetPreview.vue) says it in its header comment — *"Save: a featured save graphic — saves never carry a screenshot, so we lean on the icon + decorative backdrop"* — and its `v-else-if="asset && type === 'save'"` branch renders a decorative floppy medallion rather than the shot.

### Why it's wrong

Saves and states resolve their screenshot through the **same** code, and the player uploads one for both:

| | Save | State |
| --- | --- | --- |
| model | `Save.screenshot`, `cached_property` keyed on `(rom_id, user_id, file_name_no_ext)` (`backend/models/assets.py:121`) | `State.screenshot`, identical lookup (`:157`) |
| upload endpoint | `POST /api/saves` takes `screenshotFile` (`backend/endpoints/saves.py:173`) | `POST /api/states` takes `screenshotFile` (`endpoints/states.py:46`) |
| response schema | `SaveSchema.screenshot: ScreenshotSchema \| None` | `StateSchema.screenshot: ScreenshotSchema \| None` |
| written by the player | yes | yes |

On the write path, `EJS_onSaveSave` passes the captured frame straight into `saveSave` (`frontend/src/views/Player/EmulatorJS/Player.vue:350`), and the exit-autosave takes **one** `gameManager.screenshot()` and feeds it to the state *and* the save (`Player.vue:542-563`). `saveSave` stores it as `${filename}.png` (`EmulatorJS/utils.ts:114`) — precisely the stem `Save.screenshot` matches on.

So a save made in the browser player has a screenshot on disk, in the database, and in the API response. Only the UI declines to draw it.

### Impact

In the EmulatorJS and Stream pre-game views, the save picker is a column of identical floppy icons. That's the same "which save is this even for?" problem #3830 describes, except here the answer is already sitting in the response.

### Suggested fix

Drop the `type` gate in both components and key off the data — render the screenshot when the asset has one, fall back to the type icon when it doesn't. Correct the two header comments while there.

### Related

#4320 fixes exactly this wrong belief in `shared/AssetList.vue` (the Save data tab), where the same comment appears. This issue covers the two Play-view components that #4320 deliberately leaves alone to keep its diff scoped.

Spinnich:
**Description**

Closes #4320.
Closes #4422.

The states subtab rendered 150px `AssetStrip` tiles that showed only the
filename, a relative date and a size. It dropped the emulator chip and the exact
timestamp that the saves subtab shows for the same object, and that the old UI's
`AssetCard` showed too. Worse, both real state-naming conventions put the
distinguishing part last (`<rom name> [2026-09-01 13-35-45-123].state` from the
browser player, `SERIAL (CRC).20260901-133545123456.03.p2s` from streaming), so
the ellipsis ate exactly the part that tells two states apart, and manage mode
gated the tooltip off, leaving no hover, focus or keyboard path to the full name.

States now render through `AssetList`, the same row list saves use, so they
inherit the emulator chip and the exact timestamp for free. To keep the capture
(#3655), `AssetList`'s leading cell widens from a 36px icon square to a 96px 16:9
thumbnail. That widening is decided **from the data, not from the asset type**:
saves carry a screenshot exactly as readily as states do. `Save.screenshot` and
`State.screenshot` are the same lookup, `POST /api/saves` takes a
`screenshotFile`, and the EmulatorJS player feeds one `gameManager.screenshot()`
into both the state and the save. A `type` gate would have hidden a save's own
capture; three v2 components currently do exactly that, filed separately as
#4422. Rows without a capture keep a type-aware icon on a placeholder gradient so
filenames stay in one column.

Alongside that: the tooltip is un-gated so manage mode can reach the truncated
filename by hover **and** by keyboard focus, and save rows gain a `content_hash`
chip through the existing `HashChip` (the value was already on every `SaveSchema`
and thrown away). On phones the row goes to two lines, because the thumbnail
crushed the name column to ~55px at 390px; it's 256px now, which also helps the
saves list, already tight at that width on master.

The `AssetStrip` deletions are the other half of the swap: `layout="flow"`,
manage mode, the `#actions` slot and the `selectable` prop existed only for the
Save data tab and are now dead. Every remaining caller (the EmulatorJS and Stream
pre-game views) uses the selectable path, so tiles are unconditionally `<button>`
now. The screenshot lookup and the save/state union type that both components had
been declaring separately moved to one shared util.

Hands-on testing then surfaced the other half of the same mistake. The Play
view's `AssetPreview` and `AssetStrip` were gating the capture on
`type === 'state'` too, so a save that had one showed a floppy medallion.
`AssetPreview` even computed the URL correctly and then never used it. That bug
predates this PR, but the states swap turns it into a visible contradiction: the
row now renders the very capture the stage above it hides. Both now key off the
capture and fall back per type only when there isn't one, which closes #4422 and
leaves one rule across all four components.

**Frontend only** - no backend change, no schema change, no `npm run generate`,
no migration. One new i18n key, translated in all 18 locales.

**A note on the approach, since the issue used to argue against it.** #4320
originally proposed a persisted layout toggle and listed "render states with the
saves component" as a rejected alternative, on the grounds that #3655 proves
people navigate by thumbnail. I changed my mind while planning this: a toggle
asks every user to find and configure a setting to fix a default that is simply
worse, and the thumbnail objection goes away once the row carries the capture
rather than an icon. States are still identifiable by picture, they are just also
readable now. The issue has been updated to match, with the reversal called out
rather than quietly rewritten.

**What to look at hardest**

- The un-gated tooltip means focusing a row action shows both the button's own
  tooltip and the row tooltip. I measured the two boxes as non-overlapping and
  left it, because the fallback (re-anchoring the tip to the name element) would
  kill the keyboard path entirely, since the name is not focusable. Opinions
  welcome.
- `AssetList` decides the cell width per instance, but "Mine" and "Community" are
  two instances stacked in one scroll view and read as one table, so
  `SaveDataTab` decides once per subtab and passes the same answer to both via a
  `thumbs` override. The Play view still decides for itself. If you'd rather the
  override not exist, the alternative is living with two sections whose name
  columns don't line up.
- The `AssetStrip` surgery touches the pre-game resume flow, which this PR
  otherwise doesn't change. `AssetStrip.test.ts` locks the tile's `<button>`
  root, `aria-pressed` and `select` emit against exactly that risk, and I
  re-verified the EmulatorJS view in both themes.

Out of scope, tracked separately: a state's slot and disc (#4419) and
`content_hash` (#4418). The chips here are value-gated, not type-gated, so states
light up for free the day either ships. Device provenance is #4420, state sync is
#4421, and `missing_from_fs` gets no badge because nothing ever sets it (#4423).

---

_AI assistance disclosure: this PR was written primarily by Claude Code, under my
direction and review - investigation, implementation, tests and this description.
I directed the design (including overruling the layout-toggle approach my own
issue originally proposed), reviewed every change, and verified the results in a
browser myself. PR responses will be mine unless I say otherwise._

**Checklist**

<sub>Verified locally: `npm run typecheck` clean · `npm run test` 114 files / 1221 passed
(29 of them across five new test files) ·
`npm run build` ✓ · `check_i18n_locales.py` and `check_i18n_sorted.py` ✅ ·
`trunk fmt && trunk check` no issues · `npm run test:e2e` 19 passed / 1 failed,
the failure being `game-media-files.spec.ts:44`, which fails identically with
this branch stashed.</sub>

<sub>Browser-verified with Playwright against a real Chromium as a seeded admin,
in light and dark, with mouse, real `Tab` keypresses and at the `sm`/`xs`
breakpoints. Test data covered states with and without a capture, `emulator:
null`, both real filename shapes, a save carrying a capture, saves with and
without a `content_hash`, community items owned by another user, and the empty
state. The EmulatorJS pre-game view was re-checked for regressions from the
`AssetStrip` deletions, and all 10 Storybook stories render.</sub>

**Before / after, states subtab.** Same five save states either way.

| Before (`AssetStrip` tiles) | After (`AssetList` rows) |
| --- | --- |
| states subtab before | 01-states-dark |

<details>
<summary>Save data tab &mdash; states</summary>

01-states-dark

Dark. The emulator chip and the exact timestamp are back, and the capture leads the row.

06-states-light

Light.

02-states-dark-tooltip

The CSS-truncated filename, recoverable on hover in manage mode.

03-states-dark-keyboard

And on a real keyboard focus, which is the half a hover-only tooltip fails.

</details>

<details>
<summary>Save data tab &mdash; saves (hash chip, thumbnail on a player-written save)</summary>

04-saves-dark

Row 2 is a save the browser player wrote, so it carries its own capture; its screenshot-less siblings keep the floppy. Mine and Community agree on cell width, so the name columns line up.

07-saves-light

Light.

05-saves-dark-hash-click

HashChip revealing the full content hash.

</details>

<details>
<summary>Narrow screens (sm / xs)</summary>

08-saves-sm

sm: the name column holds at 204px.

09-states-xs

xs: the row goes to two lines so the thumbnail cannot crush the name, which keeps 256px. The exact timestamp drops out; the tooltip still has it.

10-saves-xs

Saves at xs.

</details>

<details>
<summary>EmulatorJS pre-game view &mdash; the #4422 fix, and no regression from the AssetStrip cleanup</summary>

The states tab is untouched by this PR; it is here as the regression check for
the manage-mode deletions. Tiles are still selectable buttons.

11-ejs-pregame-dark

12-ejs-pregame-light

Saves tab, with a save that carries a capture selected. This is what used to
render a floppy medallion while the row below it showed the screenshot. The two
capture-less saves still get the floppy, correctly. (The captures here are
synthetic test fixtures, hence the flat gradients.)

**[drop 25-ejs-saves-capture-dark.png here]**

**[drop 26-ejs-saves-capture-light.png here]**

Tooltip on real keyboard focus:

13-ejs-keyboard-focus

</details>


zurdi15:
Im closing this in favor of #4540

PR #4437, opened 2026-09-11
ci: build each arch on a native runner and cache layers

sdornan:
**Description**

Docker image builds took ~20 minutes, and the cause was not the C libraries.

Both `build.yml` and `test-build.yml` built `linux/arm64,linux/amd64` on a single `ubuntu-latest` (x86) runner, so the entire arm64 half ran under QEMU user-mode emulation. That penalty landed squarely on the expensive stages: `rahasher-build` (C++ with libchdr), `sigil-build` (CMake + submodules + cffi), `nginx-build` (mod_zip), and the Python C extensions. Neither workflow set `cache-from`/`cache-to`, so every run recompiled all of it from scratch even though those stages are driven by pinned `ARG`s that change a few times a year.

**Native per-arch runners.** Both workflows now use a matrix: `ubuntu-latest` for amd64, `ubuntu-24.04-arm` for arm64 (GA and free for public repos). Each job pushes an untagged, digest-addressed image, and a `merge` job stitches the digests into a manifest list with `docker buildx imagetools create`. QEMU is gone entirely, `setup-qemu-action` is removed, and the two architectures build in parallel.

`build.yml` keeps both variants: each matrix job builds `slim-image` then `full-image` (the latter is `FROM` the former, so they share nearly all their work inside one job), exports both digests, and the merge job creates one manifest list per variant. Cross-registry tagging still works because `imagetools create` copies blobs when a target sits in a different registry than the source digests, so Docker Hub tags are created from the GHCR digests.

**Registry layer cache.** A dedicated `ghcr.io/<owner>/romm-buildcache` package, keyed by arch:

- releases read and write `release-<arch>`
- preview builds read `preview-<arch>` and `release-<arch>`, and write only `preview-<arch>`

so a preview build can never overwrite the release cache. Only the full build exports the cache, since it is a superset of the slim one. GHCR login is now unconditional in `test-build.yml`, because the cache lives there even for a Docker Hub-only dispatch.

**Sticky preview comment.** The two `github-script` blocks and the `comment-id` passed between jobs are replaced by a comment keyed on a header, so the status updates in place across re-runs instead of appending a new comment each time. With the id plumbing gone, `build` and `merge` no longer depend on `prepare`: the gate moved onto `build` directly, so the build starts alongside the comment rather than behind it, and a failure posting the comment can no longer cancel it.

This also fixes a latent bug. The old comment interpolated the raw branch name, but `metadata-action` rewrites characters Docker forbids in a tag, so a branch like `claude/admiring-sagan-3447xr` publishes as `claude-admiring-sagan-3447xr` and the advertised reference 404s.

**Least-privilege permissions.** These were the only two workflows granting themselves `contents: write` and `actions: write` at the top level; every other workflow in the repo is `read-all`. Nothing here pushes commits or dispatches a workflow in this repo (the docs trigger authenticates to another repo with a PAT), and `validate-tag-semver` was inheriting the whole set to run a regex. Each job already declares what it needs, so the top-level grant is now a read-only fallback. `trigger-docs-and-web` keeps its own `actions: write` rather than my guessing at a release-only job.

The preview checkout also fetched full git history for a build that reads none, once per architecture now that the matrix has two of them. Nothing in the frontend or backend build reads git metadata, so it is gone.

**Dockerfile.** Dropped `git` from `backend-build`. The comment said it was needed for a streaming-form-data fork, but that is an ordinary PyPI dependency now and `uv.lock` has no git sources at all.

**Results**

| Build | Wall clock |
| --- | --- |
| Before (QEMU, one runner, no cache) | 20m 45s |
| After, cold cache | 4m 47s |
| After, warm cache | ~65s |

~65s is the best case, where only workflow YAML changed and every layer hits. A PR touching `backend/` or `frontend/` invalidates from that layer down, so expect something between the two figures in normal use; 4m 47s is the honest floor for a from-scratch build.

**On using prebuilt binaries instead**

Mostly not available, and it matters much less once QEMU is out of the picture. RetroAchievements ships RALibretro for Windows only, and we patch its source before building anyway; nobody publishes a musl `ngx_http_zip_module.so` for a specific nginx version. The one real opportunity is `sigil`, which is our own `rommforge/argosy-sigil`: publishing cibuildwheel musllinux + manylinux wheels there would let this Dockerfile install a wheel and delete the whole CMake stage. Left as a follow-up. On the Python side only `mariadb`, `psycopg-c` and `streaming-form-data` build from source (plus `psutil`/`memray` in the dev image); everything else already resolves to a musllinux aarch64 wheel, and `psycopg[binary]` would be a no-op because psycopg-binary publishes no musllinux wheels.

**Note:** pushing by digest leaves untagged manifests in the `romm`/`romm-testing` packages, which a GHCR cleanup policy can prune.

**Checklist**

**What is and is not proven**

`test-build.yml` has been exercised end to end by four real preview runs on this branch. The published index was inspected directly from GHCR rather than inferred from a green check, and is a correct flat OCI index carrying `linux/amd64` and `linux/arm64` plus both provenance attestations, each re-pointed at its image digest. Leaving provenance at its default is deliberate: buildx flattens the per-arch indexes correctly, so `provenance: false` would have needlessly dropped attestations we publish today.

**`build.yml` has never executed.** It shares the same matrix-and-merge pattern, but nothing triggers it except a published release, so its first real run will push to Docker Hub and GHCR under the project's identity. That is the part worth a careful read; the slim/full digest split and the cross-registry `imagetools create` are the specific things to check.

`actionlint` (with its shellcheck integration) and `prettier --check` are clean on both workflows. No unit tests, as the change is entirely CI and Dockerfile.

n/a

---

AI assistance disclosure: written by Claude Code (agentic, effectively all of the diff and this description), driven and reviewed by me.

https://claude.ai/code/session_01GQKGy6yG6bMwEQGe5bp3ad

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
   release:
     types: [published]
 
-# Every job below grants itself what it needs; this is only the fallback.
 permissions:
   contents: read
 
@@ -123,6 +122,8 @@ jobs:
           labels: ${{ steps.meta-slim.outputs.labels }}
           target: slim-image
           outputs: type=image,name=${{ env.DIGEST_IMAGE }},push-by-digest=true,name-canonical=true,push=true
+          # Deliberately not seeded from the preview cache: those layers are
+          # built from PR head content and must not reach a release image.
           cache-from: type=registry,ref=${{ env.CACHE_IMAGE }}:release-${{ matrix.arch }}
           build-args: |
             SCREENSCRAPER_DEV_ID=${{ secrets.SCREENSCRAPER_DEV_ID }}
@@ -193,9 +194,6 @@ jobs:
           path: /tmp/digests/full
           pattern: digest-full-*
           merge-multiple: true
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
 
       - name: Login to Docker Hub
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
@@ -277,7 +275,7 @@ jobs:
     permissions: {}
     needs: merge
     runs-on: ubuntu-latest
-    if: needs.merge.result == 'success' && github.event.release.prerelease == false
+    if: github.event.release.prerelease == false
     steps:
       - name: Trigger docs build
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,15 +7,15 @@ on:
   release:
     types: [published]
 
+# Every job below grants itself what it needs; this is only the fallback.
 permissions:
-  id-token: write
-  contents: write
-  packages: write
-  actions: write
+  contents: read
 
 jobs:
   validate-tag-semver:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     outputs:
       valid: ${{ steps.check_tag.outputs.valid }}
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,28 +24,40 @@ jobs:
         run: |
           TAG="${{ github.event.release.tag_name }}"
           if [[ "$TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-alpha\.[0-9]+|-beta\.[0-9]+|-rc\.[0-9]+)?$ ]]; then
-            echo "valid=true" >> $GITHUB_OUTPUT
+            echo "valid=true" >>"$GITHUB_OUTPUT"
           else
-            echo "valid=false" >> $GITHUB_OUTPUT
+            echo "valid=false" >>"$GITHUB_OUTPUT"
           fi
 
+  # Each architecture builds on a runner of its own architecture and pushes
+  # untagged, digest-addressed images; `merge` then stitches them into one index
+  # per variant. Emulating arm64 under QEMU made the C/C++ stages dominate the
+  # build. The full image is `FROM` the slim one, so the two targets share
+  # almost all of their work within a single job.
   build:
     needs: validate-tag-semver
-    runs-on: ubuntu-latest
     if: needs.validate-tag-semver.outputs.valid == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-latest
+          - platform: linux/arm64
+            arch: arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
     permissions:
       id-token: write
-      contents: write
+      contents: read
       packages: write
+    env:
+      DIGEST_IMAGE: ghcr.io/${{ github.repository_owner }}/romm
+      CACHE_IMAGE: ghcr.io/${{ github.repository_owner }}/romm-buildcache
     steps:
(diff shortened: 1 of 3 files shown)


gantoine:
> and the cause was not the C libraries

thanks claude very helpful 💀

.github/workflows/build.yml
```diff
@@ -104,42 +116,164 @@ jobs:
...
-      - name: Build slim image
+      - name: Build and push slim image by digest
         id: build-slim
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           file: docker/Dockerfile
           context: .
-          push: true
-          platforms: linux/arm64,linux/amd64
-          tags: ${{ steps.meta-slim.outputs.tags }}
```
gantoine, line 115: is `push` true by default? and `tags` are no longer needed?

sdornan:
> > and the cause was not the C libraries
> 
> 
> 
> thanks claude very helpful 💀 

To be fair to Claude, that's because it was my theory going in. 😆

.github/workflows/build.yml
sdornan: No on both counts, but neither is missing.

`push` defaults to `'false'` (checked `action.yml` at the pinned SHA). We're not using that input at all here. It's shorthand for `--output=type=registry`, and this step passes `outputs:` directly instead, where `push=true` is the image exporter's own parameter. Setting both would give buildx two output specs, so it's `outputs:` or `push:`, not both.

`tags` is deliberately gone. `push-by-digest=true` pushes the manifest **untagged**, addressed only by its digest — that's the whole point of the split. Each arch pushes a nameless manifest, the digests travel to `merge` as artifacts, and `docker buildx imagetools create` applies the real tags there while stitching both arches into one index. Passing `tags` here would be pushing a tag per architecture, which is what we're trying to avoid: the last one to finish would win and clobber the other arch.

`name=` in the `outputs` string is still needed, but only to tell buildx *which repository* to push the blobs to. `name-canonical=true` makes it report back the `repo@sha256:...` reference that the next step writes out as the digest artifact.

---

gantoine (approved):

sdornan:
@gantoine heads up — your approval landed on `e12d5721`, and two commits have gone in since. Flagging rather than letting it look reviewed end to end.

**`cc9be024`** is the substantive one. Three jobs were writing the single sticky comment, which is why `merge` had an artificial `needs: prepare` edge and a separate `report-failure` job existed — and neither covered `prepare` itself failing. That's now one terminal `report` job depending on `prepare`, so the ordering comes from the graph instead of a hack. Two consequences worth a look:

- `merge` dropped its `pull-requests: write`, so no job holds registry credentials and comment access at the same time.
- `docker/setup-buildx-action` is gone from **both** merge jobs, including the release one. `imagetools create` is a pure registry call against the preinstalled CLI plugin and never touches a builder. Verified on the preview path just now (merge went 20s → 12s); the release path is inference from the same mechanism.

**`01c8bbbb`** hoists `DIGEST_IMAGE` to workflow scope. It was declared separately in `build` and `merge`, and those two have to agree or a release pushes digests to one place and tags from another.

Both came out of `/pr-ready` (security audit → code review → simplify → review-polish). The audit came back clean; the code review also closed a tag-injection hole in `validate-tag-semver` and a case where the manifest step could push an untagged index and still exit 0 — a green release nobody can pull.

Still true and unchanged by any of this: `build.yml` has never executed, and its first run will be a live release. Re-approve or re-review as you see fit.

---

PR #4456, opened 2026-09-11
fix(frontend): stop ESLint and Prettier fighting over formatting

sdornan:
**Description**

ESLint and Prettier currently fight over formatting in the frontend.

`eslint-plugin-vue`'s `flat/recommended` enables 12 rules whose `meta.type` is `layout` (pure formatting), and `eslint.config.js` switched off exactly one of them (`vue/max-attributes-per-line`). There was no `eslint-config-prettier`, so the other 11 stayed live on code that Prettier owns. On a tree that is 100% Prettier-clean, `npx eslint .` reported **813 formatting warnings**, all in `.vue` files:

| rule | warnings |
| --- | --- |
| `vue/singleline-html-element-content-newline` | 444 |
| `vue/html-indent` | 170 |
| `vue/html-self-closing` | 102 |
| `vue/html-closing-bracket-newline` | 60 |
| `vue/multiline-html-element-content-newline` | 37 |

Some of those do not just disagree, they loop. On `src/v2/components/GameDetails/OverviewTab.vue` the two tools undid each other exactly, forever:

```
eslint --fix:   }}</span          ->  }}</span><a
                ><a
prettier -w:    }}</span><a       ->  }}</span
                                      ><a
net change after both: 0 lines
```

`eslint --fix` also rewrote 40 further lines that Prettier then accepted rather than restored (moving inline content onto its own line changes rendered whitespace, which Prettier deliberately avoids). So whichever tool ran last decided the file's style.

This adds `eslint-config-prettier` as the last entry in the flat config, which switches off the 358 stylistic rules Prettier owns, including all 12 vue layout rules. Semantic and accessibility rules are untouched, and so are this repo's deliberate overrides (`vue/component-name-in-template-casing`, `vue/prop-name-casing`, `vue/attribute-hyphenation`, `vue/attributes-order`, the custom `no-unused-vars` patterns). The `vue/max-attributes-per-line: "off"` override is removed because `eslint-config-prettier` now does that.

Second, smaller fix in the same file: `src/__generated__` sat in a config object that also carried `rules` and `languageOptions`. In flat config that makes it a per-config exclusion, not a global ignore, so the base configs still linted the generated API types and produced 263 `Unused eslint-disable directive` warnings. It moves to the `ignores`-only object and gains the `/**` suffix.

Both changes are config-only. No source file is reformatted, and no rule that catches a real defect is disabled.

Note on scope: Trunk's `eslint` definition covers only the `typescript` and `javascript` file types, so `.vue` is never ESLint-checked in CI (`trunk check --filter=eslint <file>.vue` answers "Found no applicable linters for the requested path"). That is why none of the above was visible. Wiring `.vue` into Trunk's ESLint surfaces 81 pre-existing error-severity findings across 505 files, so it needs its own PR rather than riding along here.

**Checklist**

Verified on Node 24.16.0 with npm 11.13.0, matching `frontend/.nvmrc` and `packageManager`:

- `npx eslint .` over the frontend, before -> after: total messages 1220 -> 144, warnings 1136 -> 60, errors 84 -> 84 (unchanged, so nothing real was disabled). The 813 formatting warnings and the 263 unused-disable warnings are both gone; files linted drops 1218 -> 945 as the generated types are now skipped.
- Remaining 60 warnings are all non-formatting and still enforced: `vue/attributes-order` (37), `vue/require-default-prop` (12), `vue/no-required-prop-with-default` (5), `vue/no-v-html` (4), `vue/v-slot-style` (2).
- `prettier --check` over `frontend/src/**/*.{vue,ts}` with Trunk's pinned Prettier 3.9.5 plus `.trunk/configs/.prettierrc.yaml`: clean before and after.
- `trunk check --filter=eslint,prettier` on the changed files: no issues (this also confirms Trunk's hermetic ESLint 10.7.0 loads the new config).
- The `eslint --fix` then `prettier -w` sequence on `OverviewTab.vue` now changes 0 lines.
- `npm install --package-lock-only` under npm 11.13.0 reproduces the committed lockfile byte for byte, and a clean `npm ci` installs it with `engine-strict` and `min-release-age` honoured.

No unit tests: the change is lint configuration, and its behavior is asserted by the lint run itself.

**AI assistance disclosure**

Per `CONTRIBUTING.md`: this PR was written by Claude Code, which did the investigation, the config change, and this description.

https://claude.ai/code/session_01GY64DJaHPtqSf6bokeXLyS

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -63,6 +63,14 @@ lint:
     - linters: [mypy]
       paths:
         - backend/alembic/**
+    # The v1 UI is frozen and slated for deletion, and the freeze check blocks
+    # changes to these paths, so ESLint findings here cannot lead to fixes.
+    - linters: [eslint]
+      paths:
+        - frontend/src/views/**
+        - frontend/src/components/**
+        - frontend/src/console/**
+        - frontend/src/layouts/**
   files:
     - name: vue
       extensions: [vue]
@@ -71,6 +79,7 @@ lint:
       files:
         - javascript
         - typescript
+        - vue
       commands:
         - name: lint
           run_from: ${root_or_parent_with_any_config}
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -63,14 +63,15 @@ lint:
     - linters: [mypy]
       paths:
         - backend/alembic/**
-    # The v1 UI is frozen and slated for deletion, and the freeze check blocks
-    # changes to these paths, so ESLint findings here cannot lead to fixes.
+    # The v1 UI is frozen and the freeze check blocks changes to these paths,
+    # so ESLint findings in its templates cannot lead to fixes. Its TypeScript
+    # stays covered, since that was already linted before .vue was.
     - linters: [eslint]
       paths:
-        - frontend/src/views/**
-        - frontend/src/components/**
-        - frontend/src/console/**
-        - frontend/src/layouts/**
+        - frontend/src/views/**/*.vue
+        - frontend/src/components/**/*.vue
+        - frontend/src/console/**/*.vue
+        - frontend/src/layouts/**/*.vue
   files:
     - name: vue
       extensions: [vue]
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -1,4 +1,5 @@
 import eslint from "@eslint/js";
+import prettierConfig from "eslint-config-prettier/flat";
 import vue from "eslint-plugin-vue";
 import vuea11y from "eslint-plugin-vuejs-accessibility";
 import globals from "globals";
@@ -13,7 +14,7 @@ export default tseslint.config(
   // Storybook config files live outside `src/` and aren't part of the
   // app tsconfig, so type-aware linting can't resolve them.
   {
-    ignores: [".storybook/**"],
+    ignores: [".storybook/**", "src/__generated__/**"],
   },
   {
     ignores: [
@@ -30,7 +31,6 @@ export default tseslint.config(
       "dist-ssr",
       "coverage",
       "*.local",
-      "src/__generated__",
       "*.config.js",
       "src/plugins/*.d.ts",
     ],
@@ -47,7 +47,6 @@ export default tseslint.config(
     },
     rules: {
       "vue/multi-word-component-names": "off",
-      "vue/max-attributes-per-line": "off",
       "vue/valid-v-slot": "off",
       "vue/no-use-v-if-with-v-for": "off",
       "vue/component-name-in-template-casing": [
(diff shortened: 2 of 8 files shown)


sdornan:
Thanks, addressing both points from the Copilot review.

**Comment length (nit): applied** in 6860a844. Dropped the added sentence, so the block is back to its original three lines.

**`vue/attribute-hyphenation`: not affected.** `eslint-config-prettier` does not include this rule, so placing `prettierConfig` last does not disable it. Two checks on a `.vue` file with the config as committed:

`eslint --print-config src/v2/components/GameDetails/OverviewTab.vue` resolves it as still-on, alongside the other project overrides:

```
vue/attribute-hyphenation              [2, "always"]
vue/prop-name-casing                   [2, "camelCase"]
vue/component-name-in-template-casing  [2, "PascalCase", { registeredComponentsOnly: true }]
vue/html-indent                        [0]   <- correctly disabled, Prettier owns it
```

And behaviorally, adding a camelCase attribute to a component in that file still errors:

```
259:60  error  Attribute 'myProbeAttr' must be hyphenated  vue/attribute-hyphenation
```

The 358 rules `eslint-config-prettier/flat` switches off are whitespace and punctuation only. Of the 39 `vue/*` rules in that set, every one is `meta.type: "layout"`; attribute *casing* is not something Prettier rewrites, so it is not in the list. Moving `prettierConfig` earlier would instead re-enable the 11 layout rules this PR is meant to silence, so the ordering stays as is.


---

gantoine (approved):

gantoine:
this is fine, but does it even matter if `trunk fmt` handles formatting anyways?

sdornan:
Fair question. `trunk fmt` settles what the files *look like*, but it does not stop the vue layout rules from firing, because ESLint still evaluates them against Prettier's output. On a tree that is fully `trunk fmt`-clean, `npx eslint .` reported 813 warnings, all from those 11 rules.

And it is worse than noise, because `eslint --fix` actively reverses Prettier. On `src/v2/components/GameDetails/OverviewTab.vue`:

```
eslint --fix:   }}</span          ->  }}</span><a
                ><a
prettier -w:    }}</span><a       ->  }}</span
                                      ><a
net change after both: 0 lines, and it repeats forever
```

So "Prettier handles formatting" is true of the formatter and not of the linter; they simply disagree in place.

You are right that it is currently invisible, and that is the real reason to land it now. Trunk's `eslint` definition covers only the `typescript` and `javascript` file types, so `.vue` is never ESLint-checked and all 813 warnings sit where nobody looks. Two things change that:

- #4457 adds `npm run lint`, which makes those 813 the first thing a contributor sees locally.
- The `.vue` wiring I'm preparing on top of this branch puts them in front of every PR that touches a `.vue` file. I measured that one: Trunk builds its baseline from the upstream config, where ESLint does not apply to `.vue`, so findings in a newly-linted `.vue` file count as *new* rather than existing. Without this PR, switching `.vue` on would fail PRs on formatting the formatter itself produced.

Net cost here is a dependency that only sets rules to `off`, and nothing semantic changes: error-severity findings stay at 84 before and after, and the 60 remaining warnings are all non-formatting.


---

PR #4478, opened 2026-09-12
fix(streaming): prefer the emulator's own screenshot for save-state thumbnails

LoneAngelFayt:
**Description**

Two follow-up fixes to emulator streaming (#4314): save-state thumbnails could
come out black or visibly corrupted because `pull_state_to_library()` always
preferred the browser-captured canvas frame, even for emulators that write a
better frame themselves.

That canvas is a lagging, lossy copy of the video stream, so a save taken right
after a scene change (for example, pausing) can catch it holding a torn or
missing keyframe, and a capture that misses the live canvas entirely yields a
black image. The state itself is fine in both cases; only the thumbnail is bad.

The preference order in `backend/handler/streaming/states.py` is now:

- **PCSX2** uses the screenshot embedded in every `.p2s` savestate. It is already
  downloaded as part of the state content, so it costs no extra round-trip and
  can never be a stale capture.
- **RetroArch** uses the thumbnail the broker writes beside the state, taken off
  the core's real framebuffer at the moment of the save.
- **Every other emulator** is unchanged and still prefers the browser frame,
  which is what the player actually saw and never asks the emulator to read back
  its own framebuffer (the thing that deadlocks GPU-rendered cores).

Each step falls back to the next when its source is unavailable, so no emulator
loses a thumbnail it was getting before.

**Checklist**

`backend/tests/endpoints/test_streaming.py` gains coverage for the PCSX2 and
RetroArch preferences alongside the existing browser-frame test.

**AI assistance**

This PR was written primarily by Claude Code: the code, tests and this
description were AI-generated, then reviewed and tested by me. Any follow-up
responses on this PR are AI-assisted as well.

### DIFF
--- a/backend/handler/streaming/states.py
+++ b/backend/handler/streaming/states.py
@@ -208,9 +208,8 @@ def extract_state_screenshot(emulator: str, state_content: bytes) -> bytes | Non
             with zf.open(_SCREENSHOT_ZIP_ENTRY) as entry:
                 data = entry.read(SCREENSHOT_MAX_BYTES + 1)
     except Exception as exc:
-        # No screenshot entry, or the state is not a readable zip. Never fatal,
-        # a corrupt entry (zlib.error, a truncated member) must cost the
-        # thumbnail only: the state itself still syncs.
+        # Never fatal: a missing entry, an unreadable zip or a corrupt member
+        # costs the thumbnail only, the state itself still syncs.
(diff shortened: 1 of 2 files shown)


gantoine (approved):

LoneAngelFayt:
@gantoine 
sorry for adding this after approval, but I had to fix a bunch of stuff this addresses, since now i am able to grab the browser's frame correctly

PR #4480, opened 2026-09-12
fix(tasks): fill the scan counters a job's stored stats predate

sdornan:
**Description**

Upgrading from 5.2.x with any scan in the task history breaks the Tasks view entirely:

```
{'type': 'missing', 'loc': ('response', 15, 'ScanTaskStatusResponse', 'meta', 'scan_stats', 'updated_roms'), 'msg': 'Field required',
 'input': {'total_platforms': 1, 'total_roms': 819, 'scanned_platforms': 1, 'new_platforms': 1, 'identified_platforms': 1,
           'scanned_roms': 378, 'new_roms': 378, 'identified_roms': 378, 'scanned_firmware': 0, 'new_firmware': 0}}
{'type': 'missing', 'loc': ('response', 15, 'ScanTaskStatusResponse', 'meta', 'scan_stats', 'new_files'), 'msg': 'Field required', ...}
```

That `input` is 5.2.0's `ScanStats`, exactly: ten keys, ending at `new_firmware`. 5.3.0 added `updated_roms` and `new_files` to the TypedDict as required fields, but RQ keeps each job's meta in Redis, which survives the upgrade in the `romm_redis_data` volume. `_build_task_status_response` passes that stored dict straight to the response model:

```python
meta={"scan_stats": job_meta.get("scan_stats")},
```

so one pre-upgrade scan fails validation. And since `get_tasks_status` returns a list, the failure takes out the whole response rather than the row it belongs to: the Tasks page shows nothing at all, for as long as that job stays in the finished/failed registry.

The other stats types (`ConversionStats`, `UpdateStats`, `CleanupStats`) did not change and are not affected today, but the hazard is the same shape for any of them.

**Approach.** Fill the missing counters when reading, rather than making the fields `NotRequired`. Optional fields would propagate into `frontend/src/__generated__/` and force every consumer to handle `undefined` for a value that is always present on new scans; filling keeps the API contract and the generated types exactly as they are, and needs no `npm run generate`. Zero is the honest value: a run on a release that never counted `new_files` did not find any it recorded.

The baseline is derived from `ScanStats.__annotations__` rather than hand-listed, so the next counter added to the TypedDict is covered without a second edit. That second edit is precisely what the release that added these two fields missed.

Stats a job has not reported yet stay `None`, which the UI distinguishes from a scan that counted zero.

For anyone hitting this before the fix ships, the workaround is to drop RQ's keys from Redis (`valkey-cli --scan --pattern 'rq:*' | xargs -r -n 100 valkey-cli DEL`), which costs the task history but leaves sessions signed in.

AI assistance disclosure, per `CONTRIBUTING.md`: this change was written with Claude Code (diagnosis, patch, tests and this description), reviewed before submission.

**Checklist**

Two tests in `backend/tests/endpoints/test_tasks.py`:

- `test_a_scan_predating_a_counter_reports_it_as_zero` feeds the endpoint the exact ten-key payload above and asserts a 200 with `updated_roms`/`new_files` zeroed and the counters that were stored preserved.
- `test_a_scan_that_never_reported_stats_keeps_none` pins the null case, so the fill never invents a zeroed stats block for a queued scan.

Not verified locally: no MariaDB/PostgreSQL or Python 3.14 in the environment this was written in, so the suite runs in CI. `black`, `ruff` and `isort` are clean, and the fill was exercised against the payload from the report.

n/a

https://claude.ai/code/session_01WRP83vJP7Cws6SSdE3XZ1a

---
_Generated by [Claude Code](https://claude.ai/code/session_01WRP83vJP7Cws6SSdE3XZ1a)_

### DIFF
--- a/backend/endpoints/responses/__init__.py
+++ b/backend/endpoints/responses/__init__.py
@@ -1,4 +1,4 @@
-from typing import Literal, TypedDict, Union
+from typing import Any, Final, Literal, Mapping, TypedDict, Union, cast
 
 from rq.job import JobStatus
 
@@ -18,6 +18,27 @@ class ScanStats(TypedDict):
     new_firmware: int
     updated_roms: int
     new_files: int
+
+
+# Read off the annotations so a counter added here is filled without a second
+# edit, which is the step the release that added `updated_roms` and `new_files`
+# missed.
+EMPTY_SCAN_STATS: Final[ScanStats] = cast(
+    ScanStats, dict.fromkeys(ScanStats.__annotations__, 0)
+)
+
+
+def fill_scan_stats(stats: Mapping[str, Any] | None) -> ScanStats | None:
+    """A scan's counters, with any the release that wrote them predates zeroed.
+
+    A job's meta outlives the release that stored it in Redis, so an upgrade
+    leaves stats behind that are missing every counter added since. Zero is what
+    a run that never counted one reported.
+    """
+    if stats is None:
+        return None
+
+    return cast(ScanStats, {**EMPTY_SCAN_STATS, **stats})
 
 
 class ScanTaskMeta(TypedDict):
--- a/backend/endpoints/responses/__init__.py
+++ b/backend/endpoints/responses/__init__.py
@@ -1,4 +1,4 @@
-from typing import Any, Final, Literal, Mapping, TypedDict, Union, cast
+from typing import Literal, TypedDict, Union
 
 from rq.job import JobStatus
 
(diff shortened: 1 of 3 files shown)


backend/endpoints/responses/__init__.py
```diff
@@ -20,6 +20,23 @@ class ScanStats(TypedDict):
...
+
+def fill_scan_stats(stats: Mapping[str, Any] | None) -> ScanStats | None:
+    """A scan's counters, zeroing any the release that stored them predates.
+
+    A job's meta in Redis outlives the release that wrote it.
+    """
+    if stats is None:
+        return None
+
+    return cast(ScanStats, {**EMPTY_SCAN_STATS, **stats})
```
gantoine: do we usually add functions to endpoint response files?

gantoine (changes requested):

backend/endpoints/responses/__init__.py
sdornan: Not in this one. A few of the response modules do carry helpers (`feeds.coerce_to_int`, `rom.rom_user_schema_factory`, `permission.action_keys_for`), but `responses/__init__.py` is pure declarations and I shouldn't have been the one to change that.

Moved to `endpoints/tasks.py` as `_fill_scan_stats`, right above `_build_task_status_response`, its only caller. `responses/__init__.py` is now untouched by this PR.

38a43ce2

---

gantoine (approved):

Issue #3932, opened 2026-07-24
[Bug] Migration 0100 (roms_facets triggers) fails on MariaDB/MySQL with binary logging enabled and no SUPER privilege


Spinnich:
**RomM version**
5.1.0-alpha.3 and 5.1.0-alpha.4 (the trigger-based `roms_facets` sync was introduced in alpha.3 by migration `0100_roms_facets_table`).

**Describe the bug**
Migration `0100_roms_facets_table` creates `AFTER INSERT/UPDATE` triggers on the `roms` table to keep the new `roms_facets` filter table in sync. On any MariaDB/MySQL server that has **binary logging enabled** while the RomM database user **lacks the `SUPER` privilege** and `log_bin_trust_function_creators` is off (error 1419), the server refuses all trigger DDL, so the migration aborts and the container fails to start:

```
mariadb.OperationalError: You do not have the SUPER privilege and binary logging is
enabled (you *might* want to use the less safe log_bin_trust_function_creators variable)
[SQL: DROP TRIGGER IF EXISTS roms_facets_after_insert]
ERROR:    [RomM][init] Failed to run database migrations
```

This combination is common on hardened/managed database instances and is the default on MySQL 8 (binary logging is on out of the box). RomM currently documents no `SUPER` / `log_bin_trust_function_creators` prerequisite, and the migration does not guard against or surface this condition, so affected users get a hard startup failure with only a raw SQLAlchemy traceback.

Note the error fires even on `DROP TRIGGER IF EXISTS` (added in alpha.4), so the alpha.4 idempotency changes do not help these users. Those changes made the migration safely re-runnable after a partial failure (good for anyone half-migrated by alpha.3), but they do not address the underlying privilege requirement, which is the actual blocker.

**To Reproduce**
1. Run RomM against a MariaDB/MySQL server that has binary logging enabled (`log_bin = ON`).
2. Ensure the RomM DB user does **not** have `SUPER` (a normal least-privilege app user) and that `log_bin_trust_function_creators` is off (the default).
3. Start RomM 5.1.0-alpha.3 or later so migration `0100_roms_facets_table` runs.
4. Startup aborts with the error 1419 traceback above; the container never comes up.

**Expected behavior**
The upgrade should either:
- succeed without requiring `SUPER` on the RomM DB user (e.g. an application-level sync hook instead of trigger DDL), and/or
- detect the missing privilege and fail with a clear, actionable message naming `log_bin_trust_function_creators=1` as the fix, and
[...]

PR #4484, opened 2026-09-12
fix(migrations): let every 5.3.0 revision be replayed

gantoine:
**Description**

Upgrading to 5.3.0 can fail on every start with an error that says nothing about what actually went wrong:

```
mariadb.OperationalError: Duplicate column name 'generated_primary_region'
[SQL: ALTER TABLE roms ADD COLUMN generated_primary_region VARCHAR(50) GENERATED ALWAYS AS (...) STORED]
ERROR: [RomM][init] Failed to run database migrations
```

MySQL and MariaDB auto-commit each DDL statement, so a revision that dies partway keeps what it created while `alembic_version` stays behind. Every start after that replays the revision from the top and dies on the first object it already made, hiding the real failure. The init script logs it and starts the app anyway, so the instance keeps running on a half-migrated schema.

Eleven revisions create something ahead of a step that can fail. Most just take the guard alembic already offers (`if_not_exists=True`, matching what `0113`-`0116` have always done); three needed more:

| Revision | Guarded |
|---|---|
| `0098` | the generated-column `ALTER`, plus the `create_index` loop after it |
| `0108` | the generated column (its index rebuild was already idempotent) |
| `0111` | both column adds |
| `0112` | the two generated columns on `roms`, the two on `roms_facets` |
| `0119` | both tables and their four indexes |
| `0120` | the table |
| `0121` | the column, its index, and its foreign key |
| `0122` | the table, and its inline indexes |
| `0123` | the batched generated-column `ALTER`, and the three `roms_facets` adds |
| `0126` | each of its four steps independently |
| `0128` | the generated column |

**The three that needed real logic**

- **`0122`** declares its indexes inline so InnoDB adopts one for the foreign key instead of silently adding a second on the same column. Alembic issues those as their own statements, so `if_not_exists=True` on the table does *not* cover them, and a replay dies on `Duplicate key name 'idx_rom_similarity_related_rom_id'`. It now reflects the table and fills in whichever index is missing, keeping the inline form on the fresh path.
- **`0123`** swaps every generated column it touches in one `ALTER TABLE`, dropping the rebuilt ones and adding the new ones. On a replay the adds fire against columns that are already there, so it now drops whatever is present and re-adds all of them at this revision's definition.
- **`0126`** reflects `roms` first and guards each step: add the column only if absent, backfill only the rows still missing a digest, apply `NOT NULL` only while nullable, rebuild `idx_roms_platform_id_fs_name` only while still unique. A replay resumes where the previous run stopped, so if the original failure was a real one it surfaces its own error instead of the duplicate-column one.

`full_path_digest_sql` also coalesces a NULL `fs_path`/`fs_name`, matching `compute_full_path_hash`. Both columns are declared NOT NULL, but where the two spellings disagreed the SQL folded the whole row to NULL and the backfill aborted `0126` at its `NOT NULL` step, which is one way a database ends up in this state.

**On the helper.** `has_column` is new in `utils/database.py` because SQLAlchemy's `Inspector` offers `has_table`, `has_index`, `has_schema` and `has_sequence` but no column equivalent, and alembic's `Operations` has none either. Where the library does provide one, this uses it. `add_column(if_not_exists=True)` covers most of the ground but not the raw `ALTER TABLE ... GENERATED ALWAYS AS (...) STORED` that `0098`/`0108`/`0112` need.

Every revision id is unchanged: these all shipped in an alpha, so an install that got through them sees no-ops, and an install stuck partway completes.

**Supersedes #4479 and #4481**, whose fixes are carried here in full; both are closed in favour of this. Original work by @sdornan, credited on the commit.

**Not addressed here:** the privilege prerequisite behind #3932. Nothing documents that trigger-creating revisions need `log_bin_trust_function_creators=1` or `BINLOG ADMIN` on a binlog-enabled server. That's a docs change or a pre-flight check, and a separate call.

AI assistance disclosure, per `CONTRIBUTING.md`: written with Claude Code (diagnosis, patch, tests and this description), reviewed before submission.

**Checklist**

23 tests in `backend/tests/test_migrations.py`, green on MariaDB 12.3.3 and PostgreSQL 18, each run against a database built from `0001` to head:

- `test_a_revision_replayed_over_the_migrated_schema_is_a_no_op`, parametrized over `0111`, `0120`, `0122`, `0123`, `0126` and `0128` — asserts the reflected columns *and* indexes are identical before and after the replay. Each case was confirmed to fail against the current `master` version of its revision.
- `test_the_memory_card_revision_replays_under_the_one_that_prunes_it` — `0119` re-creates the index `0125` drops, so the pair replays together, as a real upgrade runs them.
- `test_the_full_path_hash_migration_resumes_an_interrupted_run` and `test_the_hltb_migration_resumes_an_interrupted_run` build the real intermediate states and assert the replay completes them.
- `test_the_state_disc_file_migration_resumes_an_interrupted_run`, parametrized over dying before the index and before the column.
- `test_has_column_reflects_the_migrated_schema` and `test_the_publisher_split_column_add_replays` pin the shared guard and `0112`'s column add.

**A deliberate limit worth reviewing:** `0098`, `0108` and `0112` are not replayed whole. Each rebuilds a shared object — `roms_metadata`, the facets triggers, `idx_roms_sibling_cover` — at its own era's definition, and later revisions have since replaced those, so replaying one against a head schema would regress the object. Those tests exercise the guarded statements instead.

n/a


sdornan:
Correction on one line of this description, which you inherited from my #4479/#4481 and which is wrong — mine, not yours.

> The init script logs it and starts the app anyway, so the instance keeps running on a half-migrated schema.

It doesn't. `error_log()` in `docker/init_scripts/init` ends with `exit 1`:

```bash
error_log() {
	echo -e "${RED}ERROR:    ..." || true
	exit 1
}
```

so `error_log "Failed to run database migrations"` has terminated the script since the file was written. A failed upgrade stops the container; a restart policy then runs it again. That restart loop is what the original report in this line of work actually looked like, not a degraded app.

I found it by writing a patch to "fix" the non-existent problem and having `shellcheck` tell me my new lines were unreachable (SC2317), which is a fair trade. Nothing in the code here is affected — the guards stand on their own — but the sentence is worth dropping so it doesn't get quoted onward.

Everything else: the diff is strictly better than what it replaced, `0122`'s inline indexes and `0123`'s `_rebuild_columns` are both things I missed, and confirming each replay case fails against master's version of its revision is a verification step I should have done and didn't.

---

Issue #4483, opened 2026-09-12
Reconcile derived schema objects declaratively instead of redefining them per-revision


sdornan:
## Problem

`alembic_version` is a pointer into a history, not a description of the schema. MySQL and MariaDB auto-commit each DDL statement, so a revision that dies partway leaves the schema and the version counter disagreeing — and the next start replays from the top and fails on an error that describes the *replay*, not the original cause.

The shape it takes in practice: a revision adds columns, then creates triggers. Trigger DDL is refused outright on a binlog-enabled server whose app user lacks `SUPER`/`BINLOG ADMIN` with `log_bin_trust_function_creators` off (#3932, error 1419). The user applies the documented remedy, restarts, and is met by `Duplicate column name 'generated_publishers'` — with nothing to suggest their privilege fix worked.

Three PRs are in flight fixing instances of this by hand: #4479 (`0126`, `0128`), #4481 (`0098`, `0108`, `0111`, `0112`, `0121`, `0123`).

## The churn this creates

Derived objects are re-declared in full, once per revision that touches them:

| Object | Revisions that define it |
|---|---|
| `roms_metadata` view | 11 |
| `sibling_roms` view | 7 |
| `virtual_collections` view | 4 |

Four revisions carry hand-written `_rebuild_*` helpers (`0108`, `0112`, `0123`, `0127`). Every one is `DROP` + `CREATE` to the current definition — declarative reconciliation, written by hand, repeatedly. The guards added in the PRs above are the same admission in another form: reflection checks bolted onto a counter-driven framework.

It also makes each revision's blast radius wider than it looks. Because `0098`, `0108`, `0112` and `0123` each rebuild a shared object at their own era's definition, none of them can be replayed in a test against a current schema — replaying `0108` today would recreate `idx_roms_sibling_cover` without the `steam_id` that `0127` added. That is why the PRs above can only test their guarded statements, not the revisions end to end.

## Proposal

Split the schema by whether history matters.

**Data-bearing — tables, columns, foreign keys, backfills.** Keep sequential migrations. Ordering and one-way transformations are real, and no diff can tell a rename from a drop-plus-add or know how to hash every row's path into a new column.

**Derived — views, triggers, functions, dialect-specific indexes.** Declare once, reconcile against reflection at startup. A change edits one definition instead of adding a revision; a half-applied rebuild self-heals on the next boot; and the objects stop being landmines for anyone testing a migration.

## Costs, stated up front

[...]

PR #4485, opened 2026-09-12
spike(models): declare STORED generated columns on the model instead of in raw DDL

sdornan:
**Description**

A spike for #4483, not a merge candidate. It answers one question: can the mapping own the generated columns that `env.py` currently hides from autogenerate, and what does that cost?

`alembic/env.py` says those columns are *"maintained in raw SQL (dialect-specific expressions) rather than the ORM model"*. The mapping can in fact express them, since `Computed(expr, persisted=True)` renders `GENERATED ALWAYS AS (...) STORED` on both dialects. What it cannot express is one declaration whose SQL differs per dialect, and that is solved by dispatching the expression rather than the column:

```python
class PrimaryRegion(ColumnElement):
    inherit_cache = True

@compiles(PrimaryRegion, "mysql")
@compiles(PrimaryRegion, "mariadb")
def _primary_region_maria(element, compiler, **kw) -> str: ...

@compiles(PrimaryRegion, "postgresql")
def _primary_region_postgres(element, compiler, **kw) -> str: ...
```

which reduces the model declaration to:

```python
 generated_primary_region: Mapped[str | None] = mapped_column(
-    String(length=50),
-    server_default=FetchedValue(),
-    server_onupdate=FetchedValue(),
+    String(length=PRIMARY_REGION_LENGTH),
+    Computed(PrimaryRegion(), persisted=True),
 )
```

**Correction: this does not inherit the existing drift test**

An earlier revision of this description claimed that narrowing `include_object` put these two columns under `test_no_index_drift_between_models_and_migrations`, and used that to justify adding no tests. That was wrong, on two independent counts:

- the test builds its own `MigrationContext.configure(connection)` with no `include_object`, so the filter in `env.py` never applies to it. `include_object` is referenced only by the two `context.configure` calls in `env.py`;
- the test filters every diff to `INDEX_DIFF_OPS = {"add_index", "remove_index"}`, so a column-level diff never reaches its assertion.

The `include_object` change therefore only affects a developer running `alembic revision --autogenerate` by hand. Widening that test to column diffs is not a one-line follow-up either: `compare_metadata` against the live schema currently yields a `remove_column` for each of the twelve generated columns that exist on `roms` but are absent from `BaseModel.metadata`, plus table-level noise from the three view-backed models, and the index-only filter is exactly what suppresses that. A column-level drift test only becomes reachable once those twelve are in the metadata, which is the deferred-mapping step described under scope below.

**The checks this adds instead**

- `test_the_model_expression_matches_the_revision_that_created_it` compares the model's rendered expression against the literal its revision emits, per dialect. This is the guard for the duplication the next section defends: two copies of the SQL with nothing holding them together was the real gap.
- `test_the_stored_generated_column_matches_the_model_expression` compares by value, selecting the model's expression over the same rows as the stored column. This catches what the string comparison cannot, namely a later revision rebuilding the column, which is how `generated_first_release_date` and `generated_average_rating` have actually changed.
- `test_every_model_declared_generated_column_is_pinned_to_a_revision` fails if a third `Computed` column is declared without pinning it, so the guard extends itself.

`include_object` also stops carrying a hand-maintained `MODEL_OWNED_GENERATED_COLUMNS` set of column-name strings and derives the answer from `Column.computed` instead. Worth noting for review: SQLAlchemy reflection populates `computed` on the *database* side too, so a naive `object.computed is not None` would have proposed dropping all twelve unmapped generated columns. The check keeps `not reflected` for that reason.

**Scope, and why it stops where it does**

Two of the five generated columns mapped on `Rom` are declared here. The other three are not, and the reason is the useful part of this spike. Expression length at head:

| Column | MariaDB | PostgreSQL | |
|---|---|---|---|
| `generated_primary_region` | 53 | 23 | declared here |
| `generated_hltb_main_story` | 379 | 251 | declared here |
| `generated_first_release_date` | 5,302 | 2,279 | not declared |
| `generated_average_rating` | 7,874 | 5,360 | not declared |
| `generated_player_count` | - | - | assembled across five revisions |

The first two are literals. The others are *compositions*: `0123` builds them from provider source lists and per-provider helpers, and `generated_player_count`'s chain grows in `0098`, `0114`, `0115` and `0123`. Declaring those on the model means relocating the builders into the model layer, not copying strings; pasting 13KB of generated SQL into `models/` would be worse than what is there today. That relocation is a real refactor and belongs in its own change, if it is wanted at all.

Sequencing note for #4483: those three are exactly the columns whose expressions keep changing, and PostgreSQL cannot `ALTER` a generated expression, so each change is a drop and re-add. Moving them onto the model *without* the value-comparison test above would make silent divergence more likely than leaving them in the migrations, because editing `models/generated.py` looks like an ordinary code change and prompts nothing.

**What this does not buy**

Alembic detects a generated column being added or dropped. It does **not** detect a *changed* `Computed` expression. Declaring these closes the add/drop gap for a hand-run autogenerate and leaves expression drift to the two tests above rather than to Alembic.

**Costs worth arguing about**

- **The SQL exists twice by design.** `0108` and `0128` keep their literal copies. A revision that imported its DDL from the model would rewrite itself the day the expression changed, and `0127` warns about exactly this. It reads like a "one home per value" violation and isn't, which is why `models/generated.py` says so in its docstring. The revision-pinning test is what keeps the two copies honest.
- **`Computed` is stricter than `FetchedValue`.** SQLAlchemy omits the column from INSERT/UPDATE and rejects assignment. Nothing in `handler/`, `endpoints/`, `tasks/` or `models/` assigns to either column (every use is a sort key or filter), so this is safe here and arguably more correct, but it is a behaviour change, not an annotation.
- **The JSON generated columns are deliberately unmapped.** `generated_genres`, `generated_companies`, `generated_publishers` and friends are not on `Rom` at all, and are reached through the `roms_metadata` view as `RomMetadata`. Mapping them normally would add ten JSON blobs to every `roms` SELECT. `mapped_column(..., deferred=True)` would sidestep that (the column stays in `Rom.__table__` for autogenerate but is excluded from the default SELECT), which is the cheapest route to widening coverage, but it is out of scope here.

**Verification**

- Both declarations render byte-identical to the literals `0108` and `0128` emit, on the MariaDB and PostgreSQL compilers. This is now asserted by a test rather than checked by hand.
- The `include_object` derivation was checked against Alembic 1.18.4 and PostgreSQL 16 with a synthetic table covering all three cases (column in both, in the database only, in the model only), which is where the reflection behaviour noted above came from.
- The value-comparison test's mechanics were exercised standalone against PostgreSQL 16 using the DDL `0108` and `0128` actually emit: no mismatches across rows with values, empty arrays and NULLs, and a deliberately perturbed model expression is caught.
- **Not run here:** the pytest suite itself. This container only has a Python 3.14 release candidate, on which pydantic 2.13.5 fails to import, so `conftest.py` cannot load and the database-backed test is unexecuted under pytest. CI is the check on that one.
- `master` has been merged into the branch so CI runs against current code.

AI assistance disclosure, per `CONTRIBUTING.md`: written with Claude Code (design, patch and this description), reviewed before submission.

**Checklist**

Refs #4483.

n/a

https://claude.ai/code/session_01FERNCPTDFcknRQ251aST2W


sdornan:
Correcting this spike's own conclusion, since the description was being read as evidence for #4483 and the key claim in it was false.

**The claim that didn't hold.** The description said that narrowing `include_object` put these two columns under `test_no_index_drift_between_models_and_migrations`, and used that to justify adding no tests. It doesn't, for two independent reasons:

1. That test builds its own `MigrationContext.configure(connection)` with no `include_object`. The filter is referenced only by the two `context.configure` calls in `env.py`, so narrowing it has no effect on the test at all.
2. The test filters every diff to `INDEX_DIFF_OPS = {"add_index", "remove_index"}`, so a column-level diff never reaches the assertion.

Net effect: the original diff added zero test coverage, and the `include_object` change only affects a hand-run `alembic revision --autogenerate`. Worth being explicit about, because the conclusion "model-owned columns come with drift detection attached" is what #4483 would have been decided on, and it isn't true. The gap was never being tested, so nothing that was open got closed.

It also isn't a one-line follow-up to widen that test. `compare_metadata` against the live schema currently yields a `remove_column` for each of the twelve generated columns that exist on `roms` but are absent from `BaseModel.metadata`, plus table-level noise from the three view-backed models. The index-only filter is exactly what suppresses that, so a column-level drift test only becomes reachable once those twelve are in the metadata (`deferred=True` keeps them out of every `roms` SELECT while still exposing them to autogenerate).

**What I pushed.**

- Two real guards. One compares the model's rendered expression to the literal its revision emits, per dialect, which is what holds the two deliberate copies of the SQL together. The other compares by value, selecting the model's expression over the same rows as the stored column, which catches a later revision rebuilding the column. That second case is not hypothetical: it is how `generated_first_release_date` and `generated_average_rating` have actually changed.
- A third test that fails if a `Computed` column is declared without pinning its revision, so the guard extends itself.
- `include_object` now derives the answer from `Column.computed` instead of a hand-maintained set of column-name strings. One thing worth flagging for review: reflection populates `computed` on the database side too, so a naive `object.computed is not None` would have proposed dropping all twelve unmapped generated columns. The check keeps `not reflected` for that reason, which I confirmed against Alembic 1.18.4 and PostgreSQL 16 rather than inferring it.
- `master` merged in so CI runs against current code.

**Verification, and its limit.** The rendering equality and the `include_object` semantics are verified. The value-comparison test's mechanics were exercised standalone against PostgreSQL 16 using the DDL `0108` and `0128` actually emit: no mismatches across rows with values, empty arrays and NULLs, and a deliberately perturbed expression is caught. What I could not do is run pytest: this container only has a Python 3.14 release candidate, on which pydantic 2.13.5 fails to import, so `conftest.py` never loads. The database-backed test is therefore unexecuted under pytest and CI is the check on it.

**On the underlying question for #4483.** The `Computed` plus dialect-dispatch pattern works and the four renderings match byte-for-byte, so the mechanism is sound. The thing to weigh is sequencing: `generated_first_release_date`, `generated_average_rating` and `generated_player_count` are the columns whose expressions keep changing, and PostgreSQL cannot `ALTER` a generated expression, so each change is a drop and re-add. Moving those onto the model before a drift check exists would make silent divergence *more* likely than leaving them in the migrations, because editing `models/generated.py` looks like an ordinary code change and prompts nothing. Drift check first, then the big columns.

---

sdornan:
Closing along with #4483, which this was a spike for. It was never a merge candidate, and the question it set out to answer has been answered.

The finding worth keeping is recorded in the issue thread: `Computed(expr, persisted=True)` can express these columns, dispatching the expression through `@compiles` handles the per-dialect part, and alembic detects a generated column being added or dropped but not a changed expression. That last point is what mattered, and #4501 has since closed it from the other side: `ensure_roms_columns()` compares the reflected `computed.sqltext` against the desired expression and rebuilds the column when it has drifted, so the expressions have a single home and a drift check without moving them into `models/`.

That also settles the part this spike deliberately left open. The three long expressions are compositions built from provider source lists, and relocating those builders into the model layer was the cost that stopped it at two columns. `utils/roms_columns.py` is now that home, on the migration side, which is where the revisions that emit them already live.

---

PR #4486, opened 2026-09-12
fix(migrations): hold an advisory lock while upgrading

sdornan:
**Description**

Nothing serialises `alembic upgrade head`. Two processes against one database both run the chain: an HA pair, a `docker compose up` overlapping a draining container, a restart while a slow backfill is still going. MySQL and MariaDB commit each DDL statement as it lands, so the loser of that race meets objects the winner already created — the half-applied state that strands an upgrade, and one that *no* amount of idempotency guarding inside the revisions can prevent. It is the failure mode #4484 cannot reach.

`run_migrations_online` now holds an advisory lock for the duration, on the same connection the migrations run on since these locks are session-scoped: `GET_LOCK` on MySQL/MariaDB, `pg_advisory_lock` on PostgreSQL. Both entry points inherit it — the init script's `alembic upgrade head` and `main.py`'s `alembic.config.main`.

A process that cannot acquire it within `MIGRATION_LOCK_TIMEOUT` says so and stops rather than racing. The default is 900s because a digest backfill over a large library is legitimately slow, and waiting is the correct behaviour there.

**Correction, and the reason this PR shrank.** It originally carried a second change: making the init script refuse to start after a failed upgrade. That was based on a claim I got wrong, and which I repeated in #4479, #4481 and #4483 — that a failed migration leaves the app running on a half-migrated schema. It does not. `error_log()` in `docker/init_scripts/init` ends with `exit 1`, so:

```bash
else
	error_log "Failed to run database migrations"
fi
```

has terminated the script since the file was written. The container stops; a restart policy then runs it again, which is the loop this whole line of work started from. My patch added a second message that never printed and an exit that never ran — `shellcheck`'s SC2317 caught it, flagging the second `error_log` and the `exit` as unreachable while leaving the first alone, which is exactly the shape of that mistake. Reverted in e814bb3d; the init script is untouched.

AI assistance disclosure, per `CONTRIBUTING.md`: written with Claude Code (diagnosis, patch, test and this description), reviewed before submission.

**Checklist**

`test_the_migration_lock_shuts_out_a_second_connection` in `backend/tests/utils/test_database.py` holds the lock on one connection, asserts a second connection times out, and asserts it succeeds once the block releases. It runs against both engines in CI, which is what the dialect split needs.

Not verified locally: no MariaDB/PostgreSQL or Python 3.14 in the environment this was written in. `black`, `ruff`, `isort` and `shellcheck` are clean.

**Two things a reviewer should weigh:**

- The `MIGRATION_LOCK_TIMEOUT` default. 900s is calibrated to "a backfill over a big library", not measured; someone with real numbers should sanity-check it.
- `utils/database.py` is also touched by #4484 (which adds `has_column` just above this helper). Whichever lands second needs a trivial merge; no logical conflict.

n/a

https://claude.ai/code/session_01WRP83vJP7Cws6SSdE3XZ1a

### DIFF
--- a/backend/alembic/env.py
+++ b/backend/alembic/env.py
@@ -4,13 +4,14 @@ from pathlib import Path
 from alembic import context
 from sqlalchemy import create_engine
 
+from config import MIGRATION_LOCK_TIMEOUT
 from config.config_manager import ConfigManager
 from logger.logger import unify_logger
 from models import load_all_models
 from models.base import BaseModel
 from models.collection import VirtualCollection
 from models.rom import RomMetadata, SiblingRom
-from utils.database import AUTOGENERATE_EXEMPT_INDEX_NAMES
+from utils.database import AUTOGENERATE_EXEMPT_INDEX_NAMES, migration_lock
 
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
@@ -99,8 +100,12 @@ def run_migrations_online() -> None:
             include_object=include_object,
         )
 
-        with context.begin_transaction():
-            context.run_migrations()
+        # Two processes upgrading at once is how a revision ends up half
+        # applied: MySQL/MariaDB commit each DDL statement as it lands, so the
+        # loser of a race meets objects the winner has already created.
+        with migration_lock(connection, MIGRATION_LOCK_TIMEOUT):
+            with context.begin_transaction():
+                context.run_migrations()
 
 
 if context.is_offline_mode():
--- a/backend/config/__init__.py
+++ b/backend/config/__init__.py
@@ -52,6 +52,8 @@ ROM_UPLOAD_ASSEMBLING_EXT: Final[str] = "assembling"
 
 # SEVEN ZIP
 SEVEN_ZIP_TIMEOUT: Final[int] = safe_int(_get_env("SEVEN_ZIP_TIMEOUT"), 60)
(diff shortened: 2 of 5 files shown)


sdornan:
Closing this. It guards a topology RomM doesn't ship.

Only one path runs migrations in a deployment: the init script, once, before gunicorn starts. `main.py`'s call is dev-only by its own comment, and gunicorn workers never migrate. Nothing in the repo references replicas, k8s or helm. So concurrent upgrades need two stacks pointed at one database, or a dev `main.py` racing a container — real, but neither supported nor what any report here has shown.

Against that, the change isn't free. `GET_LOCK` is connection-scoped, so behind a load-balancing proxy the lock and the migration can land on different backends, meaning the mechanism quietly doesn't work for some users while still sitting in the startup path. And `MIGRATION_LOCK_TIMEOUT`'s default was a guess; too low turns a slow-but-healthy upgrade into a failed start.

Standard practice elsewhere (Flyway and Liquibase both lock), but adding it to the most safety-critical path in the app to defend an unsupported topology is not a good trade today. If migrations ever run from more than one process, this becomes necessary — the reasoning and the sketch are on #4483 so it isn't lost, and the branch stays pushed.

The original second half of this PR was reverted earlier for a worse reason: it was built on my incorrect claim that a failed upgrade leaves the app running, when `error_log` has always exited. That correction is on #4483 and #4484.

---

PR #4493, opened 2026-09-13
feat(desktop): optional Electron shell that launches games in a local emulator

sdornan:
**Description**

Adds an **optional** desktop shell under `electron/` that loads an existing RomM server in a native window and can hand a game to a **locally installed emulator** instead of an in-browser core. That last capability is the whole point of the PR: it is the one thing the existing PWA can never do.

Nothing here is required to use RomM. The web app behaves exactly as before when it is not running inside the shell, and `electron/` is a separate workspace that does not touch the existing build, Docker image, or CI.

**How a launch works**

The renderer never supplies an executable or arguments. It names a game and the libretro cores its platform supports, then the main process resolves that against the user's own emulator config:

1. The v2 UI asks the shell whether the platform is launchable (one probe per platform, cached).
2. On click, the shell downloads the ROM using the window's own session cookies and caches it (LRU, 20 GB default).
3. It resolves an emulator and spawns it.

The platform-to-core mapping is **not duplicated**: the shell is handed the candidate cores from the existing `_EJS_CORES_MAP` via `getSupportedEJSCores()` in `frontend/src/utils`, since those are already libretro core names. RetroArch is autodetected; a per-platform table covers standalone emulators (Dolphin, PCSX2, DuckStation) with a `*` wildcard row as fallback.

**Security posture**

The window loads a remote origin and renders artwork and descriptions from ten metadata providers, so the renderer is treated as untrusted:

- `contextIsolation`, `sandbox`, and `nodeIntegration: false` enforced; webview tags blocked.
- In-window navigation pinned to the configured server origin; everything else goes to the user's real browser.
- Core names must match `[a-z0-9_]+` before becoming a filesystem path, so a compromised renderer cannot point the loader outside the cores directory.
- Download URLs must resolve to the bound origin **and** an `/api/` route; protocol-relative, absolute, and traversal paths are rejected rather than normalised.
- Processes spawned with an argv array, never a shell string.
- Self-signed certificates (common on a LAN) prompt once and are remembered by fingerprint.

**Scope**

Saves and states stay wherever the local emulator writes them. Syncing them back to RomM is deliberately out of scope here; it needs per-emulator save-path knowledge and conflict handling, and is much better as its own PR once launching is proven.

Frontend changes are deliberately small: a `useNativeShell` composable, a `"native"` play target in `useGameActions`, and one action entry in the ribbon and the more-menu.

**Open question for maintainers**: the other official apps (Playnite, Android, CFW) live in separate repos. This is in-tree because it shares the platform/core map, but it is easy to split out if you would rather keep that convention.

**Checklist**

Verification run:

- `npm run typecheck`, `npm run test` (1246 passing), `npm run build` in `frontend/`
- `tsc --noEmit`, `node --test` (21 passing), ESLint in `electron/`
- `check_i18n_locales.py` and `check_i18n_sorted.py` pass; all 9 new keys added to all 18 locales
- Prettier clean across the diff

**End-to-end smoke test:** booted the packaged main process under `xvfb` against a stub RomM server with a fake core and a fake emulator that records its argv. Confirmed the full path: bridge injected → support probed → 512 KB ROM downloaded and cached → emulator spawned as `-L <core path> <rom path>`, with the ROM path surviving as a single argv entry despite spaces and parentheses in the filename. Also confirmed a clean first-run boot into the setup window.

<sub>Not verified: real RetroArch on real hardware across macOS/Windows/Linux, and the packaged installers (`electron-builder` is configured but builds are unsigned; notarization and a Windows signing cert are not wired up). Those need a machine with a display and signing identities.</sub>

<sub>None: the only new UI is one action entry that renders solely inside the desktop shell, which cannot be screenshotted from this environment.</sub>

---

**AI assistance disclosure**

This PR was written by Claude Code (Claude Opus 5) working from a maintainer's prompt, with the scope decisions (launch-only, RetroArch plus a configurable table, minimal frontend surface) chosen by the human. All code, tests, and the verification above were produced in that session. Please review it as you would any machine-assisted contribution, particularly the security-sensitive input validation in `electron/src/main/safety.ts` and `electron/src/main/emulator/resolve.ts`.

https://claude.ai/code/session_01AhBCzs8BaKzcnUKZFLLL1c

---
_Generated by [Claude Code](https://claude.ai/code/session_01AhBCzs8BaKzcnUKZFLLL1c)_


sdornan:
Addressed the Socket alerts in 9f2342d8 by removing the dependency they came from, rather than ignoring them.

**What I verified**

All three alerts were transitive under `electron-builder`, and all three are false positives on Socket's entropy heuristic:

| Package | Finding |
| --- | --- |
| `tiny-async-pool@1.3.0` | Source files are 11-24 lines with a 71 character longest line. Not obfuscated. |
| `electron-winstaller@5.4.0` | Official [`electron/windows-installer`](https://github.com/electron/windows-installer), MIT, readable source. What trips the heuristic is its `vendor/` directory of prebuilt Windows binaries (`7z.exe`, `Setup.exe`, `Microsoft.Deployment.WindowsInstaller.dll` and others). |
| `yargs@17.7.3` | Already present in `frontend/package-lock.json`, so not new to the repository. |

**Why I removed it instead of suppressing**

Suppressing alerts on your Socket policy is a maintainer's call, not mine. And `electron-builder` was only ever feeding `npm run dist`, which is not useful without code signing (no Apple Developer ID or notarization, no Windows certificate). So it was carrying three warnings and a directory of prebuilt Windows executables in exchange for a packaging path nobody can ship from yet.

Removing it resolves the alerts at the root and drops the workspace from **368 packages to 111** (lockfile 4747 → 1512 lines). Packaging returns in its own PR alongside signing.

**The feature is unaffected.** The shell still runs with `npm run dev`. I re-ran the end-to-end check after regenerating the lockfile: stub server, fake core, fake emulator recording its argv, and the full path still works (bridge injected → support probed → ROM downloaded and cached → emulator spawned as `-L <core> <rom>`). `tsc`, the 21 unit tests, and ESLint are all green on the new tree.

---

PR #4507, opened 2026-09-14
feat(streaming): let the launch screen pick which save to restore

LoneAngelFayt:
**Description**

The webstation launch screen already listed a ROM's save archives, but the claim always restored the newest one, so picking an older save did nothing. This adds a real save picker: the player chooses which stored archive gets hydrated onto the container, and the claim carries that choice through to restore.

Backend:

- `ClaimStreamingSessionRequest` gains an optional `save_id`, validated by `resolve_save_archive()` before the container is claimed. A save that is not the caller's own on this ROM returns the same 404 as a save that does not exist, so neither case leaks the other. A save that exists but cannot be restored here (wrong emulator, or a bare save file rather than a `.zip` archive, or a container whose emulator has no picker) returns 400 with the reason.
- `StreamingContainerSchema` gains `supports_save_picker`, driven by the emulator capability table. Emulators where the broker clears stale saves and restores the newest itself do not get a picker, and the UI reports that instead of offering a choice it cannot honour.
- Save hydration threads the pick through to the webstation upload. If the picked archive disappears between the pick and the claim, the launch proceeds with no save rather than silently restoring a different one.

Frontend (v2):

- The launch screen's save strip is now selectable and sends the pick on claim.
- Fixed the ordering bug behind the original report: `user_saves` arrives sorted on `updated_at`, which a content-hash write moves, so the list order did not match the order the backend restores in. The picker re-sorts on `created_at` then `id`, matching the backend exactly.
- Bare save files and other emulators' archives are filtered out of the picker, since neither can be restored onto the container.

This is stacked on #4478. The first six commits on this branch belong to that PR and will drop out of this diff once it merges. The new work here is the five commits from `feat(streaming): let the launch screen pick which save to restore` onward.

**AI assistance disclosure**

This PR was written primarily by Claude Code (Opus 5), including the code, the tests, and this description, under my direction and review. Any follow-up responses on this PR that are AI generated will be disclosed as such.

**Verification**

- `npm run typecheck`, `npm run build`: clean
- `npx vitest run`: 1241 passed across 115 files
- `uv run pytest tests/endpoints/test_streaming.py tests/endpoints/test_saves.py tests/endpoints/test_states.py tests/endpoints/test_assets.py`: 467 passed
- `python3 src/locales/check_i18n_locales.py` and `check_i18n_sorted.py`: pass
- `trunk fmt && trunk check`: clean

Not yet exercised by hand in the browser: the end to end launch with a picked save actually restoring inside a live container. The automated coverage stops at the claim payload and the backend validation.

**Checklist**


gantoine:
Two things surfaced by the review passes on this branch that are deliberately **out of scope here**, noted so they are not lost.

### 1. `_SAVE_PICKER_EMULATORS` is RomM's private copy of a broker-owned flag

`backend/handler/streaming/capabilities.py` hardcodes the set of emulators whose broker empties the save tree before a restore, mirroring the webstation broker's `clears_stale_saves` — a value that lives in another repo and that RomM has no independent way to know. The file already states the principle a few lines above, in the `_EMULATOR_CAPABILITIES` comment: enumerating this is "a second copy of the broker's core table that goes stale the moment the broker gains a platform".

The closest analogue, `memory_card_sync`, is explicitly *not* derived: it is an operator key in `config.yml` (`PLATFORM_OVERRIDE_KEYS`), because whether a container syncs whole cards is a property of that deployment.

Cost: a broker release that turns `clears_stale_saves` on for a new emulator needs a matching RomM code change before the picker appears, and an operator on a fork or a newer broker cannot turn it on at all. The other direction is worse because it is silent — if the broker stops clearing for one of the six, the pick still validates, the archive still uploads, and the container's own newer files quietly overwrite it with no error anywhere.

Options:
- **(a)** a `clears_stale_saves` key alongside `memory_card_sync` in `PLATFORM_OVERRIDE_KEYS`, resolved in `_resolve_one`, leaving the frozenset as the default rather than the authority. Entirely within this repo, and matches the precedent already set for this question.
- **(b)** have the broker report it (a field on the status/activate reply, or a capability GET) and cache it on the resolved container. The true fix, but it needs a broker-side change first.

### 2. The save list sorts on `created_at` but displays `updated_at`

`restorableSaves` in `Stream.vue` sorts on `(created_at, id)` to match the order the backend restores in, while every timestamp the rows and the preview render (`AssetStrip` sub-line, `AssetPreview` chips and tooltip) is `updated_at`.

So an archive created at 10:00 whose `updated_at` was moved to 12:00 by a content-hash rewrite sorts below one created at 11:00 and never touched — correct, since the second is what a claim restores by default — but reads "now" against the other's "1 hour ago". The tile the user reads as newest is second in a list advertised as newest-first.

Not fixed here because `AssetStrip` / `AssetPreview` are shared with the EmulatorJS player and the GameDetails "Save data" subtab, so changing the displayed timestamp is a cross-feature change rather than a save-picker one.

gantoine (approved):

PR #4539, opened 2026-09-16
fix(player): stop duplicate mkdir throwing on zip directory entries

TRusselo:
**Description**

A zip that stores a standalone directory record for a folder its file paths already created makes `downloadRom()`'s extraction callback call `FS.mkdir()` on a path that already exists, throwing `ErrnoError 20` once per record. The path-walk branch immediately above it already guards with `analyzePath().exists`; the standalone-directory branch does not.

Measured across a 506-ROM library: **455 throws over 34 ROMs**, worst case 115 in a single game. The extraction still completes, so it surfaces as console noise and wasted work rather than a hard failure, which is probably why it has gone unnoticed.

Fixed upstream in EmulatorJS `9a12941` ("Check for directory existance before attempting to create it.", 2026-04-27). That is in their `main` and tagged `v4.3.0-pre`, but it is not in any release, and this image pins `4.2.3`. So this backports it to the extracted files at build time.

Both copies are patched: `emulator.min.js` (loaded normally) and `src/emulator.js` (loaded when `emulatorjs.debug` is on). Each `sed` is followed by a `grep -q` so the build fails loudly if a later EmulatorJS version changes this code, rather than silently not applying. The whole step can be dropped once `EMULATORJS_VERSION` reaches 4.3.0.

**AI assistance disclosure**

AI-assisted, as required by `CONTRIBUTING.md`. Claude Opus 5 diagnosed the failure from browser console logs, located the upstream commit, and wrote the Dockerfile step and this description. The measurements above are from real runs against my own library, and the resulting image was built and used.

**Checklist**

<sup>No unit tests: the change is a build-time patch of a vendored asset, with the `grep -q` guards acting as the build-time assertion.</sup>


zurdi15:
Thank you for the effort but I'm not gonna merge a messy patch for something that emujs will release at some point. Lets wait for them to release 4.3.0

TRusselo:
OMFG... yeah, even I have rebased off of that new code and not running that patch anymore. DAYS AGO
sorry.

fuggin claude has been dumb AF the past week+
today it doesnt understand how version numbers work.  WTF is goin on?

I did approve the drafted PR. that is on me. my bad there, recognised the code, didnt think to check it still existed after the rebase, though wouldnt expect a PR for dead code from days before....

PR #4542, opened 2026-09-16
fix(streaming): make broker save-clearing configurable and date the picker by its sort key

LoneAngelFayt:
> [!NOTE]
> Stacked on #4533 (`streaming-patches-2`). This branch is rebased onto it, so
> the diff here currently carries that PR's commits too; they drop out once
> #4533 merges. Review the last two commits only.

**Description**

Two follow-ups gantoine recorded on #4507 as deliberately out of scope for the save picker.

**1. `clears_stale_saves` is an operator override, not a hardcoded table** (`5e217c7`)

`_SAVE_PICKER_EMULATORS` in `backend/handler/streaming/capabilities.py` was RomM's private copy of the webstation broker's `clears_stale_saves` flag, which lives in the broker's repo. Having the broker report it is the real fix, but that needs a broker-side change first, so this takes the option the note preferred: a `clears_stale_saves` key alongside `memory_card_sync` in `PLATFORM_OVERRIDE_KEYS`, resolved in `_resolve_one`, leaving the frozenset as the default rather than the authority. An operator on a fork or a newer broker can now say so in `config.yml`; everyone else is unaffected.

Two things beyond the letter of the note:

- Setting it on a non-webstation container logs a warning rather than being silently ignored, since only a webstation container restores a picked save.
- It joins `interchangeable_with`. The picker is advertised from the head of the pool, so a member that does not clear would take the pick and then let its own newer files overwrite the restored archive, with no error anywhere.

**2. The save picker showed a timestamp it did not sort on** (`d097df9`)

The picker sorts on `(created_at, id)` to match the backend's restore order, but the rows displayed `updated_at`. `recompute_save_content_hashes` writes `content_hash` through `update_save`, which bumps `updated_at` without the save changing, so after that task runs the tile a user reads as newest can sit second in a list advertised as newest-first.

Flipping the display to `created_at` globally would be worse: GameDetails and the EmulatorJS lists are ordered by `updated_at` (backend `sort_user_saves`), so a global flip just moves the mismatch. Instead `AssetList` and `AssetPreview` take a `timestamp` prop defaulting to `"updated"`, and the Stream picker, the one list ordered by `created_at`, passes `"created"`. The invariant is "show whatever the list is ordered by". Adds `rom.created` across all 18 locales.

**Checklist**

Verified: `uv run pytest tests/endpoints/test_streaming.py` (351 passed), `npx vitest run` (1306 tests), `npm run typecheck`, both i18n scripts, `trunk check --upstream upstream/master` clean.

**AI assistance**

Written with Claude Code. I directed the work, reviewed every change, and ran the test and lint suites above locally.


gantoine:
@greptileai

gantoine (approved):

PR #4564, opened 2026-09-16
feat(v2): cross-fade the isolated player launch

sdornan:
**Description**

Launching EmulatorJS (`/rom/:id/ejs`) or js-dos (`/rom/:id/jsdos`) from the gallery has always been a hard cut. It has to be a full document load: `SharedArrayBuffer` needs the COOP/COEP headers that nginx attaches only to those two paths (`~^/rom/.*/(ejs|jsdos)`, see `docker/nginx/templates/default.conf.template`), so the same-document morphs in `useViewTransition` can never span this navigation. A cross-document view transition can, and this makes those two launches cross-fade instead.

The opt-in is armed by the launch itself, not declared in the stylesheet. Neither document has a static `@view-transition`: that rule is inherited by every page, so every other hard navigation would opt in as well and cross-fade too. Instead `armCrossDocumentTransition()` records the destination in `sessionStorage` and appends the descriptor, and the parser-blocking script in `index.html` arms the player document only when the marker names its own path.

```mermaid
sequenceDiagram
    participant G as Gallery document
    participant S as sessionStorage
    participant P as Player document (COOP/COEP)
    G->>S: write /rom/1/ejs
    G->>G: append descriptor (opt in)
    G->>P: location.assign (hard load)
    P->>S: read and consume the marker
    P->>P: path matches, so append descriptor
    Note over G,P: both ends opted in, so the transition runs
```

Both ends have to opt in, which is why the marker is written first and gates the arming: arming one end alone yields a transition that is aborted, logging a console error and animating nothing. Under reduced motion neither end is armed, so that case stays the plain hard load it is today.

**For reviewers:** the obvious alternative is a static `@view-transition` in `global.css` plus `skipTransition()` in the incoming document when the launch is not marked. I measured that in Chromium and it does not hold up. An unrelated hard navigation between two same-origin-isolation documents never receives `pagereveal` on the incoming side, so the skip cannot run and the transition completes: a cross-fade on navigations outside this feature. A launch that is not marked does receive it, and the skip then produces a Chrome-internal `AbortError: Transition was skipped` in the console. Gating the opt-in avoids both, and leaves the blast radius at the player launch.

Scope is the launch only: the two isolated player documents. Returning from the player (the back links in `PlayerShell`, which are real anchors since #4560) is unchanged, as is every SPA transition, and so are PICO-8, Ruffle and the streaming view, which are SPA pushes that already morph in-document. The v1 console path `/console/rom/:id/play` is served cross-origin isolated too, but v1 never arms the marker, so it keeps its hard cut.

Compatibility: Chromium runs the transition across the COOP browsing-context swap. Firefox and the WebKit build used for verification have no cross-document support and keep the hard cut, with no console errors either way.

**Checklist**

The unit tests and a four-case harness against the real extracted head script pass, but the end-to-end pass on the running app (docker compose, so COOP/COEP are present and the emulator actually boots) is still to be done, which is why the first box is unchecked. Opening as a draft for that reason.

Pending the running-app pass described above. A cross-fade is hard to convey in a still, so the pass will capture the pre-launch gallery frame, a mid-transition frame, and the settled player, in both themes.

---

AI assistance disclosure: this change was written by Claude (Anthropic) under my direction, including the implementation, the tests, and the platform verification described above. AI was also used to review the design against the alternative rejected in the reviewers' note.


sdornan:
Superseded by #4565. This one was opened from the `sdornan/romm` fork, which broke two workflow steps for reasons unrelated to the diff: a `pull_request` run from a fork gets a read-only token, so `Build Experimental Docker Image` fails to post its comments, and its build jobs check out `github.head_ref` against `github.repository`, so the fork's branch name 404s and the docker build never starts. The branch now lives in this repo and the same change is open as #4565.

Issue #4522, opened 2026-09-15
[Bug] Manual Match cover previews blocked when RomM is served with COEP require-corp


arpadkorossy:
**RomM version**  
5.2.0

**Installation**  
Docker, accessed through a reverse proxy

**Clients tested**  
- Desktop browser (Firefox)
- Android mobile browser

**Describe the bug**

Cover artwork does not load in the **Match ROM** search-results dialog when RomM is served with the cross-origin isolation headers required for threaded EmulatorJS cores:

```http
Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Embedder-Policy: require-corp
```

The metadata search itself works correctly. Results are returned, names and provider information appear, and the ROM can be successfully matched.

However, the remote cover artwork in the search results is blocked by the browser.

After selecting the result and completing the match, the cover artwork displays normally in RomM.

**Why COOP/COEP are enabled**

These headers are enabled intentionally to support `SharedArrayBuffer` / threaded WebAssembly for EmulatorJS cores such as DOSBox Pure.

Removing `Cross-Origin-Embedder-Policy: require-corp` is therefore not a desirable workaround because it can break browser-based emulation that requires cross-origin isolation.

**To reproduce**

1. Serve RomM through a reverse proxy with:

```http
Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Embedder-Policy: require-corp
```

2. Open a ROM in RomM.
3. Select **Match ROM**.
4. Search for a game using IGDB.
5. Observe that the search result is returned successfully, but its cover image does not load.
6. Open the browser console.

For example, Firefox reports:

```text
The resource at “https://images.igdb.com/igdb/image/upload/...”
was blocked due to its Cross-Origin-Resource-Policy header
(or lack thereof).
```

7. Select the result and complete the match.
8. Observe that the game's cover art then displays normally in RomM.

The issue occurs on both desktop and Android.

**Observed behavior**

The Match ROM dialog appears to load provider artwork directly from the third-party image URL, for example:

```text
https://images.igdb.com/...
```

Because the RomM page is served with:

```http
Cross-Origin-Embedder-Policy: require-corp
```

the browser refuses to embed that cross-origin image when the image host does not return an appropriate `Cross-Origin-Resource-Policy` or CORS response.

The local RomM UI graphics and metadata-provider icons display normally.

**Expected behavior**

Manual-match cover previews should display even when RomM is configured with the COOP/COEP headers needed for supported EmulatorJS cores.

[...]

PR #4567, opened 2026-09-16
feat(v2): open the EmulatorJS and js-dos launch views without a page load

zurdi15:
**Description**

Pressing Play on an EmulatorJS or js-dos game loaded the launch view with a full document load (`window.location.assign`), while PICO-8, Ruffle and streaming navigate within the app. The reason was cross-origin isolation: nginx attaches the COOP/COEP headers only to the player URLs, and only a document loaded with them exposes the `SharedArrayBuffer` that threaded cores need. Every launch paid that reload, while only three cores need threads (`dosbox_pure`, `ppsspp`, `azahar`) and js-dos 8.4.1 does not reference `SharedArrayBuffer` at all (checked in `js-dos.js`, `emulators/wdosbox-x.js` and `wdosbox.js`).

- **SPA navigation into both launch views**, with the same cover morph the other players get. `useGameActions.play()` no longer special-cases any player; `playPath()` is the single resolver for both the navigation and the launch links.
- **Reload only when a launch needs it.** New `useIsolatedLaunch(player, romId, isIntent)`: when EmulatorJS boots a thread-required core without `SharedArrayBuffer`, the view keeps the pre-play selection (save/state, firmware, slot) in `sessionStorage`, reloads its own URL so nginx isolates the document, and boots from that intent on arrival. It refuses the reload, and the view reports the context instead, when the reload cannot help: outside a secure context, where storage would not keep the selection, or from a view that already came from one.
- **Leaving a player is a full navigation only when it has to be.** New `usePlayerExit(runtimeBound)`: the document is replaced when it is cross-origin isolated (the rest of the app cannot embed third-party images under that policy, which is what #4522 describes for Match ROM covers, and what v2 reproduced natively after every EmulatorJS session) or when the EmulatorJS runtime was injected (`emulator.min.js` declares top-level classes, so a second injection in the same document throws). Otherwise it is an SPA navigation, so a visit to the launch view that never plays costs no reload either way.
- **The js-dos runtime is a document-level singleton** (`jsDosRuntime.ts`): one injection and one asset base however many views mount, held as the in-flight promise so a second mount waits rather than injecting a second copy, and a failed load stays retryable.
- **Presence stops are scoped to a started session.** `activity:stop` is addressed to the device, so a view that never announced anything could clear what another tab was playing; `useActivityPresence` grows the `stop()` its call sites were spelling out by hand.
- **Extensibility**: a new threaded core is one entry in `areThreadsRequiredForEJSCore`; a new player that needs isolation uses the two composables the way `EmulatorJS.vue` does. Documented in `docs/FRONTEND_ARCHITECTURE.md` ("Cross-origin isolation").
- **Dev server**: `scripts/playerIsolationHeaders.ts` mirrors the nginx map, and a test asserts the two lists stay equal, so a route added to one and not the other fails CI rather than diverging silently. `SharedArrayBuffer` still requires a secure context (`localhost` or https).

Two consequences worth knowing: EmulatorJS's own "Threads" menu entry no longer appears for cores that do not require threads, because that document is not isolated; and a launch that does need the reload now spends one extra rom and firmware request, since the reloaded document fetches them again. Predicting the core before navigating would avoid that, at the cost of putting isolation logic back into `useGameActions`, so it is left as a follow-up.

```mermaid
flowchart TD
    A[Play on card / details] -->|router.push + cover morph| B[Launch view]
    B -->|Play| C{Core needs threads\nand no SharedArrayBuffer?}
    C -->|no| D[Boot in this document]
    C -->|yes, and a reload can help| E[Keep intent, location.reload]
    E -->|nginx adds COOP/COEP| F[Isolated launch view]
    F -->|apply intent| D
    C -->|yes, but it cannot| G[HTTPS required error]
    D --> H{Leave: isolated or\nEmulatorJS injected?}
    B --> H
    H -->|no| I[SPA navigation]
    H -->|yes| J[location.replace: fresh document]
```

Refs #4522 (the v2-native reproduction is gone; a proxy that sets the headers globally still hits it).

**AI assistance disclosure**: this PR was written with Claude Code (Claude Fable 5.1 and Claude Opus 5). The analysis, code, tests and docs were produced by the model under human direction and review; the js-dos and EmulatorJS runtime findings were verified against the published 8.4.1 and 4.2.3 bundles. It then went through the repo's own review passes (security audit, code review, simplification, polish), and the Greptile and Copilot review findings on this PR were applied in `8d4e85636`.

**Checklist**

- [x] I've tested the changes locally (Vitest, `vue-tsc`, `npm run build`, Trunk; browser pass pending, see below)

Tests: new suites for `useIsolatedLaunch`, `usePlayerExit`, `launchIntent`, `jsDosRuntime` and `playerIsolationHeaders` (including the nginx parity check); `useGameActions`, `JsDos`, `Stream` and `PlayerShell` updated for the SPA entry and the conditional exit. The dev server was checked directly: `/rom/:id/ejs`, `/rom/:id/jsdos` and `/console/rom/:id/play` answer with both headers, `/rom/:id` and `/rom/:id/pico8` with neither.

Browser pass still to do on a secure context: (1) Play from the gallery card and from details on a non-threaded core, no reload in or out; (2) a PSP or DOSBox Pure game, one reload at Play and automatic boot with the chosen save/state; (3) after a session, Back to details and open Match ROM, provider covers load; (4) a Win9x/DOS js-dos game boots without any reload, and a second js-dos game opened after it still loads its payloads.

No new visuals: the change is in how the launch views are reached and left. The launch views themselves are unchanged.

https://claude.ai/code/session_01R1VFUvsR5wYK3axxR1gebc


zurdi15:
Both review passes are addressed in `8d4e85636`, plus the repo's own `security-audit` / `code-review` / `simplify` / `review-polish` gauntlet.

**Greptile**

- *Storage errors break launches* (`useIsolatedLaunch`): valid, fixed. Reads and the write are guarded now. Storage that cannot be read is an absent intent, so the view opens and non-threaded games run regardless; a write that fails cancels the relaunch and the caller reports the context, rather than reloading into a document that lost the selection and would ask again forever. Covered by two new cases.

**Copilot**

- *`pagehide` can stop another session for the same device*: valid, fixed one level down. `activity:stop` is addressed to the device, so the guard belongs in `useActivityPresence`: a stop now needs a started session and happens once. That also fixes the stream player, which had the same exposure.
- *`documentBound()` misses a departure taken mid-injection*: valid, fixed. The view tracks the injection itself instead of waiting for `window.EJS_emulator` to appear.
- *js-dos asset base not carried between views*, and *two views can both inject the runtime*: both valid, fixed together. The runtime is now a document-level singleton (`jsDosRuntime.ts`) holding the in-flight promise, so a second mount waits for the load already running and follows the base it served from. A failed load stays retryable.

Also from the local passes: the reload is refused outside a secure context instead of costing a round trip to reach the same error; the dev-server header mirror was missing the console player route nginx covers, and a test now asserts the two lists stay equal; `usePlayerExit` owns the whole exit rule, so both views use its guard.

Security audit over the range came back clean: no new egress, no dependency, CI or infrastructure changes, navigation sinks fed only by router-resolved paths, and the stored intent validated field by field before use.

PR #4571, opened 2026-09-17
fix(deps): bump engine.io-client past the ws memory-exhaustion advisory

sdornan:
**Description**

Clears Dependabot alert [#203](https://github.com/rommapp/romm/security/dependabot/203): `ws` below 8.21.0, memory exhaustion DoS (GHSA-96hv-2xvq-fx4p / CVE-2026-48779).

`ws` is not declared directly. The vulnerable copy was transitive:

```
socket.io-client@4.8.3 -> engine.io-client@6.6.5 -> ws ~8.20.1
```

`engine.io-client@6.6.5` pinned `ws` to `~8.20.1`, a tilde range that can never reach the 8.21.0 fix, so the alert could not clear on its own. `engine.io-client@6.6.6` moves to `ws ~8.21.0`, and `socket.io-client@4.8.3` already permits `~6.6.1`, so the bump resolves within the ranges already declared. No `overrides` entry is needed, and the existing `overrides` block (rolldown, esbuild, form-data) is left alone.

The lockfile diff is two version bumps plus one dedupe:

- `engine.io-client` 6.6.5 to 6.6.6
- `ws` 8.20.1 to 8.21.3
- happy-dom's nested `ws@8.21.0` folds into the top-level copy

`npm audit` goes from 2 high to 0.

One scope note: Dependabot reports this as a runtime dependency because `socket.io-client` sits in `dependencies`, but `ws` is engine.io's Node transport. The browser build uses the native `WebSocket`, so exposure in the shipped web UI is low. The bump is still worth taking, since it is a clean in-range resolution and the same lockfile feeds Node-side tooling.

**Checklist**

**AI Assistance**

This PR was written primarily by Claude Code. The dependency chain analysis, the lockfile change, and the verification runs were all agent-authored and then reviewed by me.


sdornan:
@greptileai

gantoine (approved):

PR #4575, opened 2026-09-17
chore: expand ts typecheck to `frontend/scripts/**/*.ts`

andest01:
**Description**

Expand frontend's `npm run typecheck` to include Node tooling in `frontend/scripts/**/*.ts`.

Create a new `tsconfig.node.json` and target `scripts/**/*.ts` files inside the `include` property.
```JSONC
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler",
    "strict": true,
    "noEmit": true,
    "skipLibCheck": true,
    "types": ["node"]
  },
  // Aids `npm run typecheck` in expanding typechecking
  // to Node/Vite tooling that vue-tsc does not cover.
  "include": ["scripts/**/*.ts"]
}
```

Then we target it specifically by altering `npm run typecheck`:
```JSON
  "typecheck": "vue-tsc --noEmit && tsc --noEmit -p tsconfig.node.json",
```

We can expand typechecking to other files that are currently `*.js` by renaming them to `*.ts` and validating as we always have. No further changes to CI are needed.

This PR is built off of a [previous failed attempt.](https://github.com/rommapp/romm/pull/4572)
**Checklist**

# Tests:
I know the PR seems a bit trivial, but I wanted to make sure that it was a clear, simple start to working on this project. Here's how I tested it.

Front end tests were mainly sniff tests on my part. Typecheck it, test it, and build it.
```cmd
PS D:\romm\frontend> npm run typecheck && npm run test && npm run build
```

I ran most of it out of docker and from the command line, but I'd be happy to try to locally host the full build before moving on.


andest01:
OK here's an update.
image

At first, I didn't want to extend the package.json's typecheck because  I felt that was a little too far. But then I thought maybe it could be useful to extend typescript coverage outside of `vue-tsc` which I believe actually ignores (strangly?) `vite.config.ts`.

The idea is actually take it as an opportunity to add typescript to non-vue areas as well, so I tested this idea with the scripts directory. 

Verify it fails when it should. 
Code_PQeKZDNwkO

And if this fix wasn't there, what could happen:
Code_PHQCOlNpni

sdornan:
Checked this against current master. It merges clean, but one master commit since your branch point breaks it:

```
FAIL  test/playerIsolationHeaders.test.ts > the template vite.config.js points at > is the one the tests read
Error: ENOENT: no such file or directory, open 'vite.config.js'
 ❯ test/playerIsolationHeaders.test.ts:139:12
```

`frontend/test/playerIsolationHeaders.test.ts` was added after your branch point and reads `vite.config.js` by literal name (16 pass, 1 fails). Worth updating that string in the same commit when you rebase.

The typecheck half is the part with real substance. `frontend/scripts/**` was never in a tsconfig, so `build-tokens.ts`, `precompress.ts` and `playerIsolationHeaders.ts` were transpiled in CI but never checked. I confirmed the new `tsc -p tsconfig.node.json` catches a bad call in `precompress.ts` (`TS2304: Cannot find name ...`) that `vue-tsc --noEmit` passes over.

One suggestion to get more out of it. Because the config uses the callback form, a misspelled option name still slips through: `tsc` flags wrong types on known options (`build.target: 42` errors) but not unknown names. Annotating the callback's return type turns the excess-property check back on:

```ts
import { defineConfig, loadEnv, type UserConfig } from "vite";
// ...
export default defineConfig(({ mode }): UserConfig => {
```

With that annotation, `tsc` catches typos at every level and suggests the fix:

- `serverr` -> `TS2561: Object literal may only specify known properties, but 'serverr' does not exist in type 'UserConfig'. Did you mean to write 'server'?`
- `build.cssTargets` -> `... does not exist in type 'BuildEnvironmentOptions'. Did you mean to write 'cssTarget'?`
- `injectRegisterr` -> `... does not exist in type 'Partial<VitePWAOptions>'. Did you mean to write 'injectRegister'?`

Worth having, because Vite ignores unknown config keys without warning, so a typo there fails silently rather than loudly.

Keeping the callback is the right call. It is the only place `mode` comes from, and `loadEnv(mode, ...)` plus the DEV_* flags depend on it; `process.env.NODE_ENV` is only ever production/development, so the plain object form would misread the env files for any custom mode.

andest01:
@sdornan thanks!

Since there's clear value in expanding the `npm run typecheck` to outside of `vue-tsc`'s remit, how about I dramatically simplify the surface of the PR to _exclude_ the `vite.config.js` file _for now_ and just cover files that already exist? That way I can expand on the original premise by adding some better types to what we already have, but stay out of the merge conflict zone?

`vite.config.js` would no longer be renamed, but we keep the `tsconfig.node.json` file and strengthen our typescript as we see fit inside of `frontend/scripts`

That means I'm out of merge conflict heck.
firefox_vaSuPfY2yV


lol the original point of this PR was to change the name, but I'm fine with adaptation if it suits us! I'll make a PR against this PR with the adjustments and you can re-review when you get a minute.

andest01:
OK, I've altered the scope of the PR and changed the title and the description to reflect that. 

@sdornan I like this simpler PR because it's very very small and easier to merge in after your larger release is passed.

To be clear, the `vite.config.js` file is back and untouched. You shouldn't see it at all in this PR's history.

gantoine (approved):

gantoine:
@greptileai

PR #4599, opened 2026-09-17
feat: promote platforms that have games in PlatformSelect

andest01:
**Description**

OPT-IN: introduce property that promotes platforms that have games (`rom_count > 0`), and puts them at the top of the list. Then, it adds a divider and places platforms that have no games underneath it.

Added a count badge on the right.

Adds tests and storybook.

Covered specific workflows with search. When a user actively searches, it stops partitioning and displays the results alphabetically directly, as before.

Production code is unaffected by this. The idea is we can get feedback on it and then add it.

Speaking for myself, I love this feature because it makes search much more usable.

I used a bit of AI to help create `play` sections of stories as well as learn how to use `compute` in Vue. I also used AI from this PR via greptile.

**Checklist**

image

image


frontend/src/v2/components/shared/PlatformSelect.vue
```diff
@@ -156,6 +160,26 @@ const forwardedSlotNames = computed(() =>
...
+const dividerAfter = computed(() => {
+  if (!props.promoteFilled) return undefined;
+  const { promoted, remaining } = promotion.value;
+
+  // if there's nothing to promote, or nothing to remain, we don't need to divide.
+  const makesNoDifference = promoted.length === 0 || remaining.length === 0;
+  if (makesNoDifference) return undefined;
+
+  const lastKey = promoted.at(-1)![props.itemKey];
+  return (platform: Platform) => platform[props.itemKey] === lastKey;
```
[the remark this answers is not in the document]
andest01: This was addressed in later commits: we now remove the divider while actively filtering. 

See PlatformSelect.promoteFilled.test.ts and story ‘Promotion on — typing in panel search’.

frontend/src/v2/components/shared/formatPlatformRomCount.ts
```diff
@@ -0,0 +1,7 @@
+export const PLATFORM_ROM_COUNT_CAP = 9999;
+
+/** Display rom count in platform picker rows; caps at `9999+`. */
+export function formatPlatformRomCount(count: number): string {
```
andest01: I'm not sure if I'm wild about the amount of one-export files I'm making for this PR. Do we have a convention maintainers would like me to follow?

zurdi15: I would like to have one ``.vue``, one ``.ts`` and one ``.test``. Could you merge the ``platformFilled`` test in the pre-existing test file? 

Everything else looks solid

zurdi15 (changes requested):
1. Check the commento about merging test files
2. Fix the CIs

Great idea btw :)

andest01:
@zurdi15 thank you! Here are the results:
image

I found a problem with `trunk` on windows, so that was causing some problems.
But I did run this:
```terminal
$ trunk fmt && trunk check && npm run typecheck && npm run test
```
...and it passed. Let's see what happens with CI now.

**hey, quick question here**

May I actually turn it on in the search in this PR? Let's assume that the label `"on-hold"` means this won't get merged for a bit. To me, that sounds like it might be ok (with your emoji blessing!) to add it to a few places:

image

🟢 - I feel i can add it here no problem
❓ - interesting, but doing nothing here would also be acceptable.

** What would you like me to do? **

### Option A: Nothing.
Literally leave it alone. Let this PR in, even though it does nothing on production (yet).

### Option B: Just turn it on where it's marked 🟢 
image

It's pretty nice.

### Option C: Expand the idea further, allowing a callback in the `RSelect` component `dividerAfter` prop... perhaps we allow the caller to supply their own "partition" however they want?

I tried this myself with "Genre". It was a huge pain with lots of backend changes. No thanks... But intriguing.

### My Advice

I recommend you review what you like _a la carte_ in Option B.

I'm not doing anything until you let me know :)

**Currently this PR, if accepted, will make no functional changes to the front end and nobody should notice a thing**. 


AI disclosure: I wrote this text, and I had AI explore a few of the options.

andest01:
I've updated this PR so it actually is loaded into the app in the places I described above.

zurdi15 (approved):

PR #4605, opened 2026-09-17
build(frontend): raise the Chromium floor to 117, drop the hand-rolled build targets

sdornan:
**Description**

A review of whether our browser minimums should move, plus a simplification of how the build derives its targets.

**1. Chromium support floor 111 to 117.** CSS subgrid landed in Chrome/Edge 117, but the floor sat at 111, so v2 layouts using subgrid needed hand-written `@supports` fallbacks. `HLTBStrip.vue` had one (added in #4604); `Soundtrack/TrackRow.vue` and `views/Player/EmulatorJS.vue` use subgrid with no fallback and silently degrade below 117 today. Raising the floor removes the tax instead of spreading it: Chrome 111-116 is **0.43%** of users globally and Edge 111-116 is **0.00%**, both stale installs and embedded webviews rather than anyone who will ever update. The `HLTBStrip` fallback goes with it.

**Safari and iOS stay at 16.4.** That floor is doing real work for this project: 16.4 is the last stop for A11 devices (iPhone 8/X, iPad 5th/6th gen), exactly the kind of hardware repurposed as a couch front end for a library, and they can never update past it. Bumping to 17.4 would cost ~1.9pp of global users and buy nothing, since nothing in the codebase needs it.

**2. `build.target` and `build.cssTarget` are gone.** `target` was `esnext`, which is what caused the bug the `cssTarget` code existed to work around: `cssTarget` inherits `target`, and Vite's `convertTargets` skips the `esnext` entry entirely, so Lightning CSS received no targets, assumed everything was supported, and dropped the generated `-webkit-backdrop-filter`.

Vite's default already does the right thing here. `build.target` defaults to `'baseline-widely-available'` (`chrome111, edge111, firefox114, safari16.4, ios16.4`) and `cssTarget` inherits it, so both options are now simply absent. Verified this is a true no-op rather than an assumption: building with and without an explicit target produces all **858 output files byte for byte identical**.

That baseline sits below the Chromium 117 support floor above, which is fine: `build.target` bounds emitted *syntax*, not feature availability, and compiling lower only down-levels slightly more than needed. No bundler setting would make subgrid work on Chrome 111 anyway.

**3. `.browserslistrc` and the `browserslist` devDependency are deleted.** `vite.config.js` was their only consumer (Tailwind v4 does not read browserslist), so with the conversion gone they were dead config that would drift.

Net effect on the config: a 35-line hand-rolled browserslist-to-esbuild converter, two `build` options and a dependency all removed, replaced by Vite's default plus a comment marking why `target` must never be set to `esnext`.

<details>
<summary>On the Lightning CSS docs' suggested approach, and why it does not apply</summary>

The Lightning CSS docs recommend `targets: browserslistToTargets(browserslist(...))` under `css.lightningcss`, including in their Vite example. That path does not work here, for two independent reasons, both verified against `vite@8.0.16` in `node_modules`:

- `resolveCSSOptions` only populates `css.lightningcss.targets` when `css.transformer === 'lightningcss'`. Our transformer is the default, `postcss`, and switching it would displace the PostCSS pipeline.
- Even then, `minifyCSS` spreads `config.css.lightningcss` and *then* overrides `targets: convertTargets(config.build.cssTarget)`, so on the minify path `build.cssTarget` wins regardless.

Also worth knowing: `build.cssTarget` does not accept the `'baseline-widely-available'` string. Vite resolves that literal on `merged.target` only, before `cssTarget ?? target`, so setting it on `cssTarget` directly reaches `convertTargets` and throws `Unsupported target`. Leaving both unset, as this PR does, is the only clean way to get the baseline into CSS.
</details>

Nothing else in the codebase is asking for a bump. I scanned v1 and v2 for post-baseline features: `color-mix()` (470 uses), `:has()` (30) and `backdrop-filter` (71) are all inside the floor; `document.startViewTransition` is feature-detected in `useViewTransition`; `field-sizing` is deliberately unused with a comment at `RTextField.vue:621` explaining why; and there are zero uses of `Object.groupBy`, `Promise.withResolvers`, the native popover API, `structuredClone`, `toSorted`, `Intl.Segmenter` or `AbortSignal.any`.

**AI assistance disclosure**

This PR was written with AI assistance (Claude Code). The investigation, the code changes, the commits and this description were all AI-authored, then verified against the checks below.

**Checklist**

On tests: this is a build-config and CSS-deletion change with no new logic to cover. The existing suite passes unchanged (**152 files, 1540 tests**), along with `npm run typecheck`, `npm run build`, Prettier, ESLint and `check_i18n_locales.py`. I also verified in the built output that `-webkit-backdrop-filter` is still emitted (65 occurrences, the Safari glass regression described above) and that no `@supports ... subgrid` block remains.

Rendered the `HLTBStrip` compiled CSS against the real component DOM in Chromium, both themes, including a wrapping long label and a missing sub-row as an alignment probe. Subgrid keeps every value on a shared baseline across columns (measured `getBoundingClientRect().top`: `[87,87,87,87]` and `[228,228,228]`), which is the behavior the removed fallback could only approximate.

Screenshots are saved outside the repo and need to be dropped in here:

- `/tmp/claude-0/-home-user-romm/ee41c072-33da-53e4-9be1-1e0a8d741c93/scratchpad/hltb-dark.png`
- `/tmp/claude-0/-home-user-romm/ee41c072-33da-53e4-9be1-1e0a8d741c93/scratchpad/hltb-light.png`

Note the full app could not be launched in this environment (no Docker daemon available), so this is component-level browser verification rather than a pass over the running app.

---

Separately, while confirming the `-webkit-` guidance in the config comment, I found a live pre-existing bug this PR does **not** touch. Two v2 rules hand-write the `-webkit-backdrop-filter` twin after the standard property, and Lightning CSS collapses the pair to whichever is declared last, so the standard property is the one dropped:

- `src/v2/views/Player/Stream.vue:1837` → compiled `.r-v2-stream__panel` has only `-webkit-backdrop-filter`
- `src/v2/components/Player/StreamStage.vue:382` → compiled `.r-v2-stage__bar` has only `-webkit-backdrop-filter`

Both surfaces lose their blur on Chrome, Edge and Firefox and work only in Safari. Every other `backdrop-filter` rule in the build is fine. This predates this PR and is unchanged by it, so it belongs in its own change: deleting the two hand-written lines is the whole fix, since Lightning CSS generates the prefix from the targets.

https://claude.ai/code/session_01BEtFQ17BQDS52wXosTjcUU


sdornan:
Closing this without merging. The findings are captured in #4606 so they are not lost, and the branch `claude/browser-min-version-support-ojgqw4` is left in place if anyone wants the diff later.

For reference, this PR was green (all 12 checks) at the time of closing. It contained three things:

1. Raising the Chromium support floor 111 → 117 for CSS subgrid, and deleting the now-redundant `@supports` fallback in `HLTBStrip.vue`
2. Removing `build.target: 'esnext'` and `build.cssTarget` in favour of Vite's default `'baseline-widely-available'`
3. Deleting `.browserslistrc` and the `browserslist` devDependency, whose only consumer was `vite.config.js`

---

PR #4634, opened 2026-09-19
fix(patcher): budget a zip member as it is extracted

gantoine:
**Description**

Found during the `5.3.0-beta.1` security audit. Authenticated DoS, needs `roms.write`.

`_extract_zip_member` bounded the extraction two ways, and both had a gap:

- the pre-checks compare `selected.file_size` and the sum of `entry.file_size` against the cap, but that is a header field, i.e. whatever the uploader wrote there;
- the real check, `output_path.stat().st_size`, only ran once the member had already been written to the patcher temp directory in full.

So a declared size under the cap let an arbitrarily large member land on disk before anything objected.

Worth being precise about reachability, because the plain path is already safe. For an archive `zipfile` can decode, a small declared size makes `zipfile` truncate its output at `file_size` and then raise `BadZipFile: Bad CRC-32`, so the bomb never materializes. I verified that directly. The hole is the fallback: `read_zip_archive_files` hands an archive whose compression method `zipfile` cannot decode (PPMd, say) to `7zz` as a whole, and `7zz` decompresses it itself without consulting the central directory. There the header is pure fiction and the pre-check waves it through.

The fix counts bytes as they are written and stops at the cap, which is the pattern `utils/memory_cards.py` already uses for the same reason ("that header is whatever the uploader put there"). That makes the check on the finished file redundant, so it goes.

The test drives the shape of the 7zz path: a real on-disk zip declaring 3 bytes, a reader that yields far more, and an assertion that extraction bails on the first chunk past the budget rather than draining all 100. Without the fix it writes all 1600 bytes and the assertion reads `100 == 1`.

**Checklist**

`backend/tests/utils/test_rom_patcher.py` and `backend/tests/utils/test_archives.py` (50 passed, 1 skipped), `trunk fmt` / `trunk check` clean.

**AI assistance**

Written by Claude Opus 5 via Claude Code: the audit that found it, the reachability analysis, the fix, the test, and this description. Reviewed by @gantoine before opening.


gantoine:
Fixed in 2645d47 — but flagging that the stated rule does not match this repo's config, so the reasoning behind it shouldn't be reused as-is.

**Done:** `_lying_reader` is now annotated, and against `read_zip_archive_files`'s real signature rather than a generic `*args`/`**kwargs`, so the fake actually pins the contract it stands in for:

```python
def _lying_reader(
    _file_path: Path, _excluded_names: list[str], _excluded_exts: list[str]
) -> Iterator[tuple[str, int, Iterator[bytes]]]:
    def _chunks() -> Iterator[bytes]:
```

This is consistent with the file's other helpers (`_fake_patcher`, `_write_zip`), which is the real argument for the change.

**Two corrections to the rationale:**

1. *"Trunk runs mypy over backend/tests/"* — it does not. `.trunk/configs/mypy.ini` is three lines and the third is `exclude = ["tests", "alembic"]`. Nothing in `backend/tests/` is type-checked, so an unannotated fake there weakens nothing and cannot fail CI. `disallow_untyped_defs` isn't set either, so this wasn't a merge blocker.

2. *"the repository requirement that test fakes overriding production dependencies be annotated"* — the nearest precedent points the other way. `test_apply_patch_uses_archive_fallback_for_unsupported_zip_method`, twenty lines up, fakes the *same* `read_zip_archive_files` with a bare `MagicMock(return_value=...)` and no typing at all. So the convention isn't uniform, and describing it as a requirement that "must be satisfied before merging" overstates it.

Worth taking as a style improvement rather than a defect. `trunk fmt` / `trunk check` clean, 17 tests pass, and I re-confirmed the annotated fake still fails when the budget guard is removed.

